PR #2, opened 2017-04-06
add manageiq-providers-openstack.rb

durandom:
fixes failing specs


Fryguy:
Curious why this would fix it, since this is how it is in the other repos

durandom:
the generator missed to generate that file :(

Fryguy:
ohhhhh.  ok merging.

PR #3, opened 2017-04-06
Set image name to id if name is nil in Glance

tzumainn:
https://bugzilla.redhat.com/show_bug.cgi?id=1438100

PR #1, opened 2017-04-06
Handle case where instance flavor is deleted

tzumainn:
https://bugzilla.redhat.com/show_bug.cgi?id=1437624

PR #5, opened 2017-04-07
Implement Cloud Provider Graph Inventory Refresh

mansam:
This pull request implements Graph Refresh for the Openstack cloud provider. Ported over from https://github.com/ManageIQ/manageiq/pull/14241 on the core repository.


tzumainn:
@miq-bot assign @Ladas

app/models/manageiq/providers/openstack/inventory_collection_default/cloud_manager.rb
```diff
@@ -0,0 +1,269 @@
+class ManageIQ::Providers::Openstack::InventoryCollectionDefault::CloudManager < ManagerRefresh::InventoryCollectionDefault::CloudManager
+  class << self
+    def availability_zones(extra_attributes = {})
+      attributes = {
+        :inventory_object_attributes => [
```
Ladas: we should also have here a correct STI model class, then we can remove the type fro ma parser (except the Provate Public network case)

Ladas:
@mansam awesome, just a minor nit, can be done as follow up cleanup.

About the configuration for having also the old parser/refresh in place. We have 2 providers already decided to go all in (ansible and middleware) and just not support the old refresh. TBH maintaining 2 refreshes in AWS is pain, while it is possibly a safer move. :-) Nevertheless, if you decide to go all in, then the spared time of maintaining 2 refreshes could be rather used for writing specs for this 1 refresh. :-) 

E.g. would be nice to have something like this https://github.com/Ladas/manageiq-providers-amazon/blob/aa470ae003c2ac5c106c732e67822512f46ad38c/spec/models/manageiq/providers/amazon/cloud_manager/stubbed_refresher_spec.rb#L4-L4 
to test the missing cases we have in VCR specs. And also to verify the refresh is not getting stuck somewhere for high numbers of items being refreshed (going to million of records saved total and more).

Also I am hearing from production, that the Collector is slow when having 50-100 tenants. So we should probably focus on that next. Multithreaded collection + setting :all_tenants strategy instead of cycle through tenants, where possible. Also targeted refresh will fix this issue too.

That all being said, this looks great to me and ready to be merged if you decide. 👍

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -0,0 +1,426 @@
...
+  end
+
+  def cloud_tenants
+    collector.tenants.each do |t|
+      tenant = persister.cloud_tenants.find_or_build(t.id)
+      tenant.type = "ManageIQ::Providers::Openstack::CloudManager::CloudTenant"
+      tenant.name = t.name
+      tenant.description = t.description
+      tenant.enabled = t.enabled
+      # tenant.ems_ref = t.id
```
Ladas: on the other hand, it does not hurt to assign anything. Which is especially handy when we change the manager_ref, e.g. from the :name to :ems_ref

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -0,0 +1,426 @@
...
+      o.name = param_key
+      o.value = param_val
+      o.stack = persister.orchestration_stacks.lazy_find(stack.id)
+    end
+  end
+
+  def orchestration_stack_outputs(stack)
+    raw_outputs = collector.orchestration_outputs(stack)
+    raw_outputs.each do |output|
+      uid = compose_ems_ref(stack.id, output['output_key'])
```
Ladas: you can probably do collector.orchestration_outputs(stack).each, it's -1 line :-)

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -0,0 +1,426 @@
...
+      o.stack = persister.orchestration_stacks.lazy_find(stack.id)
+
+      s = persister.vms.find_or_build(uid)
+      s.orchestration_stack = persister.orchestration_stacks.lazy_find(stack.id)
+    end
+  end
+
+  def orchestration_stack_parameters(stack)
+    raw_parameters = collector.orchestration_parameters(stack)
+    raw_parameters.each do |param_key, param_val|
```
Ladas: collector.orchestration_parameters(stack).each

app/models/manageiq/providers/openstack/inventory_collection_default/cloud_manager.rb
```diff
@@ -0,0 +1,269 @@
...
+        ]
+      }
+      super(attributes.merge!(extra_attributes))
+    end
+
+    def cloud_resource_quotas(extra_attributes = {})
+      attributes = {
+        :model_class                 => CloudResourceQuota,
+        :association                 => :cloud_resource_quotas,
+        :manager_ref                 => [:ems_ref],
```
Ladas: :manager_ref  => [:ems_ref] is a default value, so you don;t probably need to specify that

Ladas:
@mansam I've done also final detailed walk through, it looks really great. Feel free to do all of the stuff as followup PR, so we move forward. :-)

tzumainn:
Quick question - can this be updated to handle the cases fixed by https://github.com/ManageIQ/manageiq-providers-openstack/pull/1 and https://github.com/ManageIQ/manageiq-providers-openstack/pull/3?

mansam:
I've worked in a fix for the deleted flavor problem, as well as the glance image name issue linked above. I also caught and fixed a problem with orchestration stack resources referring to nonexistent instances while working on those other two.

mansam:
Depends on https://github.com/ManageIQ/manageiq/pull/14753

app/models/manageiq/providers/openstack/inventory/persister/cloud_manager.rb
```diff
@@ -40,5 +40,14 @@ def initialize_inventory_collections
         orchestration_stacks_parameters
       )
     )
+
+    add_inventory_collection(
```
Ladas: @mansam nice, best to add it for orchestration stack too, without it the Stack are doing n+1 requests.

Btw. the history of this optimization is, that before we were keeping active record object around, so .parent = obj worked fine. But AR objects take a lot of space in memory, so we ditched that and now we keep only :id, that means each .parent = obj requires fetching it from the DB again, since Ancestry requires to use AR objects and can't work with just :id, so that is why we have the custom saving methods, that will effectively load it in batches as set it.

So the root of the problem is that Ancestry is not Rails friendly, since it always require 2 AR objects to fill the parent. Hopefully this might be fixed in the future.

mansam: @Ladas thanks for explaining the history behind that optimization, it always helps me to know why things are the way they are.

app/models/manageiq/providers/openstack/inventory_collection_default/cloud_manager.rb
```diff
@@ -217,7 +217,7 @@ def orchestration_stacks_resources(extra_attributes = {})
 
     def orchestration_stacks(extra_attributes = {})
       attributes = {
-        :model_class                 => ManageIQ::Providers::Openstack::CloudManager::OrchestrationStack,
+        :model_class                 => ManageIQ::Providers::CloudManager::OrchestrationStack,
```
Ladas: this looks wrong, it has to be ManageIQ::Providers::Openstack::CloudManager::OrchestrationStack

although the parser overwrites that, we will remove the class from the parser later

Ladas:
@mansam the CI failure looks unrelated. I see only 1 issue with a wrong model_class, otherwise the additions look great.

app/models/manageiq/providers/openstack/inventory_collection_default/cloud_manager.rb
mansam: I see. I was hoping I could use the parent class so that I didn't have to have a separate entry for vnfs. ;)

mansam:
@Ladas working on writing new specs helped me catch a bunch of subtle bugs

Ladas:
@mansam awesome :-)

app/models/manageiq/providers/openstack/inventory_collection_default/cloud_manager.rb
Ladas: hm, I see. So the :type passed beats the model_class so probably leave this and pass the right class in the :type

PR #8, opened 2017-04-11
Move openstack_data to manageiq-providers-openstack repo

rwsu:
The test data and helper class are used by metrics_capture_spec.
They should live in the same repository.


rwsu:
@durandom please review :)

spec/models/manageiq/providers/openstack/cloud_manager/metrics_capture_spec.rb
```diff
@@ -1,6 +1,5 @@
 describe ManageIQ::Providers::Openstack::CloudManager::MetricsCapture do
-  # FIXME: move this helper over here
-  require Rails.root.join *%w(spec tools openstack_data openstack_data_test_helper)
+  require File.expand_path(File.join(File.dirname(__FILE__), %w(.. .. .. .. .. tools openstack_data openstack_data_test_helper)))
```
durandom: instead of this relative, I'd rather do [this](https://github.com/ManageIQ/manageiq-providers-vmware/blob/9a3097d80ce7931175bb08a99c22f3f4012c4bc7/spec/models/manageiq/providers/vmware/infra_manager/refresher_storage_profile_spec.rb#L1)

```
require ManageIQ::Providers::Openstack::Engine.root.join('spec/tools/...').to_s
```

rwsu:
@durandom, change made. Please review again.

PR #16, opened 2017-04-17
Only consider network interface when finding network metrics

rwsu:
Previously we were looking at all resources when we were finding
the network interface's resource id. We can reduce the number
of resources we inspect by limiting them to network interface
resources.

This requires a fog-openstack change to add a new request:
https://github.com/fog/fog-openstack/pull/283

This PR should be merged only when the fog-openstack change
has been released and the manageiq-gems-pending gemspec has been
updated with the release version containing that change.

### DIFF
--- a/app/models/manageiq/providers/openstack/base_metrics_capture.rb
+++ b/app/models/manageiq/providers/openstack/base_metrics_capture.rb
@@ -95,27 +95,13 @@ class ManageIQ::Providers::Openstack::BaseMetricsCapture < ManageIQ::Providers::
 
   def find_meter_counters(metric_capture_module, resource_filter, metadata_filter, log_header)
     counters = list_resource_meters(resource_filter, log_header) + list_metadata_meters(metadata_filter, log_header)
-
-    # With Gnocchi, the network metrics are not associated with the instance's resource id
-    # but with the instance's network interface resource id. Here we fetch the counters
-    # for the network interface, so that the network metrics can be fetched.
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/openstack/base_metrics_capture.rb
```diff
@@ -104,7 +104,7 @@ def find_meter_counters(metric_capture_module, resource_filter, metadata_filter,
         # fetch the list of resources and use the original_resource_id and type to find
         # the network interface's resource
         original_resource_id = "#{target.ems_ref}-tap#{port.ems_ref[0..10]}"
-        resources = @perf_ems.list_resources.body
+        resources = @perf_ems.list_resources('instance_network_interface').body
```
aufi: It looks to me, that the ```instance_network_interface``` resource type is related to cloud provider only, but it is placed in file _base_metrics_capture.rb_ which is common for both infra and cloud providers.

I know that similar thing was merged few weeks ago (line 109), but I'd suggest have some separation* of cloud and infra code (metrics in cloud and infra can have different resources types AFAIK).

\* a condition or better a method or a different file.

rwsu: Ok, let me move it out into a method.

rwsu: Ok, let me move it out into a method.

PR #18, opened 2017-04-19
Add cloud refresher tests with stubbed out collector calls

mansam:
This is a set of tests for the inventory refresher with stubbed out calls that do not rely on the presence of a VCR. This is able to catch some stuff the current refreshers specs don't catch, like incorrectly setup relationships between different objects.


petrblaho:
@mansam, please, do not forget to add host_aggregates tests so we do not forget them again :-)
see https://github.com/ManageIQ/manageiq-providers-openstack/pull/35

tzumainn:
Nice! Looks good to me!

PR #19, opened 2017-04-19
Move ci_mixin/capture tests relying on OpenStack data to this repo

rwsu:
The tests uses openstack_data_test_helper and data files under
spec/tools/openstack_data/openstack_perf_data.

Corresponding tests in main repo will be removed in:
https://github.com/ManageIQ/manageiq/pull/14720


aufi:
Looks good to me :+1:  We might want have something similar for Gnocchi based metrics in some time.

PR #26, opened 2017-04-26
Refactor node management methods

rwsu:
Clean up code by removing task_id parameters from queued
methods. 

Move host_operations_mixin.rb methods to host/operations.rb


tzumainn:
Looks good, thanks!

PR #30, opened 2017-05-09
Security Groups Rules: Create, delete tasks

gildub:
Raw commands and Tasks queues for create and delete
Adds rule parsing specific for Openstack network provider

### DIFF
--- a/app/models/manageiq/providers/openstack/network_manager/security_group.rb
+++ b/app/models/manageiq/providers/openstack/network_manager/security_group.rb
@@ -103,7 +103,7 @@ class ManageIQ::Providers::Openstack::NetworkManager::SecurityGroup < ::Security
   def raw_create_security_group_rule(security_group_id, direction, options)
     options.delete_if { |_k, v| v.nil? || v.empty? }
     ext_management_system.with_provider_connection(connection_options(cloud_tenant)) do |service|
-      service.create_security_group_rule(security_group_id, direction, options)
+      service.create_security_group_rule(security_group_id, parse_direction(direction), options)
     end
   rescue => e
     _log.error "security_group=[#{name}], error: #{e}"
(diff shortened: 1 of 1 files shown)


gildub:
@miq-bot assign @aufi

app/models/manageiq/providers/openstack/network_manager/security_group.rb
```diff
@@ -55,6 +83,42 @@ def delete_security_group_queue(userid)
...
+      :priority    => MiqQueue::HIGH_PRIORITY,
+      :role        => 'ems_operations',
+      :zone        => ext_management_system.my_zone,
+      :args        => [key]
+    }
+    MiqTask.generic_action_with_callback(task_opts, queue_opts)
+  end
+
+  def raw_create_security_group_rule(security_group_id, direction, options)
+    options.delete_if { |_k, v| v.nil? || v.empty? }
```
aufi: A nit, is the "array cleaning" needed for Fog? It doesn't look to be present in other _raw_create..._ or _raw_update..._ methods in this file.

aufi:
Thanks for the PR, generally looks good. Could you add some tests?

app/models/manageiq/providers/openstack/network_manager/security_group.rb
gildub: You're right, usually Fog doesn't care about other arguments besides the one it handles.
Meanwhile if a managed argument is empty or nil that can break the API request therefore the cleansing.
That operation cannot be done before reaching the provider because they all have different argument needs.

Also, there are actually some cleaning done in other places such as [1] and [2]

[1] https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/network_manager/security_group.rb#L21
[2] https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/network_manager/network_router.rb#L12-L13

gildub:
@aufi ,

There are no spec and no structure to support tests for any of the network providers.
We need something like `spec/models/manageiq/providers/openstack/network_manager`
I believe this is out of the scope of this patch.
Therefore I created  https://github.com/ManageIQ/manageiq-providers-openstack/issues/36 as a tracker 

If you agree, please (re)review current PR which is blocking security groups rules CRUD.

Thanks

aufi:
@gildub, thanks! Pushing tests later works for me.

PR #32, opened 2017-05-11
Change Openstack::InfraManager model to use real model for subclass

gmcculloug:
The Openstack::InfraManager model was using [alias EmsInfra](https://github.com/ManageIQ/manageiq/blob/master/app/models/aliases/ems_infra.rb) instead of the real subclass `ManageIQ::Providers::InfraManager`

Found while debugging issue https://github.com/ManageIQ/manageiq-automation_engine/issues/20 which is fixed by a similar change to the service model https://github.com/ManageIQ/manageiq-automation_engine/pull/22.


agrare:
@gmcculloug looks good to me but I'm not a merger here :)
cc @aufi @tzumainn

tzumainn:
@gmcculloug Just doing a sanity check, then I'll merge - thanks!

tzumainn:
Looks good, merging!

PR #37, opened 2017-05-18
Adds security group tests

gildub:
Adds `spec/models/manageiq/providers/openstack/network_manager/`
Adds security groups spec file with security group and firewall rules CRUD.

### DIFF
--- a/app/models/manageiq/providers/openstack/network_manager/security_group.rb
+++ b/app/models/manageiq/providers/openstack/network_manager/security_group.rb
@@ -34,7 +34,6 @@ class ManageIQ::Providers::Openstack::NetworkManager::SecurityGroup < ::Security
     ext_management_system.with_provider_connection(connection_options(cloud_tenant)) do |service|
       security_group = service.create_security_group(options).body
     end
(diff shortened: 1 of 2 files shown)


gildub:
@ladas, 

I can't get pass the `:connect`, seems like there is something happening between the cloud and network providers.

Maybe you'd have a clue.

Cheers,
Gilles

gildub:
@miq-bot assign @Ladas

Ladas:
@gildub hm, try to use ems.network_manager. You might need that for the find I think. The connect could be mocked on either, since network manager delegates that call to the cloud_manager. But all network models are associated with the network manager.

gildub:
@Ladas, I figured it out, I was using ems.network_manager but used the instance instead of the class :(
Thanks.

gildub:
@aufi, @Ladas could you please review?

Ladas:
looks great 👍

aufi:
Yep, looks great to me too, ready to merge when WIP label is removed.

gildub:
@Ladas, @aufi, Thanks.

PR #38, opened 2017-05-18
Implement Openstack Swift/Cinder Graph Refresh

mansam:
This PR implements graph refresh for Swift and Cinder. It sidesteps the issue of Cinder and Swift living in the core repo but needing Openstack specific refresher implementation by subclassing those managers.

### DIFF
--- a/app/models/manageiq/providers/openstack/cinder_manager.rb
+++ b/app/models/manageiq/providers/openstack/cinder_manager.rb
@@ -1,0 +1,2 @@
+class ManageIQ::Providers::Openstack::CinderManager < ManageIQ::Providers::StorageManager::CinderManager
+end
--- a/app/models/manageiq/providers/openstack/cinder_manager_mixin.rb
+++ b/app/models/manageiq/providers/openstack/cinder_manager_mixin.rb
@@ -1,0 +1,20 @@
+module ManageIQ::Providers::Openstack::CinderManagerMixin
+  extend ActiveSupport::Concern
+  include ::CinderManagerMixin
+
+  included do
+    # TODO: how about many storage managers???
+    # Should use has_many :storage_managers,
+    has_one  :cinder_manager,
+             :foreign_key => :parent_ems_id,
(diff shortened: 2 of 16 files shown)


tzumainn:
@roliveri we're updating the refresh code to use the new graph refresh pattern, and would like to update all the OpenStack related code - however, this change would essentially change the OpenStack storage managers to reside in this repository  - @mansam correct me if I'm wrong.

I'm guessing you may have opinions on this, what do you think?

tzumainn:
:+1: from a functionality point of view!

mansam:
@tzumainn The actual implementation is still in the core repository. They're subclassed here to allow us to use graph refresh with them.

roliveri:
@tzumainn While the Cinder and Swift storage managers are used by OpenStack, they can also be used independently and in other environments. For example, the Cinder manager can be used by RHEV (at least I've been told this is a goal) and Swift can be used independently via Swiftstack. As such these managers should reside in any provider-specific repository.

roliveri:
@mansam As I commented above, the code isn't OpenStack specific and not only shouldn't it be in the OpenStack repo, it shouldn't be subclassed under OpenStack.

mansam:
Alright, thanks for clarifying that @roliveri. Going to close this for now.

roliveri:
@mansam We probably need to discuss this case - where code can be used by multiple provider types to refresh portions of their inventory, or standalone.

While standalone versions, like SwiftStack, will have their own refresh, when tied to specific provider types, like OpenStack, their refresh should probably be tied to the refresh of that provider.

mansam:
I agree, the case for Cinder/Swift as part of another provider is separate from the standalone case. It was my goal with this PR to take care of that case w/r/t the Openstack Cloud Provider while leaving the existing standalone code alone.

roliveri:
@mansam Yes, but it looks like you've duplicated some code that should be common. I would think that all of the parsing code should be common, and a good portion of the persister code should be as well. By it's nature, the cross linking code needs to be provider type (OpenStack) specific, but that portion of the code was already factored out.

PR #39, opened 2017-05-22
Security Groups Inventory fix

gildub:
Populates source_security_group_id

https://github.com/ManageIQ/manageiq/issues/15030


Ladas:
👍 awesome, a spec would be nice too :-)

tzumainn:
The security group specs are coming en masse in a separate PR, I believe (https://github.com/ManageIQ/manageiq-providers-openstack/pull/37)

PR #40, opened 2017-05-22
Edit delete method for Cloud Subnet

andyvesel:
Changed `delete_cloud_subnet_queue` method's body to call `delete_cloud_subnet` directly instead of `raw_`.
https://github.com/ManageIQ/manageiq/pull/15087
https://github.com/ManageIQ/manageiq-ui-classic/pull/1347


andyvesel:
@gildub could you review?

aufi:
@andyvesel I was not sure if MiqTask method called from Openstack-specific class should call raw_... method or not (my personal prefference is to call raw_... method here), but it looks in the code, that calling non-raw method is more common. Let's wait on @gildub opinion since he wrote that part of model code.

andyvesel:
@aufi sure, and I'll check it out once again

PR #41, opened 2017-05-22
Vcr recorder rake tasks

petrblaho:
This allows developers to use one rake task `bundle exec rake app:manageiq:providers:openstack:vcr:rerecord` to re-record VCR cassettes.


mansam:
:+1:

PR #42, opened 2017-05-23
Edit update method for cloud subnet

andyvesel:
Changed `update_cloud_subnet_queue` method's body to call `update_cloud_subnet` directly instead of `raw_`
Depends on https://github.com/ManageIQ/manageiq/pull/15192


aufi:
@andyvesel Hi, this PR looks pretty same as #40, would you consider closing it?

andyvesel:
Closing. As we discussed [here](https://github.com/ManageIQ/manageiq-providers-openstack/pull/40) We are no longer need it

PR #46, opened 2017-05-26
Add fw rule's remote group id for create/update

gildub:
We can use source_security_group_id now that we have it in the inventory:

https://github.com/ManageIQ/manageiq-providers-openstack/pull/39

### DIFF
--- a/app/models/manageiq/providers/openstack/network_manager/security_group.rb
+++ b/app/models/manageiq/providers/openstack/network_manager/security_group.rb
@@ -18,12 +18,15 @@ class ManageIQ::Providers::Openstack::NetworkManager::SecurityGroup < ::Security
   end
 
   def self.parse_security_group_rule(rule)
+    if rule["source_security_group_id"] && !rule["source_security_group_id"].empty?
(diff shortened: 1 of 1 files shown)


gildub:
@miq-bot assign @aufi

app/models/manageiq/providers/openstack/network_manager/security_group.rb
```diff
@@ -18,12 +18,15 @@ class ManageIQ::Providers::Openstack::NetworkManager::SecurityGroup < ::Security
   end
 
   def self.parse_security_group_rule(rule)
+    if rule["source_security_group_id"] && !rule["source_security_group_id"].empty?
```
aufi: Nit: Couldn't the condition be written with Rails method ```present?```

gildub: Ah yeah, I've not used it much but I'll. Thanks

gildub:
Needed by https://github.com/ManageIQ/manageiq-ui-classic/pull/1011

PR #54, opened 2017-06-10
Add execute_ and has_? workflow methods

rwsu:
Adding these methods provide a standard set of interfaces to
interact with the Workflow service. It will also allow us to remove
some duplicate code in Host.

### DIFF
--- a/lib/manageiq/providers/openstack/legacy/openstack_handle/workflow_delegate.rb
+++ b/lib/manageiq/providers/openstack/legacy/openstack_handle/workflow_delegate.rb
@@ -3,7 +3,7 @@ module OpenstackHandle
     include OpenstackHandle::HandledList
     include Vmdb::Logging
 
-    SERVICE_NAME = "Workflow"
+    SERVICE_NAME = "Workflow".freeze
 
     attr_reader :name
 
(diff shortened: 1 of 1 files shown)


lib/manageiq/providers/openstack/legacy/openstack_handle/workflow_delegate.rb
```diff
@@ -12,5 +12,62 @@ def initialize(dobj, os_handle, name)
...
+      [workflow_execution_id, workflow_state, response]
+    end
+
+    def has_workflow?(workflow_name)
+      result = false
+      @os_handle.service_for_each_accessible_tenant(SERVICE_NAME) do |svc|
+        begin
+          response = svc.get_workflow(workflow_name)
+          result = true if response.status == 200
+        rescue => err
```
chessbyte: Would this be better as `return true if response.status == 200`?

rwsu: Sure :)

tzumainn:
Would it make sense to include the ManageIQ task queue as part of the workflow execution?  I can't imagine we'd ever execute a workflow outside of a task.  Or do you think that should remain separate?

rwsu:
@tzumainn, I would lean towards not including the task queue as part of workflow execution. After successful execution, we currently kick off an ems refresh. Having the task queue as part of workflow execution would mean that we would need to pass a reference to the ems as a parameter to the workflow execution so that it kicks off the refresh if successful. Doing so makes the workflow delegate do more than it should. 

And we can't queue the ems refresh at the stack level because once the execution is queued, the stack doesn't have any visibility to the execution result,

tzumainn:
@rwsu Fair enough!  The code looks good to me; I'll merge once tests pass.

aufi:
Looks good to me!

PR #55, opened 2017-06-10
Add scale_down and scale_up tasks to OrchestrationStack

rwsu:
These tasks execute Mistral workflows instead of invoking stack_update.

Depends on #54


aufi:
Looks good to me (but I didn't test it).

tzumainn:
Question - this just adds these methods, but doesn't yet allow the UI to call them, correct? Or am I missing something?

rwsu:
@tzumainn that's correct. There is a separate PR that modifies the UI to use these methods: https://github.com/ManageIQ/manageiq-ui-classic/pull/1533

PR #56, opened 2017-06-19
Network Router: More generic use of SupportsFeatureMixin

gildub:
Migrated from https://github.com/ManageIQ/manageiq/pull/13073

### DIFF
--- a/app/models/manageiq/providers/openstack/network_manager/network_router.rb
+++ b/app/models/manageiq/providers/openstack/network_manager/network_router.rb
@@ -2,10 +2,26 @@ class ManageIQ::Providers::Openstack::NetworkManager::NetworkRouter < ::NetworkR
   include ProviderObjectMixin
   include AsyncDeleteMixin
 
-  supports :create_network_router
-  supports :delete_network_router
-  supports :update_network_router
   supports :add_interface
+
+  supports :create
+
+  supports :delete do
+    if ext_management_system.nil?
+      unsupported_reason_add(:delete, _("The Network Router is not connected to an active %{table}") % {
+        :table => ui_lookup(:table => "ext_management_systems")
+      })
+    end
+  end
+
+  supports :update do
+    if ext_management_system.nil?
+      unsupported_reason_add(:update, _("The Network Router is not connected to an active %{table}") % {
+        :table => ui_lookup(:table => "ext_management_systems")
+      })
+    end
+  end
+
   supports :remove_interface
 
   def self.raw_create_network_router(ext_management_system, options)


app/models/manageiq/providers/openstack/network_manager/network_router.rb
```diff
@@ -2,10 +2,26 @@ class ManageIQ::Providers::Openstack::NetworkManager::NetworkRouter < ::NetworkR
...
-  supports :create_network_router
-  supports :delete_network_router
-  supports :update_network_router
   supports :add_interface
+
+  supports :create
+
+  supports :delete do
+    if ext_management_system.nil?
+      unsupported_reason_add(:delete_network_router, _("The Network Router is not connected to an active %{table}") % {
```
durandom: the reason should match the one in the `supports` block.
` unsupported_reason_add(:delete ...`

app/models/manageiq/providers/openstack/network_manager/network_router.rb
```diff
@@ -2,10 +2,26 @@ class ManageIQ::Providers::Openstack::NetworkManager::NetworkRouter < ::NetworkR
...
+        :table => ui_lookup(:table => "ext_management_systems")
+      })
+    end
+  end
+
+  supports :update do
+    if ext_management_system.nil?
+      unsupported_reason_add(:update_network_router, _("The Network Router is not connected to an active %{table}") % {
```
durandom: same here

app/models/manageiq/providers/openstack/network_manager/network_router.rb
gildub: Absolutely

app/models/manageiq/providers/openstack/network_manager/network_router.rb
gildub: Same

gildub:
@durandom, to be consistent with other network item I'm just using the generic action.

tzumainn:
Is there an associated PR to manageiq-ui-classic?  I think the buttons there check for things like :create_network_router.

aufi:
Tested, I wasn't able see error (router disappeared together with network manager deletion), but anyway merging should not cause errors, trust @durandom ;-)

gildub:
@tzumainn, 

You're right we need to clean up the associated declaration, I've created 
https://github.com/ManageIQ/manageiq/pull/15451 to take care of it.

Thanks

gildub:
@aufi, thanks for the confirmation.

btw, would you mind detailing here or by email more about the "router disappeared together with network manager" ?

aufi:
@gildub Sure, I understood this change as a check if Router is associated with provider (network manager) and if not, disable update, delete action because the action couldn't be completed without valid provider.

The only case when we have an object in MIQ database which doesn't belong to any existing provider I'm aware of is case when we _delete the provider_, but some of its inventory remain in MIQ database (e.g. Host in Infra provider). I expected that Router has similar behaviour, so it stays in MIQ database event its provider (network manager) was deleted and change in this PR ensures, that such router cannot be edited or deleted.

I might be wrong and checking of valid provider is just a convention in ManageIQ providers or feel free correct me, thanks :-)

gildub:
@aufi, 

I see, well that's an interesting behaviour, I'd assume the dependent objects of a provider to clear up (cascading delete) no matter what, if that's not the case then we are in the *wild* ;).

aufi:
@gildub Yeah, in most cases there is the cascade delete, no problem.

mansam:
@miq-bot add_label fine/yes

gildub:
Fine branch is taking place within manageiq repo. The cherry pick for this is patch has been created as https://github.com/ManageIQ/manageiq/pull/17660

gildub:
Never mind my previous comment as the cherry pick can be done across.

simaishi:
Fine backport (to manageiq repo) details:

```
$ git log -1
commit 9cfc0ea4c6da05bd928576d6984e1662ed9e568f
Date:   Mon Jun 26 11:58:30 2017 +0200

    Merge pull request #56 from gildub/network_controller_network_router_fix_support_feature
    
    Network Router: More generic use of SupportsFeatureMixin
    (cherry picked from commit 71520a1bced4e606b6703631cf40093044ef1838)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1593349
    https://bugzilla.redhat.com/show_bug.cgi?id=1593353
```

simaishi:
@gildub Can you please take a look at manageiq-ui-classic Travis failure for Fine branch?
https://travis-ci.org/ManageIQ/manageiq-ui-classic/jobs/421223743

I suspect this PR is causing the failure...

```
  1) ApplicationHelper::Button::NetworkRouterNew#disabled? when at least one provider supports network router create then the button is not disabled
     Failure/Error: expect(button.disabled?).to be false
     
       expected false
            got true
     # ./spec/helpers/application_helper/buttons/network_router_new_spec.rb:7:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:75:in `block (3 levels) in <top (required)>'
     # ./spec/manageiq/spec/support/evm_spec_helper.rb:34:in `clear_caches'
     # ./spec/spec_helper.rb:75:in `block (2 levels) in <top (required)>'
  2) ApplicationHelper::Button::NetworkRouterNew#calculate_properties when at least one provider supports network router create, the button has no error in the title
     Failure/Error: expect(button[:title]).to be nil
     
       expected nil
            got #<String:474129520> => "No cloud providers support creating network routers."
     # ./spec/helpers/application_helper/buttons/network_router_new_spec.rb:30:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:75:in `block (3 levels) in <top (required)>'
     # ./spec/manageiq/spec/support/evm_spec_helper.rb:34:in `clear_caches'
     # ./spec/spec_helper.rb:75:in `block (2 levels) in <top (required)>'
```

simaishi:
@gildub ping ^

gildub:
@satoe,

I'll look at it as soon as possible. BTW don't we need to restore fine/yes flag?

simaishi:
@gildub Thanks. Since this PR is backported, it can stay as `fine/backported`. Please reference this PR in your PR that will fix the Travis failure though.

PR #57, opened 2017-06-20
Add fast refresh options

mansam:
If the `ems_openstack.refresh.is_admin` configuration option is set, then ManageIQ will attempt to use admin-only api options to significantly reduce the number of repeated API calls necessary to build the Openstack inventory. This dramatically speeds up the refresh in environments with lots of tenants. (from 9ish minutes in an environment with 100+ tenants, to less than 4 minutes)

Due to differences in the way Heat's access control system works, there is a separate flag to enable admin mode and the associated refresh speedups. This requires Heat to be configured in a particular way, but should further speed up the refresh process.
See https://bugs.launchpad.net/heat/+bug/1466694 and https://bugs.launchpad.net/keystone/+bug/968696 for more details on why this is the case.


lib/manageiq/providers/openstack/legacy/openstack_handle/handled_list.rb
```diff
@@ -15,7 +16,12 @@ def handled_list(collection_type, options = {})
       # @orchestration_service.handled_list(:resources, :stack => stack)
       # By default, it calls :all method on the Fog collection, that has unified interface in all list methods in Fog
       # and always returns detailed list.
-      multi_tenancy_class.new(self, @os_handle, self.class::SERVICE_NAME, collection_type, options,
+      multi_tenancy_class_ = if all_tenants
+                               OpenstackHandle::MultiTenancy::Option
+                             else
+                               multi_tenancy_class
+                             end
```
tzumainn: The extra _ in the variable seems a bit weird to me - maybe call it 'multi_tenancy_class' and rename the function to 'default_multi_tenancy_class' or something?

app/models/manageiq/providers/openstack/inventory/collector/cloud_manager.rb
```diff
@@ -103,33 +115,47 @@ def quotas
...
+      @tenants ||= identity_service.visible_tenants.select do |t|
+        # avoid 401 Unauth errors when checking for accessible tenants
+        # the "services" tenant is a special tenant in openstack reserved
+        # specifically for the various services
+        next if t.name == "services"
+        true
+      end
+    else
+      @tenants ||= manager.openstack_handle.accessible_tenants
+    end
```
tzumainn: Is there a clean way to put the weird 'services' filter into one place code-wise, or is that not really a practical option?

tzumainn:
Thanks!  I put in one or two minor comments/questions.  From an overall perspective: is it reasonable to update the 'classic' refresher code in this PR for backport purposes, or do you think that's better limited to a backport-specific PR?

lib/manageiq/providers/openstack/legacy/openstack_handle/handled_list.rb
mansam: Good call.

app/models/manageiq/providers/openstack/inventory/collector/cloud_manager.rb
mansam: I'll see if I can come up with a way to factor out the repetition cleanly.

mansam:
I think it would be reasonable to do it here.

app/models/manageiq/providers/openstack/network_manager/refresh_parser.rb
```diff
@@ -64,23 +64,43 @@ def parent_manager_find_device_connected_to_network_port(ems_ref)
...
+      end
     end
 
     def floating_ips
-      @network_service.handled_list(:floating_ips)
+      if @ems.kind_of?(ManageIQ::Providers::Openstack::CloudManager) && ::Settings.ems.ems_openstack.refresh.is_admin
+        @floating_ips ||= @network_service.handled_list(:floating_ips, {}, true)
+      else
+        @floating_ips ||= @network_service.handled_list(:floating_ips)
+      end
```
tzumainn: Is it possible to put this code pattern into its own private method, where you pass in :floating_ips or :routers or ... ?

tzumainn:
One more minor comment; I'll test this code out now though.  Thanks!

tzumainn:
Tests look good!  Merging once CI passes

simaishi:
@mansam @tzumainn Can this be `fine/yes` and `euwe/yes` as per flags set on the BZ?

mansam:
@simaishi Portions of this PR apply only to the new inventory object refresh system, but the speedups to the old refresher could be backported to Euwe and Fine.

simaishi:
@mansam in that case... can you please create a PR for Fine/Euwe branch for the applicable changes?

mansam:
@simaishi Will do.

simaishi:
Backported to Euwe via https://github.com/ManageIQ/manageiq/pull/15538

simaishi:
Backported to Fine via https://github.com/ManageIQ/manageiq-gems-pending/pull/224 and https://github.com/ManageIQ/manageiq/pull/15539

PR #58, opened 2017-06-27
allow raw_connect to accept encrypted passwords

jntullo:
Working on switching credential validation onto the queue using `raw_connect`. This will require passwords to be encrypted / decrypted.

See https://github.com/ManageIQ/manageiq-ui-classic/pull/1580 for related info / BZ

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ gemspec
 # Git. Remember to move these dependencies to your gemspec before releasing
 # your gem to rubygems.org.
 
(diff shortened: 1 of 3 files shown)


aufi:
Thanks for the PR. I'd suggest put a comment about using ```MiqPassword``` into https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/Gemfile#L11 since ```MiqPassword``` comes from manageiq-gems-pending repo (which we want drop from dependencies in future).

Otherwise looks good to me.

PR #59, opened 2017-07-07
NetworkManager: SupportsFeatureMixin cleanup

gildub:
Removes `create_network_router` which is redundant with `create`.


gildub:
@tzumainn, @aufi, could you please assign and/or review? Thanks

tzumainn:
Thanks for this update! I've verified that the UI no longer references :create_network_router; it's still in supports_feature_mixin.rb in the core repo, though. Is the plan to remove this first, and then create a PR to update the core repo?

gildub:
@tzumainn,  yes that's right, this is the related work on the [mixin side](https://github.com/ManageIQ/manageiq-providers-openstack/pull/59) of managiq repo, so we have to remove the caller(s) first.

PR #60, opened 2017-07-11
Use an unscoped connection to Keystone to validate API creds

mansam:
`verify_api_credentials` was checking Openstack API credentials
against the Compute service without an explicit tenant name,
which lead to OpenstackHandle making a ton of connections to
decide on how to scope the request. As a result, verifying
Openstack credentials was extremely slow. Now it makes a single
request to Keystone, which is way faster.

This PR is a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1469306

@tzumainn

### DIFF
--- a/app/models/manageiq/providers/openstack/manager_mixin.rb
+++ b/app/models/manageiq/providers/openstack/manager_mixin.rb
@@ -141,7 +141,7 @@ module ManageIQ::Providers::Openstack::ManagerMixin
   end
 
   def verify_api_credentials(options = {})
-    options[:service] = "Compute"
+    options[:service] = "Identity"
     with_provider_connection(options) {}
     true
   rescue => err


mansam:
The test failure related to AuthKeyPairs is from a change in manageiq core and is unrelated to this change.

tzumainn:
Yep, the failure is related to https://github.com/ManageIQ/manageiq/pull/15485

app/models/manageiq/providers/openstack/manager_mixin.rb
```diff
@@ -141,7 +141,7 @@ def translate_exception(err)
   end
 
   def verify_api_credentials(options = {})
-    options[:service] = "Compute"
+    options[:service] = "Identity"
```
aufi: The compute service was needed since multiregion support. Multiregion OpenStack deployments can share Identity service, but can have multiple Compute (or other) services in various regions. So it make sense to test region-scoped Compute service instead of shared Indetity service.

Could the PR be updated to use Compute service with a default tenant or some different technique?

mansam: Yeah, that solution should work as well.

mansam: https://github.com/ManageIQ/manageiq-providers-openstack/pull/62 actually makes this PR unnecessary.

mansam:
This pull request is superseded by https://github.com/ManageIQ/manageiq-providers-openstack/pull/62

PR #61, opened 2017-07-11
Update keypair spec test

tzumainn:
Fixes issue caused by changes in https://github.com/ManageIQ/manageiq/pull/15485


aufi:
Looks good to me, thanks for the fix.

PR #67, opened 2017-07-19
run provider generator to sync changes from core

durandom:
ran the generator again with

```bash
bundle exec rails g provider <name> --path ../
```

these are the changes.


aufi:
Looks good to me!

PR #68, opened 2017-07-19
Move vcr tasks to private rake tasks

durandom:
These tasks should not show up in manageiq core
I moved them to the the private rake tasks directory
and adjusted the readme


aufi:
Looks good to me!

Issue #70, opened 2017-07-26
openstack provider failed to do status update


tjyang:
* On our miq02 which is running latest code from master. The openstack provider failed to do status refresh. The connected openstack is Ubuntu openstack added as Cloud provider.
* Same openstack farm connected to Fine-2 miq fine without refresh issue.

```
[root@miq02 vmdb]# git log |head
commit b4554f7ed13368e6985d993ca8066d12a02d03a3
Merge: 56f639e 750d3c4
Author: Jason Frey <fryguy9@gmail.com>
Date:   Tue Jul 25 22:23:26 2017 -0400

    Merge pull request #15393 from zeari/unify_container_definition

    Merge container_definition and container

commit 56f639e6ed4f8e2fcef1f52e51466f891bf7305c
[root@miq02 vmdb]#
```
* Screenshot

![image](https://user-images.githubusercontent.com/378638/28637426-16e8ea9a-7207-11e7-8a77-533c4a9dd8a9.png)

*  EVM Log file around the error message.
```
<snipped>
[----] E, [2017-07-26T14:15:36.192110 #5857:c6f138] ERROR -- : MIQ(ManageIQ::Providers::Openstack::CloudManager::Refresher#refresh) EMS: [OpenStack_POC], id: [15] Refresh failed
[----] E, [2017-07-26T14:15:36.192385 #5857:c6f138] ERROR -- : [ActiveRecord::AssociationTypeMismatch]: Host(#118366160) expected, got NilClass(#13131780)  Method:[block in method_missing]
[----] E, [2017-07-26T14:15:36.192494 #5857:c6f138] ERROR -- : /opt/rubies/ruby-2.3.1/lib/ruby/gems/2.3.0/gems/activerecord-5.0.4/lib/active_record/associations/association.rb:231:in `raise_on_type_mismatch!'
/opt/rubies/ruby-2.3.1/lib/ruby/gems/2.3.0/gems/activerecord-5.0.4/lib/active_record/associations/collection_association.rb:382:in `block in replace'
/opt/rubies/ruby-2.3.1/lib/ruby/gems/2.3.0/gems/activerecord-5.0.4/lib/active_record/associations/collection_association.rb:382:in `each'
/opt/rubies/ruby-2.3.1/lib/ruby/gems/2.3.0/gems/activerecord-5.0.4/lib/active_record/associations/collection_association.rb:382:in `replace'
[...]

PR #71, opened 2017-07-27
Assign only compact and unique list of hosts

Ladas:
Assign only compact and unique list of hosts, to avoid rails
failures or duplication of the data.

Fixes:
https://github.com/ManageIQ/manageiq-providers-openstack/issues/70


Ladas:
@mansam btw. I am currently trying to get rid of these array assignments, since it prevents from using Batch SQL saver_strategy. So the most performant way is to model every DB table as an InventoryCollection.

aufi:
Looks good to me :+1:

tzumainn:
Looks good! Merging.

PR #73, opened 2017-07-31
Handle case where do_volume_creation_check gets a nil from Fog

mansam:
`do_volume_creation_check` assumes it will always get the requested volume object back from Fog, resulting in a thrown exception and an errored-out provisioning attempt if that assumption is violated due to the requested volume not being found. This fixes the assumption, allowing the provisioning workflow to continue and retry as normal.

This fixes the surface level issue in https://bugzilla.redhat.com/show_bug.cgi?id=1475260, though it doesn't explain why a 404 would be reported for an ID that came from Fog in the first place.


tzumainn:
Looks good!

simaishi:
Fine backport (to manageiq repo) details:

```
$ git log -1
commit 3bf47be83a48e61b4e6b24485c596caee6ce7012
Author: tzumainn <tzumainn@redhat.com>
Date:   Tue Aug 1 16:14:45 2017 -0400

    Merge pull request #73 from mansam/dont-crash-provision-if-volume-not-found
    
    Handle case where do_volume_creation_check gets a nil from Fog
    (cherry picked from commit 752420bc07b2b4f606be119656e2eccc4b297374)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1482170
```

PR #74, opened 2017-08-01
Targeted Refresh for Cloud VMs

mansam:
This pull request introduces the scaffold for targeted refresh and implements it for cloud VMs. This is based significantly on @Ladas's work on the Amazon provider.


tzumainn:
@Ladas could you help us take a look at this? Thanks!

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
```diff
@@ -0,0 +1,212 @@
...
+    return [] if references(:availability_zones).blank?
+    @availability_zones = references(:availability_zones).collect do |az|
+      OpenStruct.new(:zoneName => az)
+    end
+  end
+
+  def cloud_networks
+    return [] if references(:cloud_networks).blank?
+    @cloud_networks = references(:cloud_networks).collect do |network_id|
+      safe_get { network_service.networks.get(network_id) }
```
Ladas: hm, could we use a filtered handled list? OpenStack should support list filtering by UUIDs right? In larger envs and e.g. bigger stack deployment, we can have hundreds of entities here. So that would a lot of API calls

spec/models/manageiq/providers/openstack/cloud_manager/refresh_spec_common.rb
```diff
@@ -863,5 +863,107 @@ def expect_sync_cloud_tenants_with_tenants_is_queued
       expect(sync_cloud_tenant.method_name).to eq("sync_cloud_tenants_with_tenants")
       expect(sync_cloud_tenant.state).to eq(MiqQueue::STATE_READY)
     end
+
+    def assert_targeted_vm(vm_name, attributes)
```
Ladas: would be nice to have a test using existing Vm asserts, I do that in AWS to make sure Vm is the same from targeted and full refresh

Ladas:
@mansam it's great, just 1 concern with get vs filtered list API

For the specs, we should add moooore in next PRs, so every possible target has at least some coverage. :-D

In aws I have https://github.com/Ladas/manageiq-providers-amazon/blob/cd02dc09e805424ef90cf30f082f710000c09319/spec/models/manageiq/providers/amazon/cloud_manager/vcr_specs/targeted_refresh_spec.rb#L32
and I would like to add much more later

spec/models/manageiq/providers/openstack/cloud_manager/refresh_spec_common.rb
mansam: I'm not using exactly the same, since it seemed like it was not possible to acquire the external network information from the information available from just the VM. It seemed like it would take basically a full pull of inventory from Neutron to get some of it, so I left the external network stuff out of the targeted refresh and dropped the external network checks from the test.

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
mansam: Are you thinking of making a call to pull the whole inventory and then filtering that, or is there actually API support for requesting a set of UUIDs?

spec/models/manageiq/providers/openstack/cloud_manager/refresh_spec_common.rb
Ladas: Right so what I do in AWS is a full VM refresh now, so I expand more targets out of the Vm, so the result is a Vm with all network connections(IPs) and volumes.

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
Ladas: Yeah we should be able to filter N records from the API, to avoid N+1 API requests. E.g. check server list https://developer.openstack.org/api-ref/compute/ , the parameter uuid can be used to filter by the instance uuid. And we should be able to provide more of them I think

Ladas: There is API docs mentioning IN query. But also that nova does not support OR filter. https://specs.openstack.org/openstack/api-wg/guidelines/pagination_filter_sort.html#filtering

But seem like these are just some guidelines, so not sure who actually follows them. :-)

Ladas: I suppose we can merge this as a first version and optimize it where possible later. If we can't get rid of the N+1 queries, we will need to put back the condition, that will run a full refresh based on the threshold. (Best placement might be the Inventory builder)

Ladas:
@mansam just call rubocop -a on the files rubocop complains about and I am ok with merging this. Lets first build a working Targeted refresh, then optimize it. :-)

spec/models/manageiq/providers/openstack/cloud_manager/refresh_spec_common.rb
mansam: Sorry, what do you mean by "full VM refresh"? I'm not sure if you're referring to a different sort of approach, or just using a different name. :)

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
mansam: Oh, that is awesome. I didn't know it was possible to query like that. Should be easy enough to figure out what's actually supported and fix this PR to get rid of the N+1 requests.

spec/models/manageiq/providers/openstack/cloud_manager/refresh_spec_common.rb
Ladas: So for AWS, every time there is a reference to Vm (aka. Vm target), we unpack the list of references by looking at the DB:

https://github.com/Ladas/manageiq-providers-amazon/blob/6938025f2cd6a4b0504909a1be22a87e30dbc7c0/app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb#L209

and by looking at the API:
https://github.com/Ladas/manageiq-providers-amazon/blob/6938025f2cd6a4b0504909a1be22a87e30dbc7c0/app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb#L229

So we basically send more Targetsto the refresh, by scanning edges of the Vm.

-------

Then the Vm full refresh means that after 1 targeted refresh, you will get a full info about the Vm. Where the most important relations are probably IP addresses and volumes. Then it's easier to use this e.g. in a automate workflow, where 1 step is the sync refresh of the Vm and you want to work with the data in the next step.

---

Also this solves the fact that events might come in a wrong order, then the graph refresh will not be able to populate some edges. (for the 5.0 release, we will be solving these by a skeletal pre-create and a post processing where the missing edges will be planned for the next refresh)

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
Ladas: Yeah, lets do it in the next PR, you will probably need to experiment. I fear the OpenStack API is still heavily inconsistent at this. :-(

mansam: Will do, thanks for the advice.

spec/models/manageiq/providers/openstack/cloud_manager/refresh_spec_common.rb
mansam: Oh, I think I see what you mean. "Targeted VM Refresh" just being the VM object itself, and "Full Targeted VM Refresh" being the VM and its relationships. Makes sense now. I'm refreshing the VM and its relationships, but it seems like not all the relevant information to populate the network stuff was available. (Since you only get the IP address not it's ID, and you don't have enough port information to reconstruct the whole network just from the VM and its relationships due to objects from the Openstack API not having references in both directions). Though if Neutron supports filtered querying like you mentioned, maybe I can get around that. I was working on this before 2 weeks of PTO so my memory is fuzzy about the exact nature of the problem.

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
mansam: This is definitely going to be more tricky than hoped, Nova doesn't seem to respect the "in:" directive.

Ladas: yeah, I wonder which service will do :-)

spec/models/manageiq/providers/openstack/cloud_manager/refresh_spec_common.rb
Ladas: it will require some tweaking, yo ucan also do more API queries to fetch e.g. the network_ports, given you will be able to filter that API using e.g. a Vm uuid

tzumainn:
@mansam are any of the codeclimate issues relevant?

mansam:
@tzumainn I've fixed the only ones that were relevant, so this should be ok to merge once the tests go green again.

PR #76, opened 2017-08-10
Prepare to move OrchestrationTemplate to each provider

bzwei:
Move the template classes to the corresponding provider repository.
This is a step before the migration.

https://bugzilla.redhat.com/show_bug.cgi?id=1417021

This is a refactor work but a step planned to fix above bugzilla issue.

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/orchestration_template.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/orchestration_template.rb
@@ -1,0 +1,131 @@
+class ManageIQ::Providers::Openstack::CloudManager::OrchestrationTemplate < ::OrchestrationTemplate
+  def parameter_groups
+    content_hash = YAML.load(content)
+    raw_groups = content_hash["parameter_groups"]
+
+    if raw_groups
+      indexed_parameters = parameters(content_hash).index_by(&:name)
+      raw_groups.collect do |raw|
+        OrchestrationTemplate::OrchestrationParameterGroup.new(
+          :label       => raw["label"],
+          :description => raw["description"],
+          # Map each parameter name to its corresponding object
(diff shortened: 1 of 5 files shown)


app/models/manageiq/providers/openstack/cloud_manager/orchestration_template.rb
```diff
@@ -0,0 +1,131 @@
+class ManageIQ::Providers::Openstack::CloudManager::OrchestrationTemplate < ::OrchestrationTemplate
```
bzwei: This is a class moved from core repository https://github.com/ManageIQ/manageiq/blob/master/app/models/orchestration_template_hot.rb#L1

The old file will be removed after the migration is done.

app/models/manageiq/providers/openstack/cloud_manager/vnfd_template.rb
```diff
@@ -0,0 +1,69 @@
+class ManageIQ::Providers::Openstack::CloudManager::VnfdTemplate < ::OrchestrationTemplate
```
bzwei: Move from https://github.com/ManageIQ/manageiq/blob/master/app/models/orchestration_template_vnfd.rb#L1
The original file will be removed after migration is done.

@Ladas Should this class under `CloudManager` or `InfraManager`?

Ladas: it should be under a cloud manager

bzwei:
Will not fix the codeclimate issue here, since the task is simply to relocate the file.

bzwei:
@miq-bot add_label refactoring
@miq-bot assign @Ladas

Ladas:
ping @aufi can we merge this? The orchestration provisioning is in a half broken state, so we need to move forward with this. :-)

PR #77, opened 2017-08-14
Update miq-module dependency to more_core_extensions

aufi:
Miq-module was moved from gems-pending repo to more_core_extensions
repo, updating require loading.


tzumainn:
Looks good!

PR #78, opened 2017-08-16
Adding initial changelog with sprint66 info

JPrause:


tzumainn:
@JPrause thanks for adding this! is this something that someone on the openstack provider team should be doing in the future?

JPrause:
@tzumainn ideally yes. However, I'm starting to add changelogs to all repos that don't have one. I'll also be maintaining them so I can take care of this one as well.

JPrause:
@miq-bot add_label documentation

PR #79, opened 2017-08-16
Add vm security group operations

tzumainn:


tzumainn:
Depends on https://github.com/ManageIQ/manageiq/pull/15826

Operations tested from rails console.

aufi:
Looks good to me!

Rubocop complaint msg - common in OSP code, we might refactor it later.
Codeclimate - make sense to me to ignore in this case.

PR #81, opened 2017-08-24
Add class for parsing refresh targets from EmsEvents

mansam:
This class parses refresh targets from EmsEvents. It's invoked from automate, see [here](https://github.com/ManageIQ/manageiq/blob/ef367b7e2e34750d3142494c42327c5c13da3dfb/app/models/ems_event/automate.rb#L5) and https://github.com/ManageIQ/manageiq-content/pull/159.


aufi:
Looks good to me!

PR #82, opened 2017-08-24
Fix collector caching and improve collection of network relations for targeted refresh

mansam:


mansam:
The codeclimate similar code issues are not actually issues.

aufi:
Looks good to me, merging.

PR #83, opened 2017-08-25
Cloud volume backup

alexander-demichev:
Add cloud volume restore and delete raw actions.

https://github.com/ManageIQ/manageiq/pull/15891


tzumainn:
:+1: just waiting for tests to pass

PR #85, opened 2017-08-28
Add notifications for VM destroy, Cloud Volume and Cloud Volume Snapshot actions

petrblaho:
This adds generic with_notification method to be used to wrap call for fog-openstack actions.

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/vm/operations.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/vm/operations.rb
@@ -8,7 +8,12 @@ module ManageIQ::Providers::Openstack::CloudManager::Vm::Operations
 
   def raw_destroy
     raise "VM has no #{ui_lookup(:table => "ext_management_systems")}, unable to destroy VM" unless ext_management_system
-    with_provider_object(&:destroy)
+    with_notification(:vm_destroy,
+                      :options => {
+                        :subject => self,
+                      }) do
+      with_provider_object(&:destroy)
+    end
     self.update_attributes!(:raw_power_state => "DELETED")
   end
 end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -8,6 +8,8 @@ VCR.configure do |config|
   config.cassette_library_dir = File.join(ManageIQ::Providers::Openstack::Engine.root, 'spec/vcr_cassettes')
 end
 
+NotificationType.seed
+
 Dir[Rails.root.join("spec/shared/**/*.rb")].each { |f| require f }
 Dir[ManageIQ::Providers::Openstack::Engine.root.join("spec/support/**/*.rb")].each { |f| require f }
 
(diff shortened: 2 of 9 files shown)


app/models/manageiq/providers/openstack/cloud_manager/vm/operations.rb
```diff
@@ -8,7 +8,14 @@ module ManageIQ::Providers::Openstack::CloudManager::Vm::Operations
 
   def raw_destroy
     raise "VM has no #{ui_lookup(:table => "ext_management_systems")}, unable to destroy VM" unless ext_management_system
-    with_provider_object(&:destroy)
+    with_notification(:vm_destroy,
+                      :options => {
```
aufi: Thanks for the PR, the approach looks good to me.

I think, ```:instance_name => ...``` should not be wrapped in ```:options => ...```

petrblaho: In fact it should - with_notification uses keyword parameter `:options` not the ruby "options hash".

tzumainn:
looks like the hash alignment issues are a false positive caused by a nested hash.

tzumainn:
I thought that re-running the tests might pass CI, but it looks like it doesn't. Do the notifications need the _success or _error suffix, or are those supposed to be automatically determined?

app/models/manageiq/providers/openstack/cloud_manager/vm/operations.rb
aufi: Ah, thanks, I forgot about this Ruby 2.0 update.

petrblaho:
@tzumainn w/r/t `_success` and `_error` suffixes - notification type definition do _not_ need these but it is now convention to have (usually) pair of notifications. We have now definition for our notification there and they are with these suffixes. `with_notification` currently wraps some other call (fog-openstack) and fires up either success or error notification depending on the outcome and it handles adding `_success` and `_error` suffixes.

w/r/t failing tests I forgot to add `NotificationType.seed` to appropriate spec files.

spec/spec_helper.rb
```diff
@@ -8,6 +8,8 @@
   config.cassette_library_dir = File.join(ManageIQ::Providers::Openstack::Engine.root, 'spec/vcr_cassettes')
 end
 
+NotificationType.seed
```
aufi: This should be moved to https://github.com/ManageIQ/manageiq/blob/master/lib/evm_database.rb in following PR, no merge blocker.

Loicavenel:
@petrblaho I did not see this PR before but we should not call Destroy in CF but Delete.

mansam:
@petrblaho @aufi do you know if it would be possible to backport this and https://github.com/ManageIQ/manageiq/pull/15900 to Fine in order to fix https://bugzilla.redhat.com/show_bug.cgi?id=1524356?

aufi:
It should be possible IMO (not sure about Seed in spec, but that's just a test).

mansam:
Marked fine/yes. Needs to be backported for https://bugzilla.redhat.com/show_bug.cgi?id=1524356

PR #86, opened 2017-08-30
Orchestration Stack and Cloud Tenant targeted refresh

mansam:
Depends on https://github.com/ManageIQ/manageiq-providers-openstack/pull/81 and https://github.com/ManageIQ/manageiq-providers-openstack/pull/82.


tzumainn:
+1 looks like it works!

PR #89, opened 2017-09-05
Add create action for Image

andyvesel:
Added `create` method for Image

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/template.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/template.rb
@@ -73,6 +73,29 @@ class ManageIQ::Providers::Openstack::CloudManager::Template < ManageIQ::Provide
     false
   end
 
+  def self.raw_create_image(ext_management_system, create_options)
+    ext_management_system.with_provider_connection(:service => 'Image') do |service|
+      service.create_image(create_options)
+    end
+  rescue => err
+    _log.error("image=[#{name}], error=[#{err}]")
+    raise MiqException::MiqOpenstackApiRequestError, err.to_s, err.backtrace
+  end
+
+  def self.validate_create_image(ext_management_system, _options = {})
+    if ext_management_system
+      {:available => true, :message => nil}
+    else
+      {:available => false,
+       :message   => _("The Image is not connected to an active %{table}") %
+         {:table => ui_lookup(:table => "ext_management_system")}}
+    end
+  end
+
+  def self.create_image(ext_management_system, create_options)
+    raw_create_image(ext_management_system, create_options)
+  end
+
   def raw_delete_image
     ext_management_system.with_provider_connection(:service => 'Compute') do |service|
       service.delete_image(ems_ref)
--- a/spec/models/manageiq/providers/openstack/cloud_manager/template_spec.rb
+++ b/spec/models/manageiq/providers/openstack/cloud_manager/template_spec.rb
@@ -1,8 +1,39 @@
 describe ManageIQ::Providers::Openstack::CloudManager::Template do
   let(:ems) { FactoryGirl.create(:ems_openstack_with_authentication) }
+  let(:image_attributes) { {:name => "image", :ram => "1"} }
   let(:template_openstack) { FactoryGirl.create :template_openstack, :ext_management_system => ems, :ems_ref => 'one_id' }
   let(:service) { double }
 
+  context 'when create_image' do
+    before do
+      allow(ems).to receive(:with_provider_connection).with(:service => 'Image').and_yield(service)
+    end
+
+    context 'with correct data' do
+      it 'should create image' do
+        expect(service).to receive(:create_image).with(image_attributes).once
+        subject.class.create_image(ems, image_attributes)
+      end
+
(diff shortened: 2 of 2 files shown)


aufi:
@andyvesel Thanks for PR. Similar as for delete image PR, I'd suggest put there also non-raw method even keep it in same model to keep the consistency. (parent class is generic for all providers so it's not sure whether we want it also there)

And it would be great if you coudl add some basic tests for these methods.

andyvesel:
@aufi tests added

app/models/manageiq/providers/openstack/cloud_manager/template.rb
```diff
@@ -73,6 +73,29 @@ def requires_storage_for_scan?
     false
   end
 
+  def self.raw_create_image(ext_management_system, create_options)
+    ext_management_system.with_provider_connection(:service => 'Compute') do |service|
+      service.images.create(create_options)
```
aufi: Nit: I think we usually use request methods (instead of fog model's one), search for ```image_create``` at https://github.com/fog/fog-openstack/tree/master/lib/fog/compute/openstack/requests

spec/models/manageiq/providers/openstack/cloud_manager/template_spec.rb
```diff
@@ -1,8 +1,43 @@
...
   let(:ems) { FactoryGirl.create(:ems_openstack_with_authentication) }
+  let(:image_attributes) { {:name => "image", :ram => "1"} }
   let(:template_openstack) { FactoryGirl.create :template_openstack, :ext_management_system => ems, :ems_ref => 'one_id' }
   let(:service) { double }
 
+  context 'when create_image' do
+    before do
+      allow(ExtManagementSystem).to receive(:find).with(ems.id).and_return(ems)
+      allow(ems).to receive(:with_provider_connection).with(:service => 'Compute').and_yield(service)
+      allow(service).to receive(:images).and_return(images)
```
aufi: Some of these lines are not needed.  E.g. ```ExtManagementSystem``` exists in database so it can be found directly from model (created by line 2 and first usage of ```ems``` variable). For example take a look at https://gist.github.com/aufi/16acd1f29086eeb211ceabd3bc98625e I don't want push you to rewrite your code much, just for information.

spec/models/manageiq/providers/openstack/cloud_manager/template_spec.rb
```diff
@@ -1,8 +1,43 @@
...
+    end
+
+    let(:images) { double }
+
+    context 'with correct data' do
+      it 'should create image' do
+        expect(images).to receive(:create).with(image_attributes).and_return(template_openstack).once
```
aufi: Using service (and its ```create_image``` method) directly would be easier, no need to a "double mock" then.

Nit: I'm not sure if we care about returned ```template_openstack``` data from Fog call, so we don't have to test it then.

andyvesel:
@aufi thanks for review and comments. Will fix it asap

andyvesel:
@aufi fixed

app/models/manageiq/providers/openstack/cloud_manager/template.rb
```diff
@@ -73,6 +73,29 @@ def requires_storage_for_scan?
     false
   end
 
+  def self.raw_create_image(ext_management_system, create_options)
+    ext_management_system.with_provider_connection(:service => 'Compute') do |service|
```
aufi: Please, try use Image service instead of Compute.

spec/models/manageiq/providers/openstack/cloud_manager/template_spec.rb
```diff
@@ -1,8 +1,43 @@
 describe ManageIQ::Providers::Openstack::CloudManager::Template do
   let(:ems) { FactoryGirl.create(:ems_openstack_with_authentication) }
+  let(:image_attributes) { {:name => "image", :ram => "1"} }
   let(:template_openstack) { FactoryGirl.create :template_openstack, :ext_management_system => ems, :ems_ref => 'one_id' }
   let(:service) { double }
 
+  context 'when create_image' do
+    before do
+      allow(ExtManagementSystem).to receive(:find).with(ems.id).and_return(ems)
```
aufi: I believe this is line is not needed, could you check it?

spec/models/manageiq/providers/openstack/cloud_manager/template_spec.rb
```diff
@@ -1,8 +1,43 @@
...
+      allow(ems).to receive(:with_provider_connection).with(:service => 'Compute').and_yield(service)
+      allow(service).to receive(:images)
```
aufi: I believe this is line is not needed, could you check it?

spec/models/manageiq/providers/openstack/cloud_manager/template_spec.rb
```diff
@@ -1,8 +1,43 @@
...
+    end
+
+    let(:images) { double }
```
aufi: I believe this is line is not needed, could you check it?

aufi:
Looks good to me, merging.

PR #93, opened 2017-09-07
Refresher sets proper OrchestrationTemplate type

bzwei:
This is to complete the work from #76. The refresher will need to set type to the new Openstack Orchestration Template type.

Previously depending on the format (json/yaml) of the template refresher sets the type to be either `OrchestrationTemplateCfn` or `OrchestrationTemplateHot`. Now they are unified to `ManageIQ::Providers::Openstack::CloudManager::OrchestrationTemplate`. Therefore the class itself is updated to support both json and yaml formats.

Note currently we don't have  `ManageIQ::Providers::Openstack::InfraManager::OrchestrationTemplate` class. So the template collected from an infra stack will be `CloudManager::OrchestrationTemplate`. Is this acceptable?


bzwei:
@miq-bot add_label refactoring
@miq-bot add_label enhancement
@miq-bot assign @tzumainn 
cc @Ladas

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -220,7 +220,7 @@ def orchestration_template(stack)
     template = collector.orchestration_template(stack)
     if template
       o = persister.orchestration_templates.find_or_build(stack.id)
-      o.type = stack.template.format == "HOT" ? "OrchestrationTemplateHot" : "OrchestrationTemplateCfn"
```
Ladas: @bzwei why did we have CFN template format here, when people are not using HOT?

bzwei: Yes, openstack can take CFN template too. With this refactored class, now Openstack's template support both JSON (CFN) and YAML (HOT) format.

Ladas:
@bzwei hm, it should be ok to use the safe_load, right? Since this is user input, we should porbably load it safely.

bzwei:
@Ladas A typical template contains string like this `version: 2013-05-13`
`safe_load` causes an error: `*** Psych::DisallowedClass Exception: Tried to load unspecified class: Date`
`load` has no error but parses the string as a date object, which is not of our interest but at least does not fail us.

Ladas:
@bzwei right, I think we will have to solve it a different way. The load should not be used for user input data, it allows remote code execution (check the example in) https://brakemanscanner.org/docs/warning_types/remote_code_execution_yaml_load

app/models/manageiq/providers/openstack/cloud_manager/orchestration_template.rb
```diff
@@ -65,14 +65,37 @@ def self.eligible_manager_types
...
+  # returns either json or yaml
+  def format
+    content.strip.start_with?('{') ? 'json'.freeze : 'yaml'.freeze
+  end
+
+  private
+
+  def parse
+    return JSON.parse(content) if format == 'json'
+    YAML.load(content)
```
Ladas: @bzwei you should be able to use  YAML.safe_load(content, [Date])

bzwei:
@Ladas fixed

Ladas:
@tzumainn looks good to me now, can you guys review?

Ladas:
ping @tzumainn @aufi

aufi:
Looks good to me, merging.

PR #94, opened 2017-09-08
Assign Infra host hostname as hypervisor_hostname

robbmanes:
Instead of having the hostname for infrastructure managers show up as an
IP address, have it appear as the hostname as it does for other
providers like RHEV/OVirt.


robbmanes:
I saw that in OVirt/RHEV it passes the hostname to the "hostname" field and think this is supposed to be similar - if I am wrong or in left field please feel free to close.  Thanks!

robbmanes:
I stupidly opened this on master, closing and re-opening in favor of https://github.com/ManageIQ/manageiq-providers-openstack/pull/95

PR #95, opened 2017-09-08
Assign Infra host hostname as hypervisor_hostname

robbmanes:
Instead of having the hostname for infrastructure managers show up as an
IP address, have it appear as the hostname as it does for other
providers like RHEV/OVirt.

### DIFF
--- a/app/models/manageiq/providers/openstack/infra_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/openstack/infra_manager/refresh_parser.rb
@@ -172,7 +172,7 @@ module ManageIQ
         host_name           = identify_host_name(indexed_resources, host.instance_uuid, uid)
         hypervisor_hostname = identify_hypervisor_hostname(host, indexed_servers)
         ip_address          = identify_primary_ip_address(host, indexed_servers)
-        hostname            = ip_address
+        hostname            = hypervisor_hostname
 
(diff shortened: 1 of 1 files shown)


robbmanes:
I saw that in OVirt/RHEV it passes the hostname to the "hostname" field and think this is supposed to be similar - if I am wrong or in left field please feel free to close. Thanks!

tzumainn:
Hi! Thanks for making this PR. I think it partially works; the problem is that it won't work if there's no instance deployed on the node, as that's where the hypervisor_hostname comes from. That's why we default to the UUID - so the node 'name' is always consistent.

robbmanes:
Ah, understood.  Could we assign it via another method so the host name of the node would appear instead of the IP, for reporting reasons, say, with 'host_name' or some other method?

tzumainn:
To be honest I'm not 100% how the reporting works - is information pulled from the database, or via some other method? The ideal solution may depend on the answer to that.

robbmanes:
So in included MIQ reporting, the UUID is in fact used and this seems to be fine.  That being said, the hostname of the hosts using the OpenStack provider gem still is just the IP address under /host/show/$id, but I'm not sure what's the proper/more elegant solution for this right now, so I'll close this PR for that right now.  Thanks!

PR #96, opened 2017-09-08
Refactor Refresh Settings so that they are consistent with other providers

mansam:
This PR puts the refresh settings for the Openstack provider in the correct spot and splits the settings for the Network Manager into their own section in order to better match with ManageIQ convention. Necessary to support https://github.com/ManageIQ/manageiq/pull/15933/files


mansam:
WIP while I work on supporting the old settings.

ITD27M01:
@mansam 

Typo in get_images from https://github.com/ManageIQ/manageiq-providers-openstack/pull/96/files#diff-a2cc8fa91ab5a25366adbc5bfbf3901a

should be "openstack_admin?" instead of "openstack_network_admin?".

mansam:
@ITD27M01 Good catch, much appreciated.

mansam:
@tzumainn @aufi could you review when you have a chance?

mansam:
This is probably going to get refactored to remove the old "wrong" settings and add a data migration to handle the switchover.

tzumainn:
Code looks good, tested on existing provider and new provider :+1:

PR #97, opened 2017-09-13
Check provisioning status with the correct tenant scoping

mansam:
If a volume or VM is provisioned in a tenant other than the default, then using `<collection>.get()` may fail to find them during the provisioning status check because the request is scoped to the wrong tenant. This PR scopes the request to use the tenant name from the provisioning request if it is a available.

This also fixes a performance problem with the cloning check, where in order to work around the scoping problem it was listing every server from every tenant. If the tenant name is available, then it will use `servers.get()` instead.

(Also some minor co-located lint fixes)

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/provision/cloning.rb
@@ -1,7 +1,12 @@
 module ManageIQ::Providers::Openstack::CloudManager::Provision::Cloning
   def do_clone_task_check(clone_task_ref)
-    source.with_provider_connection do |openstack|
-      instance = openstack.handled_list(:servers).detect { |s| s.id == clone_task_ref }
+    connection_options = {:tenant_name => options[:cloud_tenant][1]} if options[:cloud_tenant].kind_of?(Array)
+    source.with_provider_connection(connection_options) do |openstack|
+      instance = if connection_options
+                   openstack.servers.get(clone_task_ref)
+                 else
+                   openstack.handled_list(:servers).detect { |s| s.id == clone_task_ref }
+                 end
       status   = instance.state.downcase.to_sym
 
       if status == :error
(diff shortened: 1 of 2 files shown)


mansam:
@petrblaho @aufi it would be great if you could check whether this makes sense to you. :)

mansam:
This is a possible fix for https://github.com/ManageIQ/manageiq-providers-openstack/issues/66

petrblaho:
LGTM +1

app/models/manageiq/providers/openstack/cloud_manager/provision/cloning.rb
```diff
@@ -1,8 +1,13 @@
...
-      instance = openstack.handled_list(:servers).detect { |s| s.id == clone_task_ref }
-      status   = instance.state.downcase.to_sym
+    connection_options = {:tenant_name => options[:cloud_tenant][1]} if options[:cloud_tenant].kind_of?(Array)
+    source.with_provider_connection(connection_options) do |openstack|
+      instance = if connection_options
+                   openstack.servers.get(clone_task_ref)
+                 else
+                   openstack.handled_list(:servers).detect { |s| s.id == clone_task_ref }
+                 end
+      status   = instance.state.downcase.to_sym if instance.present?
```
aufi: Thanks for the PR. I think we should initialize ```status``` local variable always to avoid _undefined variable error_ since it is part of condition on line 12.

mansam: `status   = instance.state.downcase.to_sym if instance.present?` results in `status` being defined as `nil` if the condition is false, so an additional initialization shouldn't be necessary.
```
irb(main):001:0> x = 3 if true
=> 3
irb(main):002:0> x
=> 3
irb(main):003:0> y = 3 if false
=> nil
irb(main):004:0> y
=> nil
irb(main):005:0> z
NameError: undefined local variable or method `z' for main:Object
```

aufi: Ah, thanks!

PR #100, opened 2017-09-20
Change vms state map deleted to 'terminated' instead of 'archived'

zhitongLIU:
Use `terminated` as other providers, so that vm that is deleted won't continue to show on provider current active vms page, but in `archived` vms page

manageiq should be a tool of unification but not creating new divergences

Examples of `terminated` usage in other providers 
* https://github.com/ManageIQ/manageiq-providers-amazon/blob/7835e02d792216e7c380a49f7af609945c150618/app/models/manageiq/providers/amazon/cloud_manager/vm.rb#L12
* https://github.com/ManageIQ/manageiq-providers-vmware/blob/ba1583e269620784bdbe5e367e80fed61fc95f5d/app/models/manageiq/providers/vmware/cloud_manager/vm.rb#L13

Usage of `terminated`  in manageiq
* https://github.com/ManageIQ/manageiq/blob/c1549f0067bbdbcd6a3f9cfd89b8d0631744e57e/app/models/vm_or_template.rb#L309

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/vm.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/vm.rb
@@ -26,7 +26,7 @@ class ManageIQ::Providers::Openstack::CloudManager::Vm < ManageIQ::Providers::Cl
     "ERROR"             => "non_operational",
     "BUILD"             => "wait_for_launch",
     "REBUILD"           => "wait_for_launch",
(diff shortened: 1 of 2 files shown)


zhitongLBN:
PR number 100 !!!!! :1st_place_medal:

aufi:
Thanks for the PR. Looks good to me, merging. :+1:

PR #102, opened 2017-09-25
Add update action for Image

andyvesel:
Added `update` method for Image

### DIFF
--- a/spec/models/manageiq/providers/openstack/cloud_manager/template_spec.rb
+++ b/spec/models/manageiq/providers/openstack/cloud_manager/template_spec.rb
@@ -1,6 +1,6 @@
 describe ManageIQ::Providers::Openstack::CloudManager::Template do
-  let(:ems) { FactoryGirl.create(:ems_openstack_with_authentication) }
-  let(:image_attributes) { {:name => "image", :ram => "1"} }
+  let(:ems) { FactoryGirl.create(:ems_openstack) }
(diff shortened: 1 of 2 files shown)


spec/models/manageiq/providers/openstack/cloud_manager/template_spec.rb
```diff
@@ -34,6 +35,22 @@
       end
     end
   end
+
+  context 'when update_image' do
+    before do
+      allow(ExtManagementSystem).to receive(:find).with(ems.id).and_return(ems)
+      allow(ems).to receive(:with_provider_connection).with(:service => 'Image').and_yield(service)
```
andyvesel: @aufi everytime the `ems` here appears to be `nil` and tests don't pass. How can I get a right one? Maybe something goes wrong in line 2
```let(:ems) { FactoryGirl.create(:ems_openstack_with_authentication) }```
and FactoryGirl doesn't create right ems (some issues with authentication). Could you check it pls?

aufi:
@andyvesel Hi, that looks much better! :+1:  Few nits:

It looks to me that ```let(:images)``` is not needed.
Not merge-blocker, but I'd recomment consider rename ```found_image``` to ```fog_image``` to better describe the object better.

aufi:
Looks good to me, merging!

Ignoring Rubocop complains about ```find_by_id``` method since the method comes from Fog model instead of ActiveRecord where it was deprecated.

PR #104, opened 2017-09-28
Enable provisioning from Volumes and Volume Snapshots via a proxy type

mansam:
This pull request implements proxies for Volumes and Volume Snapshots as subclasses of MiqTemplate, allowing them to be selected during the ManageIQ provisioning workflow. Openstack will automatically create a volume of the appropriate size if a Snapshot is chosen as the boot source.

See also:
 https://github.com/ManageIQ/manageiq/pull/16066
https://github.com/ManageIQ/manageiq-ui-classic/pull/2253


gmcculloug:
@agrare Can you review or assign someone.  Thanks

app/models/manageiq/providers/openstack/cloud_manager/volume_snapshot_template.rb
```diff
@@ -0,0 +1,12 @@
+class ManageIQ::Providers::Openstack::CloudManager::VolumeSnapshotTemplate < ManageIQ::Providers::CloudManager::Template
+  # VolumeSnapshotTemplates are proxies to allow provisioning instances from volumes
+  # without having to refactor the entire provisioning workflow to support types
+  # other than VmOrTemplate subtypes. VolumeSnapshotTemplates are created 1-to-1 during
+  # inventory refresh for each eligible snapshot.
+
+  belongs_to :cloud_tenant
+
+  def volume_snapshot_template?
```
agrare: Where is this used?  Better to just use `.kind_of?(klass)` I think

app/models/manageiq/providers/openstack/cloud_manager/volume_template.rb
```diff
@@ -0,0 +1,12 @@
+class ManageIQ::Providers::Openstack::CloudManager::VolumeTemplate < ManageIQ::Providers::CloudManager::Template
+  # VolumeTemplates are proxies to allow provisioning instances from volumes
+  # without having to refactor the entire provisioning workflow to support types
+  # other than VmOrTemplate subtypes. VolumeTemplates are created 1-to-1 during
+  # inventory refresh for each eligible bootable volume.
+
+  belongs_to :cloud_tenant
+
+  def volume_template?
```
agrare: Same

agrare:
So I'm not a huge fan of creating a dummy miq_template record in refresh so that we can provision from it instead of changing the provision workflow to be able to use the volume as the source, @Fryguy thoughts?

spec/models/manageiq/providers/openstack/cloud_manager/refresh_spec_common.rb
```diff
@@ -296,9 +309,17 @@ def assert_ems
       expect(@ems.key_pairs.size).to          eq compute_data.key_pairs.count
       security_groups_count = @ems.security_groups.count { |x| x.name != 'default' }
       expect(security_groups_count).to        eq security_groups_count
-      expect(@ems.vms_and_templates.size).to  eq vms_count + images_count
+      if ::Settings.ems.ems_openstack.try(:refresh).try(:inventory_object_refresh) || ::Settings.ems.ems_refresh.try(:openstack).try(:inventory_object_refresh)
```
Fryguy: Alternately, fetch_path is usable on Settings as well, so this could be

```ruby
Settings.ems.fetch_path(:ems_openstack, :refresh, : inventory_object_refresh) || Settings.ems.fetch_path(:ems_refresh, :openstack, : inventory_object_refresh)
```

On that note, why are there even 2 ways to configure things?  Pick whichever one you want to be the "official" one (I suggest the former, so you can own the setting in your repo), and create a data migration to convert the other one.

app/models/manageiq/providers/openstack/inventory/collector/cloud_manager.rb
```diff
@@ -144,4 +144,22 @@ def orchestration_resources(stack)
...
   end
+
+  def volume_templates
+    return [] unless volume_service
+    return @volume_templates if @volume_templates.any?
+    @volume_templates = volume_service.handled_list(:volumes, {:status => "available"}, ::Settings.ems.ems_openstack.refresh.is_admin)
+  end
+
+  def volumes_by_id
+    @volumes_by_id ||= Hash[volume_templates.collect { |f| [f.id, f] }]
```
Fryguy: Alternately, you can use the index_by method.

```ruby
@volumes_by_id ||= volume_templates.index_by(&:id)
```

Fryguy:
@agrare I don't see it creating dummy data, unless I'm mistaken about how the collector works?

agrare:
@Fryguy there already are `CloudVolume` and `CloudVolumeSnapshot` models, this also creates volumes as a `MiqTemplate` type instead of a `CloudVolume` type.  So a duplicate record of another type is probably more correct than a dummy record.

app/models/manageiq/providers/openstack/inventory/collector/cloud_manager.rb
mansam: Thanks for that shortcut!

spec/models/manageiq/providers/openstack/cloud_manager/refresh_spec_common.rb
mansam: Good to know fetch_path is available on Settings. There are two ways to configure things because the latter (incorrect) option was put in first and merged into a release. I put together a PR to support the correct way and the old way to avoid breaking anyone's deployment since I wasn't aware I could use a migration to change things over. Since the PR that implements the new openstack-owned approach wasn't merged yet, I should be able to implement a migration and get this all cleaned up. :)

mansam:
@agrare @Fryguy other than the minor changes regarding `volume_snapshot_template?` and `volume_template?`, did you have any other thoughts? This feature is a requirement for 4.6 so I'd like to know what else needs to be done to get it in.

app/models/manageiq/providers/openstack/cloud_manager/volume_snapshot_template.rb
mansam: These are in line with `image?` and `snapshot?` which are present on the parent class and used for differentiating type in the UI.

agrare: :+1: thanks

agrare:
Given the requirements for provisioning I'm good with this, ideally we would be able to provision from things other than a template but I understand that would be a large undertaking.

@mansam if this is ready can you take it out of `[WIP]` and rebase to fix up the conflicts?

mansam:
@agrare rebase in progress :)

PR #108, opened 2017-10-04
Event parser sets host id

alexander-demichev:
Event parser now sets host_id attribute. It tries to find host by `ems_ref_obj` using `instance_id` attribute from event's payload.
[BZ link](https://bugzilla.redhat.com/show_bug.cgi?id=1377786)


aufi:
Thanks for PR! Looks good to me.

Ignoring rubocop complains - the code is common in given file.

PR #109, opened 2017-10-06
Use floating_ip_address instead of name for creation messages

mansam:
Needs to be merged along with https://github.com/ManageIQ/manageiq-ui-classic/pull/2312 to fix https://bugzilla.redhat.com/show_bug.cgi?id=1490918


aufi:
Loks good to me, merging :+1:

PR #110, opened 2017-10-06
Trim neutron error messages out of fog responses

mansam:
This PR trims Neutron error messages out of fog responses, so that the UI can display readable messages instead of the mess of json and header info that currently appears. This is related to the problem described in https://bugzilla.redhat.com/show_bug.cgi?id=1476666, which will eventually be addressed as I go through the rest of the provider.


tzumainn:
:+1: Looks good!

PR #111, opened 2017-10-06
Trim error messages out of cloud tenant fog responses

mansam:
Makes the errors returned from the API when a cloud tenant action fails human readable. Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1476666, though problems of this category exist all over the place.


aufi:

PR #113, opened 2017-10-11
Filter out resources with blank physical_resource_id

Ladas:
Filter out resources with blank physical_resource_id.
Before we were ignoring those with:
https://github.com/Ladas/manageiq-providers-openstack/blob/d0799c4a5766650f2d3d149875c1b58f8a16040f/app/models/manageiq/providers/openstack/refresh_parser_common/orchestration_stacks.rb#L165

Without ignoring thise results in error :
[NoMethodError]: undefined method `ems_ref=' for nil:NilClass  Method:[block in method_missing]

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
@@ -170,7 +170,7 @@ class ManageIQ::Providers::Openstack::Inventory::Parser::CloudManager < ManageIQ
     raw_resources = collector.orchestration_resources(stack)
     # reject resources that don't have a physical resource id, because that
     # means they failed to be successfully created
-    raw_resources.reject! { |r| r.physical_resource_id.nil? }
+    raw_resources.reject! { |r| r.physical_resource_id.blank? }
     raw_resources.each do |resource|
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -173,6 +173,8 @@ def orchestration_stack_resources(stack, stack_inventory_object)
     raw_resources.reject! { |r| r.physical_resource_id.nil? }
```
mansam: There's already a check on this line for nil, just like in the old refresher. I'm guessing it should be changed to a check for blank?

Ladas: hum, you might be right

Ladas: yeah, the build method checks blank? https://github.com/Ladas/manageiq/blob/17c884ab58ea6e7f0aad11dd87982adf24409c17/app/models/manager_refresh/inventory_collection.rb#L754

mansam:
This should fix https://bugzilla.redhat.com/show_bug.cgi?id=1500314

PR #114, opened 2017-10-11
Filter out resources with no physical_resource_id

Ladas:
Filter out resources with no physical_resource_id.
Before we were ignoring those with:
https://github.com/Ladas/manageiq-providers-openstack/blob/d0799c4a5766650f2d3d149875c1b58f8a16040f/app/models/manageiq/providers/openstack/refresh_parser_common/orchestration_stacks.rb#L165

Without ignoring thise results in error :
[NoMethodError]: undefined method `ems_ref=' for nil:NilClass  Method:[block in method_missing]


Ladas:
duplicate

PR #115, opened 2017-10-12
EMS infra: Openstack discover

gildub:
Adds initial discovery for Openstack undercloud (ems infra) probing baremetal (Ironic) API 
https://bugzilla.redhat.com/show_bug.cgi?id=1212947
Depends upon https://github.com/ManageIQ/manageiq-network_discovery/pull/5

### DIFF
--- a/app/models/manageiq/providers/openstack/infra_discovery.rb
+++ b/app/models/manageiq/providers/openstack/infra_discovery.rb
@@ -1,0 +1,17 @@
+require 'manageiq/network_discovery/port_scanner'
+
+module ManageIQ::Providers::Openstack
+  class InfraDiscovery
+    def self.probe(ost)
+      res = ""
+      if PortScanner.portOpen(ost, 6385)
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/openstack/infra_discovery.rb
```diff
@@ -0,0 +1,17 @@
+require 'manageiq/network_discovery/port_scanner'
+
+module ManageIQ::Providers::Openstack
+  class InfraDiscovery
+    def self.probe(ost)
+      res = ""
+      if PortScanner.portOpen(ost, 6385)
```
Fryguy: If possible, can we change the PortScanner to be fully namespaced under ManageIQ::NetworkDiscovery, since you are already making changes in there?

gildub: Absolutely! Thanks

aufi:
Looks good to me, the dependent PR is merged, merging!

PR #117, opened 2017-10-13
Move queuing logic for scaling from ui-classic

rwsu:
Moved logic into this repo for deciding whether to use Mistral
workflows if they are available or whether to use direct stack
update if the undercloud environment is OSP 9 or lower.

### DIFF
--- a/app/models/manageiq/providers/openstack/infra_manager/orchestration_stack.rb
+++ b/app/models/manageiq/providers/openstack/infra_manager/orchestration_stack.rb
@@ -60,14 +60,6 @@ class ManageIQ::Providers::Openstack::InfraManager::OrchestrationStack < ::Orche
     raise MiqException::MiqOrchestrationStatusError, err.to_s, err.backtrace
   end
 
-  def queue_post_scaledown_task(services, task_id = nil)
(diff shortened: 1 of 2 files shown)


tzumainn:
Is there a corresponding UI PR to make sure that the method changes are in sync, or is this completely standalone?

tzumainn:
Oh, nevermind, I see it now

app/models/manageiq/providers/openstack/infra_manager/orchestration_stack.rb
```diff
@@ -138,5 +160,17 @@ def log_and_raise_update_error(method_name, err)
     raise MiqException::MiqOrchestrationUpdateError, err.to_s, err.backtrace
   end
 
+  def workflow_service
+    ext_management_system.workflow_service
```
mansam: Should this be `ext_management_system.openstack_handle.workflow_service` or (or `detect_workflow_service` to handle the exception)?

mansam: Nevermind, I see that `ext_management_system.workflow_service` is actually calling `ext_management_system.openstack_handle.detect_workflow_service` on the InfraManager.

PR #118, opened 2017-10-13
update raw connect method to accomodate OpenStack complexity

tzumainn:
The raw_connect method is used in provider validation; this change allows us to validate complex cases such as keystone v3 and ssl

Depends on https://github.com/ManageIQ/manageiq-ui-classic/pull/2394

https://bugzilla.redhat.com/show_bug.cgi?id=1500379


petrblaho:
Tested on TripleO setup, both undercloud and overcloud credentials validations work.

LGTM +1

PR #119, opened 2017-10-16
Only update tenant mapping for the network manager if it's present

mansam:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1471146 by checking for the presence of the network manager before attempting to modify it.


tzumainn:
:+1: makes sense!

PR #120, opened 2017-10-16
Trim key pair errors out of api responses

mansam:
Depends on https://github.com/ManageIQ/manageiq-providers-openstack/pull/110
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1466397


tzumainn:
:+1: works well, thanks!

PR #123, opened 2017-10-20
Trim Volume error messages out of Fog responses

mansam:


aufi:
Looks good to me, merging!

PR #124, opened 2017-10-21
User Sync

rwsu:
Interface for syncing users and tenant/group memberships from
OpenStack into ManageIQ.

### DIFF
--- a/app/models/manageiq/providers/openstack/identity_sync_mixin.rb
+++ b/app/models/manageiq/providers/openstack/identity_sync_mixin.rb
@@ -12,7 +12,10 @@ module ManageIQ::Providers::Openstack::IdentitySyncMixin
     openstack_users = list_users
     openstack_users.each do |u|
       username = u["name"]
+      user_uuid = u["id"]
       next if skip_user?(username)
+      user_projects = keystone.list_user_projects_tenants(user_uuid)
+      next unless user_projects.count.positive?
       user = User.find_by(:userid => username)
       users << u if user.nil?
     end
@@ -91,6 +94,10 @@ module ManageIQ::Providers::Openstack::IdentitySyncMixin
       # user already exist with this user name
       # if email doesn't match, then this record should be skipped
       user = nil if user.email != email
+    elsif keystone.list_user_projects_tenants(openstack_uuid).count.zero?
+      # don't create a new user if the user is not a member of
+      # any tenants in OpenStack because the user's current_group
+      # attribute will be nil and will not be able to login.
     else
       user = User.new
       user.name = username
--- a/app/models/manageiq/providers/openstack/identity_sync_mixin.rb
+++ b/app/models/manageiq/providers/openstack/identity_sync_mixin.rb
@@ -78,7 +78,10 @@ module ManageIQ::Providers::Openstack::IdentitySyncMixin
   end
 
   def skip_user?(username)
-    users_to_skip = ["admin", "neutron", "heat", "aodh", "cinder", "swift", "glance", "placement", "gnocchi", "nova", "heat-cfn", "panko", "ceilometer"]
+    users_to_skip = ['admin', 'neutron', 'heat', 'aodh', 'cinder', 'swift',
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/openstack/identity_sync_mixin.rb
```diff
@@ -0,0 +1,180 @@
+module ManageIQ::Providers::Openstack::IdentitySyncMixin
+  def list_users
+    # V2 requires adminURL as endpoint_type
+    connection_options = {:service => "Identity", :openstack_endpoint_type => 'adminURL'}
+    ext_management_system.with_provider_connection(connection_options) do |service|
+      service.list_users.body["users"]
+    end
+  end
```
tzumainn: Would keystone.users.all also work? The identity delegate code fore creating tenants uses that for listing all users, although it does so differently for keystone v2 and v3. Maybe it makes sense to create a list_users, list_users_v2, and list_users_v3 in the delegate (and also put the create methods in there as well)?

tzumainn: Nevermind about the create users, I just realized those are ManageIQ users being created, so that can't go in identity_delegate. I think list_users may belong there, though.

app/models/manageiq/providers/openstack/identity_sync_mixin.rb
```diff
@@ -0,0 +1,183 @@
...
+      end
+    end
+  end
+
+  def keystone
+    openstack_handle.identity_service
+  end
+
+  def skip_user?(username)
+    users_to_skip = ["admin", "neutron", "heat", "aodh", "cinder", "swift", "glance", "placement", "gnocchi", "nova", "heat-cfn", "panko", "ceilometer"]
```
tzumainn: I tested against an undercloud-used-as-a-cloud-provider, and noticed the following users were also created: mistral, zaqar-websocket, ironic, ironic-inspector, zaqar. I know that there are some plans in bringing these services into the overcloud - would it be possible to add them to this list?

tzumainn:
Thanks! I left two comments on the code, let me know if you can address them.

The password feature didn't work for me; I was unable to log into the ManageIQ web UI with the password I set.

app/models/manageiq/providers/openstack/identity_sync_mixin.rb
rwsu: I agree that list_users looks out of place in identity_sync_mixin. But it ddidn't work when I tried to put an equivalent method in identity_delegate like:

    def list_all_users
      # list_users fails in V2, because the URI is invalid
      # V2 requires adminURL as endpoint_type
      connection_options = {:service => "Identity", :openstack_endpoint_type => 'adminURL'}
      connection = @os_handle.connect(connection_options)
      connection.list_users.body["users"]
    end

will result in 

Expected([200, 204]) <=> Actual(404 Not Found) excon.error.response :body => "{\"error\": {\"message\": \"The resource could not be found.\", \"code\": 404, \"title\": \"Not Found\"}}" :cookies => [ ] :headers => { "Content-Length" => "93" "Content-Type" => "application/json" "Date" => "Wed, 25 Oct 2017 20:25:04 GMT" "Server" => "Apache" "Vary" => "X-Auth-Token" "x-openstack-request-id" => "req-695652d6-2cf0-496a-81a8-e502f148624f" } :host => "192.168.24.11" :local_address => "192.168.122.1" :local_port => 41668 :path => "/v2.0//users" :port => 5000 :reason_phrase => "Not Found" :remote_ip => "192.168.24.11" :status => 404 :status_line => "HTTP/1.1 404 Not Found\r\n" [ems_cloud/sync_users]

rwsu:
@tzumainn  I've added additional changes to not create a new user if the user is not a member of any tenants in OpenStack. I also added a change to the UI PR to disable the User Sync button if tenant mapping is not enabled. These changes address the password bugs you found. Thanks for testing.

mansam:
LGTM! :rocket:

PR #128, opened 2017-10-24
Adds vm_snapshot_success Notification creation

petrblaho:
Creates notification with `vm_snapshot_success` type with Vm as `subject` for `create` and `remove` as `snapshot_op`.

Notification Type is defined in https://github.com/ManageIQ/manageiq/pull/16286 which is needed for this one.

Partially solves https://bugzilla.redhat.com/show_bug.cgi?id=1429313


aufi:
@petrblaho Thanks for PR, I'm getting following error
```ActiveModel::UnknownAttributeError: unknown attribute 'snapshot_op' for Notification.```

petrblaho:
@aufi good catch, thank you. My bad for sure... Updating in a minute...

aufi:
Works, looks good to me, merging.

PR #130, opened 2017-10-25
Trim error messages from fog responses for remaining models

mansam:
Should make the UI prettier when things go wrong, instead of spraying poorly formatted JSON everywhere.


aufi:
Looks good to me, merging!

PR #132, opened 2017-10-25
Translate exceptions from raw_connect

mansam:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1501329, which occurred because validation when creating a new provider is now done by calling `raw_connect` with the task queue, bypassing the exception translation that occurs with `verify_credentials` is called directly.


aufi:
Looks good to me, merging!

PR #141, opened 2017-11-01
manageiq-gems-pending is already from manageiq itself

Fryguy:
While this doesn't have a problem on master because the lines in both Gemfiles (this one and manageiq/manageiq's Gemfile) are identical, this causes an issue on release branches because manageiq's gemfile says the new branch name (e.g. gaprindashvili), but then this file has a different branch name (e.g. master), and that blows up.  While we *could* just change it on the branch to the new release branch name, we don't want to have to remember to maintain that.

@simaishi Please review.
cc @aufi

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,8 +8,5 @@ gemspec
 # Git. Remember to move these dependencies to your gemspec before releasing
 # your gem to rubygems.org.
 
-# We are using 'miq-password' from gems-pending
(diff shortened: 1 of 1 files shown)


simaishi:

```
$ git log -1
commit 4a3f8e247468fca4884e9b1e9c3f31fe94e983b1
Date:   Wed Nov 1 16:12:26 2017 -0400

    Merge pull request #141 from Fryguy/fix_gemfile
    
    manageiq-gems-pending is already from manageiq itself
    (cherry picked from commit 587f3a225b460b36959b1a5e17bb2c60db60dd45)
```

PR #145, opened 2017-11-07
Remove floating_ip_address from the create request if it is blank

mansam:
Fixes a problem described in https://bugzilla.redhat.com/show_bug.cgi?id=1490918 where leaving the optional floating ip field on the creation form blank resulted in an API error.

### DIFF
--- a/app/models/manageiq/providers/openstack/network_manager/floating_ip.rb
+++ b/app/models/manageiq/providers/openstack/network_manager/floating_ip.rb
@@ -27,7 +27,6 @@ class ManageIQ::Providers::Openstack::NetworkManager::FloatingIp < ::FloatingIp
     floating_network_id = CloudNetwork.find(options[:cloud_network_id]).ems_ref
 
     raw_options = remapping(options)
-
     ext_management_system.with_provider_connection(connection_options(cloud_tenant)) do |service|
       floating_ip = service.create_floating_ip(floating_network_id, raw_options)
     end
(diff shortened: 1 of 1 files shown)


mansam:
@aufi when you get a chance, could you take a look at this? :)

app/models/manageiq/providers/openstack/network_manager/floating_ip.rb
```diff
@@ -42,6 +41,12 @@ def self.remapping(options)
     new_options[:floating_ip_address] = options[:address] if options[:address]
```
aufi: Hi Sam, sure, sorry for later reponse. Looks good, but from my point of view it would be even better adding ```.present?``` to the end of this line instead of allow setting option to nil or empty and deleting it later.

```new_options[:floating_ip_address] = options[:address] if options[:address].present?```

Comment (mansam) [reply app/models/manageiq/providers/openstack/network_manager/floating_ip.rb]: The UI form actually sets `new_options[:floating_ip_address]`, so that does need to be tested and blank values deleted either way. I'm not sure what the purpose is of the `options[:address]` line, but I've left it there since I assume it's there for a reason.

**Comment by aufi** [comment]
Ok, this is propably more safe, thanks for explanation.

**Comment by simaishi** [comment]

```
$ git log -1
commit 476a0dd7d9c43dac3c7fe8a55ff9f21618c604b9
Date:   Wed Nov 15 10:14:55 2017 +0100

    Merge pull request #145 from mansam/remove-blank-address-from-floating-ip-option-mapping
    
    Remove floating_ip_address from the create request if it is blank
    (cherry picked from commit 62c754be89b396bb9713d00241b8c489023caeb1)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1513625
```

PR #147, opened 2017-11-07
Handle attempts to delete volumes that have already been deleted

mansam:
This handles an exception that would occur when attempting to delete
a volume that was already deleted from the provider, but the deletion
was not yet reflected in ManageIQ. The exception is now caught and
a nice error message is shown.

https://bugzilla.redhat.com/show_bug.cgi?id=1487234

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume.rb
@@ -56,15 +56,8 @@ class ManageIQ::Providers::Openstack::CloudManager::CloudVolume < ::CloudVolume
   def validate_delete_volume
     msg = validate_volume
     return {:available => msg[:available], :message => msg[:message]} unless msg[:available]
-    begin
-      if with_provider_object(&:status) == "in-use"
-        return validation_failed("Delete Volume", "Can't delete volume that is in use.")
-      end
-    # NoMethodError here usually means the above block got a nil while checking the volume
-    # status, probably because the user already deleted the volume. Assume that's the case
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/openstack/cloud_manager/cloud_volume.rb
```diff
@@ -56,8 +56,15 @@ def raw_update_volume(options)
...
-    if with_provider_object(&:status) == "in-use"
-      return validation_failed("Create Volume", "Can't delete volume that is in use.")
+    begin
+      if with_provider_object(&:status) == "in-use"
+        return validation_failed("Delete Volume", "Can't delete volume that is in use.")
+      end
+    # NoMethodError here usually means the above block got a nil while checking the volume
+    # status, probably because the user already deleted the volume. Assume that's the case
+    # and show a reassuring message.
+    rescue NoMethodError
```
aufi: Hi @mansam , thanks for the PR!

I think there can be other cases when NoMethodError is raised than ```Volume already queued for deletion```.

We can merge this in case of emergency, but I'd like discuss changing the line ```with_provider_object(&:status) == "in-use"``` to stop connecting to OpenStack and reading the ```status``` information from MIQ database record instead. Do you know if this would be possible? Thanks!

mansam: Probably possible, I'll try it and see if I run into any problems.

aufi:
Looks good to me, thanks!

aufi:
It looks that backport is not needed from the BZ, but feel free to update it!

PR #154, opened 2017-11-13
Consolidate openstack refresh workers

mansam:
See https://github.com/ManageIQ/manageiq/pull/16465

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/refresher.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/refresher.rb
@@ -71,6 +71,18 @@ module ManageIQ::Providers
         @targets_by_ems_id[ems_id] = all_targets
       end
 
+      # sort the EMSes to be refreshed with cloud managers before other EMSes.
+      # since @targets_by_ems_id is a hash, we have to insert the items into a new
+      # hash in the order we want them to appear.
+      sorted_ems_targets = {}
+      # pull out the IDs of cloud managers and reinsert them in a new hash first, to take advantage of preserved insertion order
+      cloud_manager_ids = @targets_by_ems_id.keys.select { |key| @ems_by_ems_id[key].kind_of? ManageIQ::Providers::Openstack::CloudManager }
+      cloud_manager_ids.each { |ems_id| sorted_ems_targets[ems_id] = @targets_by_ems_id.delete(ems_id) }
+      # now that the cloud managers have been removed from @targets_by_ems_id, move the rest of the values
+      # over to the new hash and then replace @targets_by_ems_id.
+      @targets_by_ems_id.keys.each { |ems_id| sorted_ems_targets[ems_id] = @targets_by_ems_id.delete(ems_id) }
+      @targets_by_ems_id = sorted_ems_targets
+
       super
     end
 
--- a/app/models/manageiq/providers/openstack/cloud_manager/refresh_worker.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/refresh_worker.rb
@@ -1,6 +1,22 @@
 class ManageIQ::Providers::Openstack::CloudManager::RefreshWorker < ::MiqEmsRefreshWorker
   require_nested :Runner
 
+  # overriding queue_name_for_ems so PerEmsWorkerMixin picks up *all* of the
+  # Openstack-manager types from here.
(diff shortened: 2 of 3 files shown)


aufi:
Hi Sam, thanks for the PR, looks great :+1:  Do you think this can be an initial cause of https://bugzilla.redhat.com/show_bug.cgi?id=1508006#c0 (duplicate inventory when EMS is was deleted propably during its refresh)?

mansam:
@aufi I could see it being related. There's currently a bug where the NetworkManager is refreshed first so relationships aren't formed between network objects and cloud tenants that this fixes, so I could see there being other timing related problems that this might help with.

Ladas:
@mansam @aufi hm, not sure if this is a cause of the bug. It looks like the original networks were deleted (since those should be under transaction with deleting ems), but then the refresh completed and created the networks again?

Somebody else was adding guards, that you can't delete manager if it has jobs running I think.

mansam:
@Ladas Ah, deleting the provider while it's currently refreshing would certainly cause that.

Ladas:
@mansam I think the proof could be that the networks do have ems_id filled, but that ems_id doesn't exist. So again, for 5.0, we should fix this, but the first line of defense should be DB foreign_key on ems_id :-)

app/models/manageiq/providers/openstack/cloud_manager/refresher.rb
```diff
@@ -68,6 +68,13 @@ def preprocess_targets
           end
         end
 
+        # if the cloud manager is due to be refreshed, make sure it's the first target.
+        cloud_manager_index = all_targets.find_index { |t| t.kind_of? ManageIQ::Providers::Openstack::CloudManager }
```
Ladas: isn't this stored as 
```ruby
@targets_by_ems_id {
  1 => [cloud_ems, target1, target2] ,
  2 => [network_ems],
  3 => [cinder_ems],
  4 => [swift_ems]
```

then the sorting like this would not help I think

mansam: hrm, could be. I'll recheck.

gildub:
@Ladas,  do you mean foreign key with cascade one? Effectively that's seems inevitable to keep data integrity.

gildub:
@aufi, breaking data integrity is *bad*, I realized that cascade delete is not activated at DB level (foreign key with cascade on) as it's only setup at the Rails level (`:dependent => :destroy`). Therefore I'm keep to add that to the features list for CF 5.0. What do you think?

app/models/manageiq/providers/openstack/cloud_manager/refresher.rb
mansam: Looks like you're right, but in that case there's no sorting that could help here. `ems_refresher_mixin.rb:refresh` would have to do the sorting, and adding a special case there seems bad.

Ladas: So, it is pretty horrible, but you might be able to change the @targets_by_ems_id here. Probably before returning it, you would @targets_by_ems_id = {}, and then write it there in a new order. There is @ems_by_ems_id that you can use, with combination of keys of targets_by_ems_id, s you don;t have to query for EMS.

Ladas:
@gildub priority for 5.0 is to have unique keys (to avoid duplication and to be able to store the data in parallel). 

Foreign_keys would be nice to have, but since have share the tables with STI, we have many places where 1 provider doesn't fill the foreign key, so we will need strategies to approach that.

Then cascade delete, that will be probably the hardest, since that would just break usage of Rails hooks on some places.

agrare:
@mansam can you update your queue_name_for_ems to match the new format from? https://github.com/ManageIQ/manageiq-providers-amazon/pull/361/files

simaishi:

```
$ git log -1
commit 7f85c825db866f3f866f6fe8983e8126282193e6
Date:   Fri Dec 8 09:32:56 2017 -0500

    Merge pull request #154 from mansam/consolidate-openstack-refresh-workers
    
    Consolidate openstack refresh workers
    (cherry picked from commit 9295127e97d1fe960ae5ec50466b35e01cc8c4c3)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1524590
```

PR #160, opened 2017-11-15
[WIP] Validate presence of ems_id for vnfd template

Ladas:
Validate presence of ems_id for vnfd template

Fixes BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1437640

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/vnfd_template.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/vnfd_template.rb
@@ -5,6 +5,8 @@ class ManageIQ::Providers::Openstack::CloudManager::VnfdTemplate < ::Orchestrati
   before_update :raw_update, :if => :remote_proxy?
   before_destroy :raw_destroy, :if => :remote_proxy?
 
+  validates :ems_id, :presence => true
+
   def raw_create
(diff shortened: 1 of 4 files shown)


Ladas:
before 
screenshot from 2017-03-30 11_29_00 1

after:
screenshot from 2017-11-15 10-12-12

Ladas:
@martinpovolny @mzazrivec where do I set a translation for :ems_id field, now it translates as Ems, I want it as a Provider

Seems like the usual way is putting this in controller though, which is horrible https://github.com/Ladas/manageiq-ui-classic/blob/5c6903b88e26c538192972bbc28719267d41d4d7/app/controllers/catalog_controller.rb#L1105

But is that because we can't set the correct translation using model validations?

mzazrivec:
@Ladas look into `locale/en.yml`, `activerecord.attributes` subtree.

Ladas:
@mzazrivec ah, thanks, I must be blind I would swear that file was empty when I checked it :-)

mansam:
@Ladas does this need any other changes or is it ready to merge?

Ladas:
@mansam I haven't figured out the locale string I need to set, but that will be a core PR. Lets probably wait till I push that, so I don't forget. :-)

mansam:
@Ladas sounds good :)

PR #161, opened 2017-11-15
add error message if FIP assigned to router

alexander-demichev:
Router can't be deleted if has assigned floating ips.
https://bugzilla.redhat.com/show_bug.cgi?id=1470754


alexander-demichev:
https://github.com/ManageIQ/manageiq-ui-classic/pull/2737

aufi:
After more detailed reading of the BZ, it looks we should lookup instances with Floating IP connected via the Router [1]. _Router.floating_ips_ method returns list of floating_ip on the connected network, we propably could filter just FIPs which are associated with a VM on such network.

@alexander-demichev please include steps how this should be tested (positive and negative example), thanks!

[1] http://paste.openstack.org/show/626489/

aufi:
Looks good to me, merging!

aufi:
BZ doesn't specify G-release as target release, feel free update it if needed.

PR #173, opened 2017-12-05
Fix Provisioning of disconnected VolumeTemplate

aufi:
Since provisioning from OpenStack Volumes was enabled, situation when original
Volume template can be disconnected from EMS by refresher can happen and it
doesn't look to be possible fix it on OpenStack side.

This PR allows EMS lookup fallback to options[:src_ems_id] when getting EMS
from source object fails.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1518381

Links
----------------

* BZ https://bugzilla.redhat.com/show_bug.cgi?id=1518381
* PRs adding provision from OpenStack Volumes for better understanding https://github.com/ManageIQ/manageiq/pull/16066 and  https://github.com/ManageIQ/manageiq-providers-openstack/pull/104

Steps for Testing/QA
-------------------------------
Vm provision process from OpenStack Volume should pass even refresh of given provider runs faster than provisioning process.

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/provision/cloning.rb
@@ -1,4 +1,13 @@
 module ManageIQ::Providers::Openstack::CloudManager::Provision::Cloning
+  def find_destination_in_vmdb(ems_ref)
+    super
+  rescue NoMethodError => ex
+    # TODO: this should not be needed after we update refresh to not disconnect VmOrTemplate from EMS
+    _log.debug("Unable to find Provison Source ExtmanagementSystem: #{ex}")
+    _log.debug("Trying use attribute src_ems_id=#{options[:src_ems_id].try(:first)} instead.")
+    vm_model_class.find_by(:ems_id => options[:src_ems_id].try(:first), :ems_ref => ems_ref)
+  end
+
   def do_clone_task_check(clone_task_ref)
     connection_options = {:tenant_name => options[:cloud_tenant][1]} if options[:cloud_tenant].kind_of?(Array)
     source.with_provider_connection(connection_options) do |openstack|


app/models/manageiq/providers/openstack/cloud_manager/provision/cloning.rb
```diff
@@ -1,4 +1,12 @@
 module ManageIQ::Providers::Openstack::CloudManager::Provision::Cloning
+  def find_destination_in_vmdb(ems_ref)
+    vm_model_class.find_by(:ems_id => source.ext_management_system.id, :ems_ref => ems_ref)
+  rescue NoMethodError => ex
+    _log.debug("Unable to find Provison Source ExtmanagementSystem: #{ex}")
+    _log.debug("Trying use attribute src_ems_id=#{options[:src_ems_id].try(:first)} instead.")
+    vm_model_class.find_by(:ems_id => options[:src_ems_id].try(:first), :ems_ref => ems_ref)
```
Ladas: I wonder if we can just use this call right away? Or is there a case when options[:src_ems_id] is nil?

app/models/manageiq/providers/openstack/cloud_manager/provision/cloning.rb
```diff
@@ -1,4 +1,12 @@
 module ManageIQ::Providers::Openstack::CloudManager::Provision::Cloning
+  def find_destination_in_vmdb(ems_ref)
+    vm_model_class.find_by(:ems_id => source.ext_management_system.id, :ems_ref => ems_ref)
```
Ladas: if we will keep this ,maybe we can just call ```super```

app/models/manageiq/providers/openstack/cloud_manager/provision/cloning.rb
aufi: I'm not aware of such case, but I'd prefer keep original method using _source_ to feel safe..

app/models/manageiq/providers/openstack/cloud_manager/provision/cloning.rb
aufi: Make sense, thanks, will update it.

app/models/manageiq/providers/openstack/cloud_manager/provision/cloning.rb
Ladas: right, it bugs me a bit that this will basically fail everytime for volume_template, so maybe we should not use it?

Ladas:
@aufi also, can you add todo, that this will not be needed when we do soft delete of VmOrTemplate without removing the vm_or_template foreign_key?

aufi:
Tested locally and should be **working** (kind of proof: checked timestamps - VolumeTemplate was disconnected from EMS approx 20 seconds before last update of provisioned Vm and provision succeed).

@miq-bot rm_label wip

simaishi:

```
$ git log -1
commit 4bdd3c5fc4de6e9c87be41ff25ffe3dc9fdcf3be
Author: Samuel Lucidi <mansam@csh.rit.edu>
Date:   Wed Dec 13 13:44:07 2017 -0500

    Merge pull request #173 from aufi/fix_disconnected_volume_template
    
    Fix Provisioning of disconnected VolumeTemplate
    (cherry picked from commit e14333a5777fdaccd9c93988dae1ef8a62168e53)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1526062
```

PR #175, opened 2017-12-07
Improve Targeted Refresh for Cloud and Network managers

mansam:
https://bugzilla.redhat.com/show_bug.cgi?id=1519214

I'm basing the events I catch on ceilometer event definitions, but I'm not sure that I have a complete list. @aufi, do you know if there is a better source than https://github.com/openstack/ceilometer/blob/master/ceilometer/pipeline/data/event_definitions.yaml ?


app/models/manageiq/providers/openstack/network_manager/event_target_parser.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  private
+
+  # Parses list of ManagerRefresh::Target(s) out of the given EmsEvent
+  #
+  # @param ems_event [EmsEvent] EmsEvent object
+  # @return [Array] Array of ManagerRefresh::Target objects
+  def parse_ems_event_targets(ems_event)
+    target_collection = ManagerRefresh::TargetCollection.new(:manager => ems_event.ext_management_system, :event => ems_event)
```
Ladas: Hm, the separate event catcher for network manager complicates stuff.

I think we need to set here parent_manager:
ManagerRefresh::TargetCollection.new(:manager => ems_event.ext_management_system.parent_manager, :event => ems_event)

Since all of the targeted refresh runs in the cloud manager and the targets are bucketed based on manager. Storage wise, this doesn't matter, since each persistor assigns the right manager for saving. So this is more for which ems and therefore worker should process it.

Ladas:
@mansam so for ```/System/event_handlers/refresh``` in automate to work (to parse some targets and send them for refresh), we need to cover all events listed in https://github.com/Ladas/manageiq-content/tree/33e254dd0fcc2e47f379130f8dcf21b51ab30e10/content/automate/ManageIQ/System/Event/EmsEvent/OpenStack.class

But if we could base it on generic definition like https://github.com/openstack/ceilometer/blob/master/ceilometer/pipeline/data/event_definitions.yaml, that would be awesome. That would mean it will work for more events than defined in our automate.

mansam:
@Ladas @aufi The list of events from manageiq-content is very helpful, but unfortunately I'm not sure it's sufficient. If I understand correctly, we capture events from Ceilometer, and Ceilometer changes the format of the events from the actual services quite a bit. Without more detail about the actual event definitions we get from Ceilometer it's not really possible to parse the events (since I don't know what fields are available, and they are not consistently named), and not every event in manageiq-content has an equivalent in the default Ceilometer event definition yaml. I've done a lot of looking over the past couple days and in the past when first working on this and am coming up empty handed for better documentation.

Ladas:
@mansam right so, AWS has no docs for the event payloads, so what I ended up doing was 

+++ b/app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
@@ -190,6 +190,13 @@ class ManageIQ::Providers::Amazon::CloudManager::EventCatcher::Stream
       return
     end
 
+    ####
+    File.open(Rails.root.join("events", event["eventType"]+".json" ), "w") do |file|
+      file.write(message.body)
+    end
+    ####
+
+
     $log.info("#{log_header} Found SNS Message with message type #{event["eventType"]} coming from #{event[:event_source]}")

------

Then I was clicking around, recording the real events. Then I am using the payload of the real events for the specs, to make sure I am parsing the right targets. Also, for CloudWatch events, the ids are in a form of "instance_id, image_id", so I don't even care what kind of event it is. Any event referencing instance_id, will produce VM target.

-------

So yeah, if the event payloads are different in AMQP/Ceilometer/WhatIsTheNewServiceName, than this kind of sucks. 

The AWS has similar issue, sometimes the event has VpcID, other times vpcId https://github.com/Ladas/manageiq-providers-amazon/blob/4c66893fd1a91dc5fd3aa259a7c0ddd044c65576/app/models/manageiq/providers/amazon/cloud_manager/event_target_parser.rb#L96

mansam:
@Ladas Hrm, good point that if the IDs are named in a recognizable fashion then it doesn't matter at all what event it came from.

blomquisg:
@mansam, can you rebase and repush?

mansam:
@blomquisg updated

blomquisg:
@Ladas, do you still have outstanding changes you're looking for?  If not, can you clear/change your review?

app/models/manageiq/providers/openstack/cloud_manager/refresher.rb
```diff
@@ -1,75 +1,14 @@
...
-
-    def parse_legacy_inventory(ems)
-      ManageIQ::Providers::Openstack::CloudManager::RefreshParser.ems_inv_to_hashes(ems, refresher_options)
-    end
-
+  class Openstack::CloudManager::Refresher < ManageIQ::Providers::BaseManager::ManagerRefresher
     def save_inventory(ems, target, inventory_collections)
       super
       EmsRefresh.queue_refresh(ems.network_manager) if target.kind_of?(ManageIQ::Providers::BaseManager)
       EmsRefresh.queue_refresh(ems.cinder_manager) if target.kind_of?(ManageIQ::Providers::BaseManager)
```
Ladas: so maybe, we will have to do a cinder full refresh always here, we should track that as a separate issue though

app/models/manageiq/providers/openstack/refresh_parser_common/helper_methods.rb
```diff
@@ -3,15 +3,15 @@ module Openstack
     module RefreshParserCommon
       module HelperMethods
         def openstack_admin?
-          ::Settings.ems.ems_refresh.openstack.try(:is_admin)
+          ::Settings.ems_refresh.openstack.try(:is_admin)
```
Ladas: a nit for other PR probably:

we usually abstract ```::Settings.ems_refresh.openstack``` as a method , in new refresh parser it available on collector https://github.com/Ladas/manageiq/blob/b9e6b294604d744f59e3bf270e16264f067fb85a/app/models/manager_refresh/inventory/collector.rb#L11

mansam: I agree that it should be abstracted out in the future.

spec/models/manageiq/providers/openstack/cloud_manager/refresher_rhos_juno_spec.rb
```diff
@@ -49,16 +49,16 @@
 
         assert_common
       end
-      ::Settings.ems.ems_refresh.openstack.is_admin = false
-      ::Settings.ems.ems_refresh.openstack_network.is_admin = false
+      ::Settings.ems_refresh.openstack.is_admin = false
```
Ladas: as commented in other PR, this should use 

https://github.com/Ladas/manageiq-providers-amazon/blob/7b1a7226dea0566fd17d3b06271782e213bf4969/spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb#L47

Best to clean them all up in a refactoring PR

mansam:
@Ladas Thanks for the review. In the future for APIs that support filtering correctly, it might make sense to make a list request that is filtered by the desired IDs. Unfortunately filtering doesn't seem to be supported everywhere (like networking, it seems).

aufi:
LGTM, merging!

simaishi:

```
$ git log -1
commit 1c1ba1a8be210bbcfd109b777cedb89af0e507e7
Date:   Wed Jan 10 10:06:54 2018 +0100

    Merge pull request #175 from mansam/event-target-parser-network-manager
    
    Improve Targeted Refresh for Cloud and Network managers
    (cherry picked from commit 92aed61171736d5d96d11bbf25a609347d29727b)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1533225
```

Issue #174, opened 2017-12-07
Disclosure of sensitive information after registering openstack cloud provider 


dikonoor:
Password used for registering OpenStack Cloud Provider displayed in the evm.log. This is a security exposure and might be present in the older releases of ManageIQ as well and 'might' in all likelihood need a public CVE (common Vulnerability Evaluation) as this would fall under 'Information Exposure'. The evm.log has a read all file permissions, which basically means that any user logged into the system where ManageIQ is installed will be able to retrieve credentials of the Cloud Provider user. 

As ManageIQ requires the credentials of a user with admin role on OpenStack, this basically implies that a hacker can grab the credentials, log into OpenStack and do just anything with the underlying infrastructure.

Even though this was found while registering OpenStack Cloud Provider, it is possible that this exposure exists for all other supported cloud providers. This needs to be verified. As I don't use other cloud providers, I will not be able to verify this.

**Environment Details**
ManageIQ Release - Gaprinshdanvili (Beta)
OpenStack Pike

**Steps to Reproduce**
1. Register a new OpenStack Cloud Provider to Manage IQ.
2. Credentials are valid and registration is successful.
3. Wait for few minutes and check the log and you should find the following:

[----] I, [2017-12-06T22:56:41.069274 #89849:fb10d0]  INFO -- : MIQ(MiqQueue#m_callback) Message id: [96433], Invoking Callback with args: ["Finished", "ok", "Message delivered successfully", "#<Fog::Compute::OpenStack::Real:155741740 @openstack_auth_url=\"https://x.x.x.x:5000/v3/auth/tokens\" 
@openstack_username=\"user1\" **@openstack_api_key=\"foobar123\"** 

@openstack_endpoint_type=\"publicURL\" @openstack_tenant=\"abc-default\" @openstack_project_name=\"abc-default\" @openstack_project_domain_id=\"default\" @openstack_user_domain_id=\"default\" @openstack_region=\"RegionOne\" @auth_token=\"gAAAAABaKLv5h3mHgv0YGGrZPYueqhedGQ8nvwMPJo80hHs2-pQinCPKE6ebPppKAnwwHn-nA1zmPzhEONMQXj_cAW6AbQ7yKWfzCmi51TGLc_U8o..."]

4. Please note that foobar123 is the password for user1 that I had used for registration. 

[...]

PR #176, opened 2017-12-08
Return true instead of connection object for creds validation

aufi:
A validate credentials action on Cloud Provider uses raw_connect method on ManagerMixin via MiqQueue,
that leads to exposing credentials to log entry related to MiqQueue handling.

Changed to return true instead of connection object from the method.

Related github issue: https://github.com/ManageIQ/manageiq-providers-openstack/issues/174

### DIFF
--- a/app/models/manageiq/providers/openstack/manager_mixin.rb
+++ b/app/models/manageiq/providers/openstack/manager_mixin.rb
@@ -12,6 +12,7 @@ module ManageIQ::Providers::Openstack::ManagerMixin
   #
   module ClassMethods
     def raw_connect(password, params, service = "Compute")
+      # TODO: this method seems to be used only for Provider credentials validation
       ems = new
       ems.name                   = params[:name].strip
       ems.provider_region        = params[:provider_region]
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/openstack/manager_mixin.rb
```diff
@@ -12,6 +12,7 @@ module ManageIQ::Providers::Openstack::ManagerMixin
   #
   module ClassMethods
     def raw_connect(password, params, service = "Compute")
+      # TODO: this method seems to be used only for Provider credentials validation
```
Ladas: isn't this used also to with_provider connection, etc? So for all actions?

aufi: There is a method with the same name in OpenStack Handle, which is used for accessing OpenStack. I didn't find other usage, but it would be great have more eyes checking it!

Fryguy: I would expect raw_connect to return the object...and with_provider_connection definitely uses the object returned by connect (which should be obtained from raw_connect.  I don't think returning true here is the right approach.

aufi:
Agree on need for more general solution, closing this PR. Discussion should will follow on the issue #174

PR #182, opened 2017-12-18
Ensure that subnets are dissociated from routers in the ManageIQ inventory when their interfaces are removed on the OSP side

mansam:
Defaults the `network_router` field on subnets to `nil` during refresh. If the interface linking subnet to router still exists on the OSP side, the association will be restored while updating ports.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1449260

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/parser/network_manager.rb
+++ b/app/models/manageiq/providers/openstack/inventory/parser/network_manager.rb
@@ -48,6 +48,10 @@ class ManageIQ::Providers::Openstack::Inventory::Parser::NetworkManager < Manage
         subnet.parent_cloud_subnet = persister.cloud_subnets.lazy_find(s.attributes["vsd_managed"])
         subnet.cloud_tenant = persister.cloud_tenants.lazy_find(s.tenant_id)
         subnet.cloud_network = network
+        # Always default this to nil so that if a interface that connects this
+        # to a router are removed, then the association is properly broken when
+        # the subnet is refreshed.
+        subnet.network_router = nil
       end
     end
   end
--- a/app/models/manageiq/providers/openstack/inventory/parser/network_manager.rb
+++ b/app/models/manageiq/providers/openstack/inventory/parser/network_manager.rb
@@ -48,10 +48,6 @@ class ManageIQ::Providers::Openstack::Inventory::Parser::NetworkManager < Manage
         subnet.parent_cloud_subnet = persister.cloud_subnets.lazy_find(s.attributes["vsd_managed"])
         subnet.cloud_tenant = persister.cloud_tenants.lazy_find(s.tenant_id)
         subnet.cloud_network = network
-        # Always default this to nil so that if a interface that connects this
-        # to a router are removed, then the association is properly broken when
-        # the subnet is refreshed.
-        subnet.network_router = nil
       end
     end
   end
--- a/app/models/manageiq/providers/openstack/network_manager/network_port.rb
+++ b/app/models/manageiq/providers/openstack/network_manager/network_port.rb
@@ -1,2 +1,12 @@
 class ManageIQ::Providers::Openstack::NetworkManager::NetworkPort < ::NetworkPort
+
+  def disconnect_port
+    # Some ports link subnets to routers, so
+    # sever that association if the port is removed
+    cloud_subnets.each do |subnet|
+      subnet.network_router = nil
+      subnet.save!
+    end
+    delete
+  end
 end
--- a/app/models/manageiq/providers/openstack/network_manager/network_port.rb
+++ b/app/models/manageiq/providers/openstack/network_manager/network_port.rb
@@ -1,5 +1,4 @@
 class ManageIQ::Providers::Openstack::NetworkManager::NetworkPort < ::NetworkPort
-
   def disconnect_port
     # Some ports link subnets to routers, so
     # sever that association if the port is removed
--- a/spec/models/manageiq/providers/openstack/network_manager/cloud_subnet_refresh_spec.rb
+++ b/spec/models/manageiq/providers/openstack/network_manager/cloud_subnet_refresh_spec.rb
@@ -27,6 +27,34 @@ describe ManageIQ::Providers::Openstack::NetworkManager::CloudSubnet do
       expect(CloudSubnet.first.network_router_id).to be(nil)
     end
 
+    it "targeted refresh shouldn't remove subnet link to router unless the interface is deleted" do
+      setup_mocked_collector
+      ::Settings.ems_refresh.openstack_network.allow_targeted_refresh = true
+
+      EmsRefresh.refresh(@ems)
+      expect(CloudSubnet.count).to eq(1)
+      expect(CloudNetwork.count).to eq(1)
+      expect(NetworkPort.count).to eq(1)
+      expect(NetworkRouter.count).to eq(1)
+      expect(CloudSubnet.first.network_router_id).to eq(NetworkRouter.first.id)
+
+      target = ManagerRefresh::Target.new(
+          :manager     => @ems.parent_manager,
+          :association => :cloud_networks,
+          :manager_ref => {
+            :ems_ref => "cloud_network_1"
+          }
+        )
+      setup_mocked_targeted_collector
+      EmsRefresh.refresh(target)
+      expect(CloudSubnet.count).to eq(1)
+      expect(CloudNetwork.count).to eq(1)
+      expect(NetworkPort.count).to eq(1)
+      expect(NetworkRouter.count).to eq(1)
+      expect(CloudSubnet.first.network_router_id).to eq(NetworkRouter.first.id)
+      ::Settings.ems_refresh.openstack_network.allow_targeted_refresh = false
+    end
+
     def mocked_network_ports
       [OpenStruct.new(
         :id              => "network_port_1",
@@ -62,6 +90,7 @@ describe ManageIQ::Providers::Openstack::NetworkManager::CloudSubnet do
 
     def setup_mocked_collector
       allow_any_instance_of(ManageIQ::Providers::Openstack::Inventory::Collector::CloudManager).to receive(:availability_zones).and_return([])
+      allow_any_instance_of(ManageIQ::Providers::Openstack::Inventory::Collector::CloudManager).to receive(:cloud_services).and_return([])
       allow_any_instance_of(ManageIQ::Providers::Openstack::Inventory::Collector::CloudManager).to receive(:vms).and_return([])
       allow_any_instance_of(ManageIQ::Providers::Openstack::Inventory::Collector::CloudManager).to receive(:tenants).and_return([])
       allow_any_instance_of(ManageIQ::Providers::Openstack::Inventory::Collector::NetworkManager).to receive(:orchestration_stacks).and_return([])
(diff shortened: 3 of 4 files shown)


app/models/manageiq/providers/openstack/inventory/parser/network_manager.rb
```diff
@@ -48,6 +48,10 @@ def cloud_networks
         subnet.parent_cloud_subnet = persister.cloud_subnets.lazy_find(s.attributes["vsd_managed"])
         subnet.cloud_tenant = persister.cloud_tenants.lazy_find(s.tenant_id)
         subnet.cloud_network = network
+        # Always default this to nil so that if a interface that connects this
+        # to a router are removed, then the association is properly broken when
+        # the subnet is refreshed.
+        subnet.network_router = nil
```
aufi: Just quick question @mansam, where is the code which connects subnet to a network router?

Ladas: hm, this might not be the best. I think that targeted refresh would be then nullifying the relation (unless it always refreshes together with the interface)

Best would be to just define dependent => :nullify on the relation? That way if router is deleted, the relation disappears. But we probably can't do that on the network_port? Hm we could do custom disconnect method that would nullify it when port is being deleted.

spec/models/manageiq/providers/openstack/network_manager/cloud_subnet_refresh_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    before do
+      parent_ems = FactoryGirl.create(:ems_openstack_with_authentication)
+      @ems = parent_ems.network_manager
+      EvmSpecHelper.local_miq_server(:zone => Zone.seed)
+    end
+
+    it "removes subnet link to router if the interface is deleted" do
+      setup_mocked_collector
+
+      EmsRefresh.refresh(@ems)
```
Ladas: we will also need targeted refresh spec, checking refresh of the subnet is not nullifying the relation

app/models/manageiq/providers/openstack/inventory/parser/network_manager.rb
mansam: It doesn't seem like we could do dependent => :nullify since the port is the thing that gets deleted, but the relationship that has to change is the subnet -> router relationship. I looked into using a custom destroy hook, but `save_inventory_multi` uses `delete` so the destroy hook never gets called.

Ladas: right, save_inventory_multi is for old refresh (that doesn't have targeted refresh)

for having custom delete method you need to specify it like https://github.com/Ladas/manageiq/blob/0815e847bd63282129eee90869bf9bf321c0224d/app/models/manager_refresh/inventory_collection_default.rb#L7

app/models/manageiq/providers/openstack/network_manager/network_port.rb
```diff
@@ -1,2 +1,12 @@
 class ManageIQ::Providers::Openstack::NetworkManager::NetworkPort < ::NetworkPort
+
+  def disconnect_port
+    # Some ports link subnets to routers, so
+    # sever that association if the port is removed
+    cloud_subnets.each do |subnet|
```
Ladas: These might be the wrong subnets? I think we fill these in only for Vm's network ports?

We should look at ```belongs_to :device, :polymorphic => true``` , that relations for subnets is being fixed in https://github.com/ManageIQ/manageiq-providers-openstack/pull/188/files (could you verify it actually returns subnet here?)

Also we should check if we can conventionalize this by subnet.network_router = nil, only if the router is the router mentioned in the NetowrkPort (we might not store this). To avoid nullifying when the subnet was just connected somewhere else.

Hm, this approach is getting more complex than it should.

Ladas: maybe we should even remodel the relation, to not use ```subnet.network_router```, but to just use the network_port and sending the query through it

mansam: The router gets returned out of `find_device_object` correctly there. Both `parser.network_ports` and `find_device_object` use the `network_port.fixed_ips` to associate subnets so I was thinking those were the correct subnets. The tests pass correctly, at least. (though perhaps this catches unrelated subnets? I don't have a good mental model of all the circumstances in which subnets get attached to ports and routers.)

Ladas: Right so there is NetworkPort which is interface of a VM (Vm is the device) and there is a NetworkPort that is interface of router. I think the network_port.cloud_subnets works only for Vm's interfaces.

spec/models/manageiq/providers/openstack/network_manager/cloud_subnet_refresh_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+      expect(CloudSubnet.count).to eq(1)
+      expect(CloudNetwork.count).to eq(1)
+      expect(NetworkPort.count).to eq(0)
+      expect(NetworkRouter.count).to eq(1)
+      expect(CloudSubnet.first.network_router_id).to be(nil)
+    end
+
+    it "targeted refresh shouldn't remove subnet link to router unless the interface is deleted" do
+      setup_mocked_collector
+      ::Settings.ems_refresh.openstack_network.allow_targeted_refresh = true
```
Ladas: btw. there is a helper for stubbing settings

https://github.com/Ladas/manageiq-providers-amazon/blob/7b1a7226dea0566fd17d3b06271782e213bf4969/spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb#L47

Does the direct assignment work?

Ladas: hm, the ```::Settings.ems_refresh.openstack_network.allow_targeted_refresh = false``` might bite us, since specs run in parallel in CI. So we should use the stub that is isolated to 1 spec

app/models/manageiq/providers/openstack/network_manager/network_port.rb
mansam: It does appear to work here, so that must not be the case that it only works on VMs.

spec/models/manageiq/providers/openstack/network_manager/cloud_subnet_refresh_spec.rb
mansam: The direct assignment does work, but you're right that there's risk of cross contamination since the tests are in parallel. I'll adjust to use the settings stub.

Ladas: right, probably do another cleanup PR, since it's in all specs

app/models/manageiq/providers/openstack/network_manager/network_port.rb
Ladas: hm, the spec looks correct. So it must work :-)

Ladas: Though a spec case where this might break is that the targeted refresh contains both removing of old interface and adding new one. That might cause CloudSubnet.first.network_router_id to be nil, while it should point to a new router. Could you add this spec scenario?

mansam: Good catch, I'll write another spec.

mansam: Wrote a spec for the situation where the port is removed and a new one is created in the same refresh. `router_id` becomes nil as you suspected, but I don't understand why exactly. 

I inspected what was happening when the custom delete is called, and it looks like at that time the replacement NetworkPort doesn't exist yet. So if the old NetworkPort is the only one in the database at the time, wouldn't the new NetworkPort have to be created afterwards and set the id again?

mansam:
This comment got buried by github after I touched the code the review was on, so I am repeating it here:

@Ladas I wrote a spec for the situation where the port is removed and a new one is created in the same refresh. network_router_id becomes nil as you suspected, but I don't understand why exactly.

I inspected what was happening when the custom delete is called, and it looks like at that time the replacement NetworkPort doesn't exist yet. So if the old NetworkPort is the only one in the database at the time, wouldn't the new NetworkPort have to be created afterwards and set the id again?

Ladas:
@mansam right, so this is because of the ordering in the graph. The subnet table is processed first and is storing the new relation to new router. After that, the network_ports are processed, where part of it we delete the network port and touch a subnet that was already saved correctly.

app/models/manageiq/providers/openstack/network_manager/network_port.rb
```diff
@@ -1,2 +1,11 @@
 class ManageIQ::Providers::Openstack::NetworkManager::NetworkPort < ::NetworkPort
+  def disconnect_port
+    # Some ports link subnets to routers, so
+    # sever that association if the port is removed
+    cloud_subnets.each do |subnet|
+      subnet.network_router = nil
```
Ladas: So to correctly cover the latest spec. We would need info on network_port, to what router it relates to (ems_ref). Then here, we could check that if it's still the same router, we can nullify it. If it's a different router we should not touch it.

But it's kind of clunky. I think in the end, we should look into how to change the modeling. So that 1 model doesn't have to touch relations of other models. So e.g. doing the actual relation through NetworkPort and deleting the subnet.network_router foreign key.

mansam: The subnet gets network_router set on it as part of parsing network ports. I take your comment about timing mean that the whole inventory is collected and parsed, and then subnets are saved before network ports, resulting in the property being cleared?

We can't compare whether the router is the same, because the user could delete and recreate the same router relationship. That would shrink the edge case though. I agree that the appropriate long term solution is to fix the DB modeling, but a fix that touches the DB is probably too high impact for this part of the release cycle.

chessbyte: > We can't compare whether the router is the same, because the user could delete and recreate the same router relationship

@mansam @aufi that sounds like very rare, race condition kind of situation that we can fix with the more elegant solution involving DB Schema change.  Would the current simpler fix cover 99.99% of use cases?

Ladas: yeah, just comparing will probably get to 99% cases being correct. And the 1% will be fixed by next full refresh.

Ladas: we could also just set the test as pending, the current state is like 95% correct, where the 5% will be fixed by full refresh. Setting/changing router interfaces is a very rare operation, comparing to others, so it should not be an issue.

Ladas:
@mansam the CI failure looks unrelated

aufi:
The code looks good to me, I'm afraid of failing tests (re-triggered test and still failing).

Ladas:
@aufi the failing tests are being fixed in https://github.com/ManageIQ/manageiq/pull/16824, I think you can merge all PRs, that have only these 3 failures

aufi:
@Ladas Ok, thanks for info. LGTM, merging!

simaishi:

```
$ git log -1
commit fcb6a7bb6297e9893fab361aa7d243577ebdf320
Date:   Tue Jan 16 17:04:38 2018 +0100

    Merge pull request #182 from mansam/disconnect-subnets-from-routers-during-refresh
    
    Ensure that subnets are dissociated from routers in the ManageIQ inventory when their interfaces are removed on the OSP side
    (cherry picked from commit 5acca17ce02c444fb77812b777ee74c92d7b4de2)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1535155
```

Issue #164, opened 2017-11-23
Wrong index for lazy find


Ladas:
@mansam I've noticed these

https://github.com/Ladas/manageiq-providers-openstack/blob/ce4cd9acf7c2c2f236503fa6bb972b8f4aee7d1e/app/models/manageiq/providers/openstack/inventory/parser/network_manager.rb#L32

https://github.com/Ladas/manageiq-providers-openstack/blob/ce4cd9acf7c2c2f236503fa6bb972b8f4aee7d1e/app/models/manageiq/providers/openstack/inventory/parser/network_manager.rb#L130

the orchestration_stack_by_resource_id returns a Fog object, but we need ems_ref string for the lazy find.

----

I am working on the internal indexes refacotrings, so then, we would see these fails.

mansam:
Nice catch, thanks Ladas.

PR #183, opened 2018-01-03
Use proper nested references in parser

Ladas:
Use proper nested references in parser. Adding nested
lazy_finds to correctly model a complex link.
  
Fixes  https://github.com/ManageIQ/manageiq-providers-openstack/issues/164

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/parser/network_manager.rb
+++ b/app/models/manageiq/providers/openstack/inventory/parser/network_manager.rb
@@ -29,7 +29,9 @@ class ManageIQ::Providers::Openstack::Inventory::Parser::NetworkManager < Manage
       network.vlan_transparent = n.attributes["vlan_transparent"]
       network.maximum_transmission_unit = n.attributes["mtu"]
       network.cloud_tenant = persister.cloud_tenants.lazy_find(n.tenant_id)
-      network.orchestration_stack = persister.orchestration_stacks.lazy_find(collector.orchestration_stack_by_resource_id(n.id))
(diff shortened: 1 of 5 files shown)


Ladas:
@mansam this shows how the new interface looks like https://github.com/ManageIQ/manageiq-providers-amazon/pull/383#discussion_r159257932 and it has validations making sure we are passing correct keys and values

Ladas:
@mansam passes specs, it's ready for review

app/models/manageiq/providers/openstack/inventory/parser/network_manager.rb
```diff
@@ -29,7 +29,9 @@ def cloud_networks
       network.vlan_transparent = n.attributes["vlan_transparent"]
       network.maximum_transmission_unit = n.attributes["mtu"]
       network.cloud_tenant = persister.cloud_tenants.lazy_find(n.tenant_id)
-      network.orchestration_stack = persister.orchestration_stacks.lazy_find(collector.orchestration_stack_by_resource_id(n.id))
+      network.orchestration_stack = persister.orchestration_stacks_resources.lazy_find(
+        collector.orchestration_stack_by_resource_id(n.id).try(:physical_resource_id), :key => :stack
```
Ladas: @mansam so not sure if this is anywhere else, but usually each entity relates to resource, so we wan't lazy_find on resource. But since we want to store stack, we lazily access the stack of the resource by ```:key => :stack```

mansam:
Thanks @ladas, the changes look good.

Ladas:
@mansam this should fix the failing CI

PR #184, opened 2018-01-03
Return empty AR relation instead of nil for ::InfraManager#cloud_tenants

lpichler:
InfraProvider don't have relation to cloud_tenants - there was
created only empty method returning nil to preserve consitency - but on
some places is expected AR relation object.

For example screen add new subnet, there are listed network managers and
their parent manager's cloud_tenants but they can be infra manager whithout
cloud_tenants.

Such place, where was expected AR Relation object is here:
https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/subcollections/cloud_tenants.rb#L5

which was reached for example by this request:

http://localhost:3000/api/providers/:id/cloud_tenants but it fails later
when object has been infra manager and the method cloud_tenants_query_resource
was returning nil.

@miq-bot add_label graprindashvili/yes, bug

cc @gtanzillo 
@miq-bot assign @aufi 

### Reproducer
zytkbnyvnw

### Links
it was found during BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1463422

### DIFF
--- a/app/models/manageiq/providers/openstack/infra_manager.rb
+++ b/app/models/manageiq/providers/openstack/infra_manager.rb
@@ -31,6 +31,7 @@ class ManageIQ::Providers::Openstack::InfraManager < ManageIQ::Providers::InfraM
   end
 
   def cloud_tenants
+    self.class.none
   end
 
   def host_aggregates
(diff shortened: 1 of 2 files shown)


aufi:
Thanks for PR @lpichler! LGTM, merging.

simaishi:

```
$ git log -1
commit 8c59c3a3cfdc376db5b3fdeedf1b43ee0547580b
Date:   Thu Jan 4 10:29:48 2018 +0100

    Merge pull request #184 from lpichler/return_empty_for_cloud_tenants
    
    Return empty AR relation instead of nil for ::InfraManager#cloud_tenants
    (cherry picked from commit 37acd49c820af306d5ed152ffbfcf1c6609deaad)
```

PR #186, opened 2018-01-04
Use only hypervisor hostname to match infra host with cloud vm

petrblaho:
Hosts are stored in db with only hostname part of hostname.domain which
we can see in OpenStack.
That caused issue with matching VMs with Hosts when Cloud OpenStack had
Infra OpenStack related to it.

Solves https://bugzilla.redhat.com/show_bug.cgi?id=1528162

screenshot-2018-1-4 manageiq infrastructure providers
screenshot-2018-1-4 manageiq nodes

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
@@ -286,7 +286,7 @@ class ManageIQ::Providers::Openstack::Inventory::Parser::CloudManager < ManageIQ
 
     collector.vms.each do |s|
       if hosts && !s.os_ext_srv_attr_host.blank?
-        parent_host = hosts.find_by('lower(hypervisor_hostname) = ?', s.os_ext_srv_attr_host.downcase)
+        parent_host = hosts.find_by('lower(hypervisor_hostname) = ? OR lower(hypervisor_hostname) = ?', s.os_ext_srv_attr_host.split('.').first.downcase, s.os_ext_srv_attr_host.downcase)
         parent_cluster = parent_host.try(:ems_cluster)
       else
         parent_host = nil


aufi:
The fix looks reasonable to me, waiting till tests - IP/hostname issue is fixed.

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -286,7 +286,7 @@ def vms
 
     collector.vms.each do |s|
       if hosts && !s.os_ext_srv_attr_host.blank?
-        parent_host = hosts.find_by('lower(hypervisor_hostname) = ?', s.os_ext_srv_attr_host.downcase)
+        parent_host = hosts.find_by('lower(hypervisor_hostname) = ?', s.os_ext_srv_attr_host.split('.').first.downcase)
```
aufi: Maybe ```split(/\.[^0-9]/)```? (and IPv6 doesn't have dots)..

blomquisg: Can you use `os_ext_srv_attr_hypervisor_hostname`?  Is that field available on the OSP server model from the API?

https://ask.openstack.org/en/question/50587/os-ext-srv-attrhypervisor_hostname-and-os-ext-srv-attrhost-difference/

blomquisg:
The test that's failing seems to have nothing to do with the changes here, but something in the OpenStack Refresher is broken, so maybe related??

> Failures:
>  1) ManageIQ::Providers::Openstack::CloudManager::Refresher when paired with a infrastructure provider user instance should inherit cpu_speed of compute host
>     Failure/Error: expect(vm.hardware.cpu_speed).to eq(@cpu_speed) if vm.name !='EmsRefreshSpec-Shelved'
>     
>       expected: 2800
>            got: nil
>     
>       (compared using ==)
> ```
>     # ./spec/models/manageiq/providers/openstack/cloud_manager/refresher_rhos_juno_spec.rb:103:in `block (4 levels) in <top (required)>'
>     # ./spec/models/manageiq/providers/openstack/cloud_manager/refresher_rhos_juno_spec.rb:102:in `block (3 levels) in <top (required)>'
> ```

Additionally, I think you should add a test for this change.  The trick is since the host is probably coming from a different provider (undercloud), you might not be able to do this in the basic refresh test.  Maybe mock an undercloud provider with a host that matches one of the VMs in the VCR cassette?

petrblaho:
@blomquisg in fact it is connected - in tests we have IP addresses instead hostnames and fqdns so my change with `split('.').first` causes failing test. During test run these VMs are not matched with Hosts the same way as in BZ this solves and so the VMs do not have matching hardware association and so no cpu_speed.

I will enhance that matching code to probably match both `split('.').first` and full value.
Still investigating possible values there and implications.

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
petrblaho: `os_ext_srv_attr_hypervisor_hostname` should be available in OSP running on top of libvirt/kvm. I am not sure we should generally suppose that this field is available when running OS on other hypervisor (as answer you linked states).

Docs https://developer.openstack.org/api-ref/compute/#show-server-details says that `host` is `The name of the compute host on which this instance is running.` and `hypervisor_hostname` is `The hypervisor host name provided by the Nova virt driver. For the Ironic driver, it is the Ironic node uuid.` That seems like it should be populated by other hypervisors than libvirt/kvm. Anyway in TripleO installed OSP using libvirt/kvm both `os_ext_srv_attr_hypervisor_hostname` and `os_ext_srv_attr_host` have the same value.

Hosts in ManageIQ got their `hypervisor_hostname` attribute from `name` of server in Nova from Undercloud (https://github.com/ManageIQ/manageiq-providers-openstack/blob/fe9a981fffdd20fb6c0fd40dd319b624f782cf12/app/models/manageiq/providers/openstack/infra_manager/refresh_parser.rb#L331) and this `name` is defined by TripleO Heat Templates used to deploy OSP.

So in the end we trust that TripleO (THT) sets the hostnames for compute nodes to the same value (almost - hostname vs fqdn) as their names in Undercloud Nova.

petrblaho:
I updated code with new 'compound' matching for both splitted and full string.

aufi:
Looks good to me, merging!

simaishi:

```
$ git log -1
commit 42bf9cedab8b408cbee1226e73521845606247c8
Date:   Mon Jan 8 09:59:44 2018 +0100

    Merge pull request #186 from petrblaho/add-hostname-domain-split-to-check-in-inventory-parser-cloud-manager-bz1528162
    
    Use only hypervisor hostname to match infra host with cloud vm
    (cherry picked from commit a662f258223fd79cf5de305ee9c00dc802e023c1)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1532347
```

PR #194, opened 2018-01-12
Implement graph refresh for the Cinder manager

mansam:
https://bugzilla.redhat.com/show_bug.cgi?id=1532796


app/models/manageiq/providers/openstack/cloud_manager.rb
```diff
@@ -78,7 +78,7 @@ def ensure_network_manager
 
   def ensure_cinder_manager
     return false if cinder_manager
-    build_cinder_manager(:type => 'ManageIQ::Providers::StorageManager::CinderManager')
+    build_cinder_manager(:type => 'ManageIQ::Providers::Openstack::StorageManager::CinderManager')
```
Ladas: hm I don't think we can easily change the STI subclass of the manager now, since we would need a migrations that will change existing  records. Can we keep the old one? Since we need to backport this.

We can then do upstream only patch changing this, with migration.

app/models/manageiq/providers/openstack/inventory/parser/storage_manager/cinder_manager.rb
```diff
@@ -0,0 +1,95 @@
...
+        _log.warn "#{log_header}: EMS: #{ems.name}, Instance: #{a['server_id']}"
+        next
+      end
+
+      dev = File.basename(a['device'])
+
+      attachment_names = {'vda' => 'Root disk', 'vdb' => 'Ephemeral disk', 'vdc' => 'Swap disk'}
+      persister.disks.find_or_build_by(
+        # FIXME: find works here, but lazy_find doesn't... I don't understand why
+        :hardware    => persister.hardwares.find(a["server_id"]),
```
Ladas: can we do lazy_find ?

mansam: I'll change it and see if the tests agree ;) that comment was written quite a while ago and things have changed since then.

app/models/manageiq/providers/openstack/cloud_manager.rb
mansam: I had a feeling that might be an issue. So to clarify, you'd recommend just leaving this line as `build_cinder_manager(:type => 'ManageIQ::Providers::StorageManager::CinderManager')` and then changing it later along with a migration in another patch?

app/models/manageiq/providers/openstack/storage_manager/cinder_manager/event_target_parser.rb
```diff
@@ -0,0 +1,44 @@
...
+  #
+  # @param ems_event [EmsEvent] EmsEvent object
+  # @return [Array] Array of ManagerRefresh::Target objects
+  def parse_ems_event_targets(ems_event)
+    target_collection = ManagerRefresh::TargetCollection.new(:manager => ems_event.ext_management_system.parent_manager, :event => ems_event)
+
+    # there's almost always a tenant id regardless of event type
+    collect_identity_tenant_references!(target_collection, ems_event)
+
+    # WIP
```
Ladas: right, this will be for standalone block storage based entities, can we also try to extract targets as part of Vm in collector?

e.g. amazon look both in db and API:
https://github.com/Ladas/manageiq-providers-amazon/blob/52636410835d6319627afd5e489cefe199e81a01/app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb#L203

https://github.com/Ladas/manageiq-providers-amazon/blob/52636410835d6319627afd5e489cefe199e81a01/app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb#L249

# And seems like I miss extracting it from the db
---

It's currently the only way how to get full targeted refresh of a Vm

mansam: Yeah absolutely, volumes should be pulled as part of that.

app/models/manageiq/providers/openstack/cloud_manager.rb
Ladas: yes. Since we can't backport the migration now, that would be only for upstream and future releases

app/models/manageiq/providers/openstack/inventory/parser/storage_manager/cinder_manager.rb
Ladas: hum, ok, so if that is the case, we can keep there find for now, since hardwares should be always available (for full and targeted). Otherwise the find would degrade to N+1 queries

Ladas:
@mansam the CI failure is most probably caused by https://github.com/ManageIQ/manageiq/pull/16688/files#r161458751 , once this PR is ready, we can merge, that issue will be fixed separately.

Ladas:
@mansam tomorrow is the last day, then it's non blocker only. Will you be able to finish until then?

mansam:
@Ladas Yeah, I will get this finished by then.

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
```diff
@@ -287,6 +298,11 @@ def infer_related_vm_ems_refs_api!
       add_simple_target!(:cloud_tenants, vm.tenant_id)
       add_simple_target!(:flavors, vm.flavor["id"])
 
+      # pull the attachments from the raw attribute to avoid Fog making an unnecessary call
+      # to inflate the volumes before we need them
+      vm.attributes.fetch('os-extended-volumes:volumes_attached', []).each do |attachment|
+        add_simple_target!(:cloud_volumes, attachment["id"])
```
mansam: @Ladas I'm having trouble getting the targeted collector to pick up the Cloud Volumes. They're added as simple targets here and I can see in my tests that this part works, but the collector method on line 194 never gets called, nor does the parser. I'm not sure what I've done differently that is causing these volume targets to not get parsed, so if you could take a look when you have time that would be much appreciated. I have a feeling I've just been staring at this too long and am simply missing something obvious. :)

app/models/manageiq/providers/openstack/builder.rb
```diff
@@ -4,6 +4,14 @@ def build_inventory(ems, target)
       case target
       when ManageIQ::Providers::Openstack::CloudManager
         cloud_manager_inventory(ems, target)
+      when ManageIQ::Providers::Openstack::StorageManager::CinderManager
+        inventory(
+          ems,
+          target,
+          ManageIQ::Providers::Openstack::Inventory::Collector::StorageManager::CinderManager,
+          ManageIQ::Providers::Openstack::Inventory::Persister::StorageManager::CinderManager,
+          [ManageIQ::Providers::Openstack::Inventory::Parser::StorageManager::CinderManager]
```
Ladas: @mansam for your question below, you need to ```ManageIQ::Providers::Openstack::Inventory::Parser::StorageManager::CinderManager```

to line 30, so ManagerRefresh::TargetCollection (targeted refresh) invokes all parsers

config/settings.yml
```diff
@@ -9,6 +9,11 @@
   :openstack_network:
     :is_admin: false
     :inventory_object_refresh: true
+    :allow_targeted_refresh: true
+  :cinder:
+    :is_admin: false
```
Ladas: just a quick quiestion, the :all_tenants param doesn't work for storage and network?

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
Ladas: @mansam responded https://github.com/ManageIQ/manageiq-providers-openstack/pull/194/files#r161680537

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
```diff
@@ -191,6 +191,14 @@ def vms_by_id
     @vms_by_id ||= Hash[vms.collect { |s| [s.id, s] }]
   end
 
+  def cloud_volumes
```
Ladas: we need the same for cloud_volume_backups and cloud_volume_snapshots, otherwise the targeted refresh would be deleting those (because they would not be found in in the API)

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
```diff
@@ -191,6 +191,14 @@ def vms_by_id
...
+    return [] if references(:cloud_volumes).blank?
+    return @cloud_volumes if @cloud_volumes.any?
+    @cloud_volumes = references(:cloud_volumes).collect do |volume_id|
+      safe_get { cinder_service.volumes.get(volume_id) }
```
Ladas: @mansam just a sanity check, ```cinder_service.volumes.get(volume_id) ``` works across differnet tenants, right?

config/settings.yml
mansam: It does, but they're an option the defaults to being off because at some point someone told me we supported refresh with a non-admin user so I was trying to work around that.

Ladas:
@mansam btw. there is a quick simple spec that the targeted refresh is not deleting something out if its scope https://github.com/Ladas/manageiq-providers-amazon/blob/5372348b633b8c7f37813388013e4e38af402a72/spec/models/manageiq/providers/amazon/cloud_manager/vcr_specs/targeted_refresh_scope_spec.rb#L38

Basically using existing full refresh, then targeted refresh, then full refresh asserts should pass, since the targeted refresh should just update existing data.

Might be for another PR, just to give it more coverage.

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
mansam: It doesn't, sadly. `cinder_service.volumes.all(:id => volume_id, :all_tenants => true)` works, but that requires admin. I might have to just make a scoped request for each volume.

Ladas: right, so probably lets target this as perf issues for next minor release? Could you do some perf test with 200-500 tenants (and more)? That is what we see in production. Then lets talk to Loic where to target this.

Or maybe I understand it badly? How many API queries will this cause with having 500 tenants?

mansam:
Two of the VCR instances have either gone down or stopped responding. I'll update the VCRs for this as soon as I can.

Ladas:
@mansam so you can probably drop those VCR specs :-) I think I had reports that some of the envs were acting strange. Also, we should deploy new version, since most of the versions we test are EOL now. :-)

Ladas:
@mansam what is the state of this? Can we merge soon? I have a bunch of core refactorings I would like to merge after this, so you have an easier backport. :-)

mansam:
@Ladas Should be ready, let me just doublecheck

mansam:
@Ladas This should be good to go now.

app/models/manageiq/providers/openstack/cinder_manager_mixin.rb
```diff
@@ -0,0 +1,20 @@
...
+  include ::CinderManagerMixin
+
+  included do
+    # TODO: how about many storage managers???
+    # Should use has_many :storage_managers,
+    has_one  :cinder_manager,
+             :foreign_key => :parent_ems_id,
+             :class_name  => "ManageIQ::Providers::StorageManager::CinderManager",
+             :autosave    => true,
+             :dependent   => :destroy
```
Ladas: hm, I think we were removing the :dependent => :destroy lately https://github.com/ManageIQ/manageiq-providers-openstack/pull/208/files

We might want to drop it?

Ladas: Hm, though this passes the CI, so could you maybe do that as separate PR? We might need to backport both then.

aufi:
Looks good to me and trust @Ladas too, merging!

simaishi:
@mansam This PR as well as https://github.com/ManageIQ/manageiq-providers-openstack/pull/242 conflicts on VCRs as well...

I know you already have a Gaprindashvili PR for https://github.com/ManageIQ/manageiq-providers-openstack/pull/265.  Not sure what needs to be done for this and PR 242...

mansam:
@simaishi Unfortunately this always happens with PRs involving VCR updates. I'll create some backport PRs for Gaprindashvili and just have to update the VCRs on them as each one gets merged in turn.

simaishi:
Backported to Gaprindashvili via https://github.com/ManageIQ/manageiq-providers-openstack/pull/274

PR #196, opened 2018-01-16
Added keystone_v3_domain_id to api_allowed_attributes method

snecklifter:
This is a follow on PR from ManageIQ/manageiq#16802/
and ManageIQ/manageiq-api#279


juliancheal:
👍 LGTM

blomquisg:
@mansam, probably need you or @aufi to look over this test failure.  It looks legit, but probably not related to this patch.

mansam:
@blomquisg the test failure seems like it's fixed by https://github.com/ManageIQ/manageiq-providers-openstack/pull/198, which I assume is a fix for something that changed in core. Should be unrelated to this patch.

PR #198, opened 2018-01-17
[WIP] Provider base class handles the managers' destroy now

jameswnl:


mansam:
@jameswnl is there more to be done on this since it's marked WIP? I'm assuming the reason the rest of the PRs have failing tests is because something on core changed and this hasn't been merged yet.

jameswnl:
@mansam this should be the only change needed, to sync up with https://github.com/ManageIQ/manageiq/pull/16755 which makes the `Provider.destroy` handling destroy of contained managers.

I've checked with UI team when user is destroying an Openstack provider, UI invokes the `ManageIQ::Providers::CloudManager.destroy_queue`, not the provider object.  Do you know when will this `provider` record got removed?

mansam:
@jameswnl I'm sorry, I don't think I quite understand the question. `destroy_queue` puts a message on the queue to asynchronously call `destroy` on the provider. That message should be received and acted on quickly.

mansam:
@jameswnl If this is ready to be merged, could you remove the [WIP]?

jameswnl:
@miq-bot remove_labels wip

PR #206, opened 2018-01-23
Ensure update of more_core_extensions gem

aufi:
Hostname validation was added to newest version, updating gemspec to include it.


aufi:
Looks to be related to https://github.com/ManageIQ/manageiq/pull/16714

PR #207, opened 2018-01-23
[WIP] Fix destroy provider

aufi:
Change introduced in ManageIQ/manageiq#16755 is not
compatible with OSP provider which uses cascade deletion of related providers.

Nullification introduced in #198
breaks tests and possibly affects Infra provider deletion.

Trying override destroy method with original from AR until we get better solution (preferably in main repo).


aufi:
@Ladas First attempt..

aufi:
@agrare Ok, thanks for review. Closing in favor of https://github.com/ManageIQ/manageiq-providers-openstack/pull/208

PR #209, opened 2018-01-24
Bypass the superclass orchestrated destroy for this Provider.

Fryguy:
In the OpenStack provider, the Provider instance is only tightly coupled
to the InfraManager.  That is, when the InfraManager is created, then
the Provider is created and vice-versa, when the InfraManager is
destroyed, we need to destroy the Provider.  The CloudManager objects
associated to that Provider should *not* be destroyed, and only need to
be nullified.  For other Providers, the Provider instance is what is
destroyed, so by default the orchestrated destroy destroys all of its
managers, but we here don't want that to happen, so we are bypassing.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1538232

@Ladas @aufi Please review
cc @agrare @jameswnl

### DIFF
--- a/app/models/manageiq/providers/openstack/provider.rb
+++ b/app/models/manageiq/providers/openstack/provider.rb
@@ -6,11 +6,26 @@ class ManageIQ::Providers::Openstack::Provider < ::Provider
   has_many :cloud_ems,
            :foreign_key => "provider_id",
            :class_name  => "ManageIQ::Providers::Openstack::CloudManager",
+           :dependent   => :nullify,
            :autosave    => true
   has_many :network_managers,
            :foreign_key => "provider_id",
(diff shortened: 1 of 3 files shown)


simaishi:

```
$ git log -1
commit 53194e3ee7ef6f7fed17c09eba3def77661a82ac
Date:   Thu Jan 25 10:39:52 2018 +0100

    Merge pull request #209 from Fryguy/fix_provider_destroy
    
    Bypass the superclass orchestrated destroy for this Provider.
    (cherry picked from commit 158b93f4315516a7d05b3baa1b75e5c90bc26285)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1538240
```

PR #210, opened 2018-01-25
Store selected user sync roles as custom attributes.

rwsu:
Selected roles are saved as provider custom attributes.

If roles are changed between different syncs, user memberships from
previously selected roles should be removed.


aufi:
Looks good to me, merging.

PR #212, opened 2018-01-29
Filter out duplicates during inventory collection

mansam:
Initial fix for https://bugzilla.redhat.com/show_bug.cgi?id=1528663

There are some disadvantages that make it difficult to fix the bug by changing pagination in openstack_handle::pagination::marker.rb, since pagination can be different for each service and collection, Fog can make it difficult to get the raw response, and some Fog collections behave differently from others. To avoid these problems, this fix implements a filter for uniqueness at the point where the inventory is collected.

Long term the correct fix is to totally reimplement the way that paging is done in OpenstackHandle/Fog.


app/models/manageiq/providers/openstack/inventory/collector/cloud_manager.rb
```diff
@@ -15,7 +15,7 @@ def availability_zones
 
   def cloud_services
     return @cloud_services if @cloud_services.any?
-    @cloud_services = compute_service.handled_list(:services, {}, openstack_admin?)
+    @cloud_services = uniques(compute_service.handled_list(:services, {}, openstack_admin?))
```
Ladas: we can ignore graph refresh, its indexes will not allow duplicate records to be passed to persistor

Ladas:
@mansam we will need to rerecord the VCRs, conflict with the storage targeted refresh?

mansam:
@Ladas Probably. They're currently rerecording.

aufi:
Looks good to me, merging!

mansam:
@Ladas Yeah, specs to reproduce this are on my log of things to do.

mansam:
@miq-bot add_label fine/yes

simaishi:
@mansam For backporting to Gaprindashvili branch, all VCR files are conflicting on cherry-pick. If I should take all files from this PR ignoring the conflict, I can handle that. Otherwise, please create a PR for Gaprindashvili.

For backporting to Fine branch, there are only 3 files that are clean cherry-pick. Please create a PR for Fine branch.

simaishi:
Backported to Gaprindashvili via https://github.com/ManageIQ/manageiq-providers-openstack/pull/219

simaishi:
Backported to Fine via https://github.com/ManageIQ/manageiq/pull/16934

PR #218, opened 2018-02-01
Don't require CinderManager in inventory classes

mansam:
This require shouldn't be necessary and causes an exception.


aufi:
LGTM

PR #222, opened 2018-02-02
Add proper error message if network type not supported

alexander-demichev:
When creating network, add proper error message if network type is not supported by provider.
Fixes [link](https://bugzilla.redhat.com/show_bug.cgi?id=1540692)
screen shot 2018-02-02 at 13 22 10

### DIFF
--- a/app/models/manageiq/providers/openstack/network_manager/cloud_network.rb
+++ b/app/models/manageiq/providers/openstack/network_manager/cloud_network.rb
@@ -50,7 +50,13 @@ class ManageIQ::Providers::Openstack::NetworkManager::CloudNetwork < ::CloudNetw
     {:ems_ref => network.id, :name => options[:name]}
   rescue => e
     _log.error "network=[#{options[:name]}], error: #{e}"
-    raise MiqException::MiqNetworkCreateError, parse_error_message_from_neutron_response(e), e.backtrace
+    parsed_error = parse_error_message_from_neutron_response(e)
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/openstack/network_manager/cloud_network.rb
```diff
@@ -50,7 +50,13 @@ def self.raw_create_cloud_network(ext_management_system, options)
     {:ems_ref => network.id, :name => options[:name]}
   rescue => e
     _log.error "network=[#{options[:name]}], error: #{e}"
-    raise MiqException::MiqNetworkCreateError, parse_error_message_from_neutron_response(e), e.backtrace
+    parsed_error = parse_error_message_from_neutron_response(e)
+    if parsed_error =~ /Invalid input for operation: network_type value '.*' not supported\./
+      error_message = _("Network type '#{options[:provider_network_type]}' is not supported by the Provider")
+      raise MiqException::MiqNetworkCreateError, error_message, e.backtrace
```
mansam: Looks good! But you could do this without two "raises" by assigning back to `parsed_error` :)

alexander-demichev: Thanks for review, does it look better now?

gildub:
LGTM. Just the Rubocop warning - I wasn't used to assign the value that way but it makes sense.

PR #232, opened 2018-02-16
Collect all tenants during vm targeted refresh

agrare:
To prevent fog-openstack building the wrong query for
.tenants.find_by_id which leads to a 404 error query all tenants during
a targeted refresh.


agrare:
@Ladas Looks like the VCRs will need to be re-recorded.

Alternatively we can only call `.visible_tenants` if `identity_service` doesn't respond to `projects`.

aufi:
The change looks good to me, thanks for the PR :+1: 

I uploaded VCR recording for this change as https://github.com/ManageIQ/manageiq-providers-openstack/pull/233

agrare:
Closing in favor of https://github.com/ManageIQ/manageiq-providers-openstack/pull/233

PR #233, opened 2018-02-19
Bz 1546109 fix targeted refresh clearing vm cloud tenant for v2

aufi:
To prevent fog-openstack building the wrong query for
.tenants.find_by_id which leads to a 404 error query all tenants during
a targeted refresh.

Adam's PR with VCR reloaded, so tests are passing. https://github.com/ManageIQ/manageiq-providers-openstack/pull/232


simaishi:
VCRs are conflicting during cherry-pick to gaprindashvili branch. Should I take all VCRs from this PR, ignoring what's there now?

agrare:
This still fails with the VCRs from master, looks like will have to be re-recorded to gaprindashvili

aufi:
I'm going to rerecord VCRs with G-branch.

aufi:
@agrare VCR rerecord fails for G-branch with and even without the change introduced in this PR, something seems to be broken - http://paste.openstack.org/show/NeM4FFYUMcQj1Yif4DnX/

agrare:
cc @Ladas ^^

Ladas:
@aufi hm, that code is not in g-branch, is it possible you were trying this on master code by mistake?

simaishi:
Backported to Gaprindashvili via https://github.com/ManageIQ/manageiq-providers-openstack/pull/235

PR #236, opened 2018-02-26
Add delete_queue method for Template

andyvesel:
Added missing method `delete_image_queue` for templates in order to use it in API `delete` action
https://github.com/ManageIQ/manageiq-api/pull/328


mansam:
Looks right, can confirm that locally the tests pass on the related manageiq-api PR when this is merged.

PR #243, opened 2018-03-01
Don't lose VM volume attachments when refreshing the cloud inventory

mansam:
When the CloudManager refreshes, it removes from the inventory any Disks that were only added by the CinderManager as part of attaching Volumes to VMs since the CloudManager doesn't find them while collecting inventory. As a result, Volumes will have correct statuses but incorrect numbers of attachments, leading to confusing user experience and various UI actions failing.
This patch causes the CloudManager to be aware of VM volume attachments so that the associations are not broken between Cloud and Cinder refreshes.

This is a solution for https://bugzilla.redhat.com/show_bug.cgi?id=1544344

Depends on https://github.com/ManageIQ/manageiq-providers-openstack/pull/242


aufi:
Looks good to me, merging!

PR #247, opened 2018-03-07
Refactor node management into supports features

rwsu:
Add additional checks to verify workflow or baremetal service
is available in order for the feature to be supported.

Also check workflow that will be executed is also available
in order for the feature to be supported.

Depends on https://github.com/ManageIQ/manageiq/pull/17086

Partially fixes https://bugzilla.redhat.com/show_bug.cgi?id=1542974
Requires additional UI update.


mansam:
Looks great! Just need to let the supports feature mixin know about the new capability so the tests go green. :)

PR #253, opened 2018-03-15
Parse volume attachment/detachment messages from fog responses

mansam:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1530345

image

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume/operations.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume/operations.rb
@@ -19,7 +19,7 @@ module ManageIQ::Providers::Openstack::CloudManager::CloudVolume::Operations
       end
     end
   rescue => e
-    volume_name = name.blank? ? ems_ref : name
+    volume_name = name.presence || ems_ref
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/openstack/cloud_manager/cloud_volume/operations.rb
```diff
@@ -18,6 +18,10 @@ def raw_attach_volume(server_ems_ref, device = nil)
         service.servers.get(server_ems_ref).attach_volume(ems_ref, device)
       end
     end
+  rescue => e
+    volume_name = name.empty? ? ems_ref : name
```
aufi: Wouldn't be better use ```name.blank?``` to not fail when _name_ would be _nil_ or nil cannot appear there?

mansam: I think you're right that `blank?` is the better choice here.

mansam: Ended up using "presence" because rubocop complained :)

aufi:

PR #265, opened 2018-04-03
Avoid tenant discovery recursion

mansam:
https://bugzilla.redhat.com/show_bug.cgi?id=1561637

### DIFF
--- a/lib/manageiq/providers/openstack/legacy/openstack_handle/handle.rb
+++ b/lib/manageiq/providers/openstack/legacy/openstack_handle/handle.rb
@@ -166,17 +166,19 @@ module OpenstackHandle
       opts     = options.dup
       service  = (opts.delete(:service) || "Compute").to_s.camelize
       tenant   = opts.delete(:tenant_name)
+      discover_tenants = opts.fetch(:discover_tenants, true)
+      opts.delete(:discover_tenants)
       domain   = domain_id
 
       # Do not send auth_type to fog, it throws warning
       opts.delete(:auth_type)
 
-      unless tenant
-        tenant ||= default_tenant_name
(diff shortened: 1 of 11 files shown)


lib/manageiq/providers/openstack/legacy/openstack_handle/handle.rb
```diff
@@ -380,7 +382,7 @@ def accessible_tenant_names
     def default_tenant_name
       return @default_tenant_name ||= "admin" if tenant_accessible?("admin")
       tenant_names.each do |name|
-        next if name == "services"
+        next if name.blank? || name == "services"
```
aufi: Is there an example for case when name.blannk? == true? I've left it in example code from experimenting with tenant with empty name, but I haven't found such case in OSP.

mansam: I didn't think so, but I thought perhaps you had a good reason for it. I'll pull the change if you agree it's unlikely to come up.

aufi: Yep, I'd suggest drop it in that case.

aufi:
Looks good to me :+1:

simaishi:
@mansam Most of the VCR cassettes are conflicting during backport to gaprindashvili. Should I just take the VCRs from this PR ignoring the conflicts, or do you need to re-generate for the G branch?

mansam:
@simaishi Unfortunately I have to regenerate them for the G branch. I'll create a new PR and then ping you.

simaishi:
Backported to Gaprindashvili via https://github.com/ManageIQ/manageiq-providers-openstack/pull/268

PR #272, opened 2018-04-11
Use the correct id when collecting quotas from Neutron

mansam:
Quotas from Neutron don't seem to contain an 'id' field like from Nova, so the current collection fails to gather the networking quotas correctly.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1390456


aufi:
Seems to be correct, tenant_id is present in API at least from Newton.
LGTM :+1:

PR #284, opened 2018-05-02
bz1573893: duplicate opts hash before modifying in raw_connect_try_ssl

sseago:
https://bugzilla.redhat.com/show_bug.cgi?id=1573893

raw_connect_try_ssl modifies the opts hash (pulling out :ssl_options and then
using it. This means that if raw_connect_try_ssl is called twice with the
same opts hash, the second time fails because of the side effect of the first
call. With this change, the method works on a copy of the hash rather than
the original

### DIFF
--- a/lib/manageiq/providers/openstack/legacy/openstack_handle/handle.rb
+++ b/lib/manageiq/providers/openstack/legacy/openstack_handle/handle.rb
@@ -46,8 +46,9 @@ module OpenstackHandle
       end
     end
 
-    def self.raw_connect_try_ssl(username, password, address, port, service = "Compute", opts = nil, api_version = nil,
+    def self.raw_connect_try_ssl(username, password, address, port, service = "Compute", options = nil, api_version = nil,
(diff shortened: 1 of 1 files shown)


aufi:
Thanks for the PR Scott! Looks good to me, the codeclimate complains about amount of arguments of a method which was there before, so might be fixed later, merging.

simaishi:
Fine backport (to manageiq-gems-pending repo) details:

```
$ git log -1
commit 5cbe2c76e99b4f91b068e545b43eb20740699cd0
Date:   Thu May 3 10:04:38 2018 +0200

    Merge pull request #284 from sseago/bz1573893
    
    bz1573893: duplicate opts hash before modifying in raw_connect_try_ssl
    (cherry picked from commit 70f8b6d6be73108bd9dc62cfc4cbef3bbe9b3739)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1574571
```

PR #286, opened 2018-05-08
Combine InfraManager and child manager refresh queues

mansam:
When the CloudManager's refresh queues were consolidated,
the individual NetworkManager refresh worker was removed.
However, the InfraManager's queues were not combined at
the same time, resulting in Infra network refresh events
not being picked up. This commit ensures that the Infra
Manager's refresh worker picks up queued refreshes for
its child managers as well.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1572760


aufi:
Looks good to me, merging!

PR #289, opened 2018-05-11
Default volume size value in provisioning

alexander-demichev:
This PR add default volume size value for instance provisioning in case if only name is specified. A placeholder with warning will be placed in UI field.
https://bugzilla.redhat.com/show_bug.cgi?id=1417215


aufi:
Backend part itself is enough to not fail OpenStack API call, UI part needs to be discussed further.
Looks good to me, merging!

PR #297, opened 2018-05-25
Bug 1518304: Don't show events tab for openstack network provider edit

sseago:
Add supports entry for events configuration for both infra and cloud providers.

After https://github.com/ManageIQ/manageiq-ui-classic/pull/3832 this is needed to show the event tab for the cloud and infra provider edit page.


sseago:
Closing PR, since the original bug fix no longer needs to make provider-level changes.

PR #300, opened 2018-06-07
Add notifications for Router delete

andyvesel:
Added notifications for deleting Network Router. 
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1518734
Depends on https://github.com/ManageIQ/manageiq/pull/17514


mansam:
Looks good, but same question as on https://github.com/ManageIQ/manageiq-providers-openstack/pull/301, because I don't know :)

PR #301, opened 2018-06-11
Add notifications for Subnets, Networks

andyvesel:
Added notifications for deleting Cloud Network and Cloud Subnet
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1470210
Depends on https://github.com/ManageIQ/manageiq/pull/17556

### DIFF
--- a/app/models/manageiq/providers/openstack/network_manager/cloud_network.rb
+++ b/app/models/manageiq/providers/openstack/network_manager/cloud_network.rb
@@ -60,8 +60,13 @@ class ManageIQ::Providers::Openstack::NetworkManager::CloudNetwork < ::CloudNetw
   end
 
   def raw_delete_cloud_network
-    ext_management_system.with_provider_connection(connection_options(cloud_tenant)) do |service|
-      service.delete_network(ems_ref)
+    with_notification(:cloud_network_delete,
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/openstack/network_manager/cloud_network.rb
```diff
@@ -60,8 +60,13 @@ def self.raw_create_cloud_network(ext_management_system, options)
   end
 
   def raw_delete_cloud_network
-    ext_management_system.with_provider_connection(connection_options(cloud_tenant)) do |service|
-      service.delete_network(ems_ref)
+    with_notification(:cloud_network_delete,
+                      :options => {
+                        :subject => self,
```
mansam: When an object is passed as `:subject` like this, does it get translated nicely into a string for the notification message, or does it show up as an object reference?

mansam:
Looks good, just curious about that one thing I commented on.

app/models/manageiq/providers/openstack/network_manager/cloud_network.rb
mansam: This is how the rest of these notifications are written, so I assume it's fine

PR #302, opened 2018-06-13
Add display name for flavor

mzazrivec:
This is pluggable version of https://github.com/ManageIQ/manageiq/pull/16944


mansam:
LGTM

PR #303, opened 2018-06-14
Change expected partial refresh error expection

agrare:
Needed for https://github.com/ManageIQ/manageiq/pull/17472


agrare:
Thanks @mansam

PR #305, opened 2018-06-19
(Depends on ManageIQ/manageiq#17610) Collect Cinder Volume Types during Inventory Refresh

mansam:
https://bugzilla.redhat.com/show_bug.cgi?id=1592900

This PR collects Volume Types from Cinder as part of a full refresh so that they can be used when creating new volumes. This is especially important for Openstack V2V infrastructure mapping.

Depends on https://github.com/ManageIQ/manageiq-schema/pull/223 and https://github.com/ManageIQ/manageiq/pull/17610


aufi:
Dependencies were merged, LGTM, merging :+1:

PR #309, opened 2018-06-26
Add template decorator

alexander-demichev:
Add a UI decorator for template. Method ```provisioning_volume_size_tooltip ``` returns message that will be displayed in instance provisioning.

### DIFF
--- a/app/decorators/manageiq/providers/openstack/cloud_manager/template_decorator.rb
+++ b/app/decorators/manageiq/providers/openstack/cloud_manager/template_decorator.rb
@@ -1,0 +1,5 @@
+class ManageIQ::Providers::Openstack::CloudManager::TemplateDecorator < MiqTemplateDecorator
+  def provisioning_volume_size_tooltip
+    _("Default value is 1")
+  end
+end


alexander-demichev:
@miq_bot add_reviewer @skateman

alexander-demichev:
@miq-bot add_reviewer @skateman

app/decorators/manageiq/providers/openstack/cloud_manager/template_decorator.rb
```diff
@@ -0,0 +1,9 @@
+class ManageIQ::Providers::Openstack::CloudManager::TemplateDecorator < MiqTemplateDecorator
+  def single_quad
```
skateman: You don't need this line since you're inheriting the method from `MiqTemplateDecorator`.

app/decorators/manageiq/providers/openstack/cloud_manager/template_decorator.rb
```diff
@@ -0,0 +1,9 @@
...
+    {}
+  end
+
+  def size_tooltip
```
skateman: I think you should give it a little more meaningful name, like `provisioning_volume_size_tooltip` but you might come up with something better.

alexander-demichev:
https://github.com/ManageIQ/manageiq-ui-classic/pull/3928

aufi:
@alexander-demichev Could you paste a link to the decorator's PR which should be merged before this PR?

skateman:
@aufi https://github.com/ManageIQ/manageiq-ui-classic/pull/4209

PR #310, opened 2018-06-26
Fix tenant associations on VolumeSnapshotTemplates

mansam:
Found and fixed two issues: one was that if a volume snapshot was available but the volume it was based on wasn't available at the time of refresh, that volume wouldn't be collected. The other issue is that the inventory parser was trying to read the wrong field to collect the tenant, and as such wasn't finding one. This was causing VolumeSnapshotTemplates to appear visible to all tenants.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1589028


aufi:
Thanks @mansam for the fix! Looks good to me, merging :+1:

simaishi:
@mansam Most of the vcr cassettes conflict backporting to Gaprindashvili branch. Please create a separate PR.

mansam:
@simaishi will do

simaishi:
Backported to Gaprindashvili via https://github.com/ManageIQ/manageiq-providers-openstack/pull/312

PR #311, opened 2018-06-27
Added allow_delete method on AuthKeyPairs

juliancheal:
Fixes BZ https://bugzilla.redhat.com/show_bug.cgi?id=1418481

When a local user auth key is created that can be deleted, but not
ones imported from the provider.

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/auth_key_pair.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/auth_key_pair.rb
@@ -32,6 +32,14 @@ class ManageIQ::Providers::Openstack::CloudManager::AuthKeyPair < ManageIQ::Prov
   end
 
   def validate_delete_key_pair
-    {:available => true, :message => nil}
+    {:available => allow_delete?, :message => nil}
+  end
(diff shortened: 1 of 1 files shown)


juliancheal:
TODO: need to return error if a key is tried to be deleted that shouldn't. AKA from the key list view, in UI (or api too?)

mansam:
@juliancheal I'm a bit puzzled by this. The `auth_key` field is (to my knowledge) only present when a key pair has been created through ManageIQ; when collecting preexisting keys from Openstack during refresh, the private key is unavailable so `auth_key` will be blank. `allow_delete?` however returns true only when the `auth_key` isn't present. This would mean that keys created in ManageIQ would not be deletable since they have an `auth_key` present, whereas other preexisting keys collected from the inventory would be. That doesn't seem like the intended behavior, or am I misunderstanding what you mean by "imported"?

juliancheal:
Amazon PR https://github.com/ManageIQ/manageiq-providers-amazon/pull/457 and UI PR https://github.com/ManageIQ/manageiq-ui-classic/pull/4222

PR #318, opened 2018-07-16
Parse all keypairs with assigned userids

andyvesel:
This parses keypairs with userids over users

Depends on 
https://github.com/ManageIQ/manageiq/pull/17693
https://github.com/ManageIQ/manageiq/pull/17714
https://github.com/fog/fog-openstack/pull/406

Related BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1589766

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/collector/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/inventory/collector/cloud_manager.rb
@@ -65,7 +65,18 @@ class ManageIQ::Providers::Openstack::Inventory::Collector::CloudManager < Manag
 
   def key_pairs
     return @key_pairs if @key_pairs.any?
-    @key_pairs = compute_service.handled_list(:key_pairs, {}, openstack_admin?)
+    users = User.all
+
+    results = users.map do |u|
+      next if u.settings[:openstack_user_id].nil?
+      @key_pairs = compute_service.handled_list(:key_pairs, {:user_id => u.settings[:openstack_user_id]}, openstack_admin?)
+
+      @key_pairs.map do |kp|
+        kp.user_id = u.settings[:openstack_user_id]
+        kp
+      end
+    end.flatten
+    results
   end
 
   def quotas


app/models/manageiq/providers/openstack/inventory/collector/cloud_manager.rb
```diff
@@ -65,7 +65,17 @@ def images
 
   def key_pairs
     return @key_pairs if @key_pairs.any?
-    @key_pairs = compute_service.handled_list(:key_pairs, {}, openstack_admin?)
+    users = User.all
```
aufi: @andyvesel We might need filter only Users which came from TenantMapping (e.g. with _openstack_user_id_ to not kill refresh in case of many users in CF and let's hope that CU will not have hundreds OpenStack users mapped into CF).

andyvesel: added `next if` condition to skip users with `settings.[:openstack_user_id]nil? `

aufi: Maybe also include admin user

JPrause:
@andyvesel if this can be backported, please add the gaprindashvili/yes label

app/models/manageiq/providers/openstack/inventory/collector/cloud_manager.rb
```diff
@@ -65,7 +65,18 @@ def images
...
+
+    results = users.map do |u|
+      next if u.settings[:openstack_user_id].nil?
+      @key_pairs = compute_service.handled_list(:key_pairs, {:user_id => u.settings[:openstack_user_id]}, openstack_admin?)
+
+      @key_pairs.map do |kp|
+        kp.user_id = u.settings[:openstack_user_id]
```
agrare, line 75: @andyvesel what does `u.settings[:openstack_user_id]` look like?  It probably isn't equal to our `user.id` in our database right?

agrare (commented):
Sorry to resurrect an old PR but we're looking at implementing this again

PR #323, opened 2018-07-24
Targeted refresh workaround for events that don't identify objects

mansam:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1596058 in combination with https://github.com/ManageIQ/manageiq-content/pull/371 and https://github.com/ManageIQ/manageiq-content/pull/372

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
@@ -70,10 +70,12 @@ class ManageIQ::Providers::Openstack::Inventory::Collector::TargetCollection < M
   end
 
   def floating_ips
-    return [] if references(:floating_ips).blank?
+    return [] if references(:floating_ips).blank? && references(:floating_ips_by_address).blank?
     return @floating_ips if @floating_ips.any?
-    @floating_ips = references(:floating_ips).collect do |floating_ip|
+    @floating_ips = references(:floating_ips_by_address).collect do |floating_ip|
       safe_get { network_service.floating_ips.all(:floating_ip_address => floating_ip).first }
+    end.compact + references(:floating_ips).collect do |floating_ip_id|
+      safe_get { network_service.floating_ips.get(floating_ip_id) }
     end.compact
   end
 
@@ -133,11 +135,11 @@ class ManageIQ::Providers::Openstack::Inventory::Collector::TargetCollection < M
   end
 
   def key_pairs
+    # keypair notifications from panko don't include ids, so
+    # we will just refresh all the keypairs if we get an event.
     return [] if references(:key_pairs).blank?
     return @key_pairs if @key_pairs.any?
-    @key_pairs = references(:key_pairs).collect do |key_pair_id|
-      safe_get { compute_service.key_pairs.get(key_pair_id) }
-    end.compact
+    @key_pairs = compute_service.handled_list(:key_pairs, {}, openstack_admin?)
   end
 
   def flavors_by_id
@@ -211,11 +213,11 @@ class ManageIQ::Providers::Openstack::Inventory::Collector::TargetCollection < M
   end
 
   def cloud_volume_backups
+    # backup notifications from panko don't include ids, so
+    # we will just refresh all the backups if we get an event.
     return [] if references(:cloud_volume_backups).blank?
     return @cloud_volume_backups if @cloud_volume_backups.any?
-    @cloud_volume_backups = targets_by_association(:cloud_volume_backups).collect do |target|
-      scoped_get_backup(target.manager_ref[:ems_ref], target.options[:tenant_id])
-    end.compact
+    @cloud_volume_backups = cinder_service.handled_list(:list_backups_detailed, {:__request_body_index => "backups"}, cinder_admin?)
   end
 
   def scoped_get_volume(volume_id, tenant_id)
@@ -309,7 +311,7 @@ class ManageIQ::Providers::Openstack::Inventory::Collector::TargetCollection < M
 
       all_stacks.each { |s| add_simple_target!(:orchestration_stacks, s.ems_ref, :tenant_id => s.cloud_tenant.ems_ref) }
       vm.cloud_networks.collect(&:ems_ref).compact.each { |ems_ref| add_simple_target!(:cloud_networks, ems_ref) }
-      vm.floating_ips.collect(&:address).compact.each { |address| add_simple_target!(:floating_ips, address) }
+      vm.floating_ips.collect(&:ems_ref).compact.each { |ems_ref| add_simple_target!(:floating_ips, ems_ref) }
       vm.network_ports.collect(&:ems_ref).compact.each do |ems_ref|
         add_simple_target!(:network_ports, ems_ref)
       end
@@ -344,7 +346,7 @@ class ManageIQ::Providers::Openstack::Inventory::Collector::TargetCollection < M
       vm.security_groups.each do |sg|
         add_simple_target!(:security_groups, sg.id)
       end
-      add_simple_target!(:floating_ips, vm.public_ip_address) unless vm.public_ip_address.blank?
+      add_simple_target!(:floating_ips_by_address, vm.public_ip_address) if vm.public_ip_address.present?
     end
     target.manager_refs_by_association_reset
     floating_ips.each do |floating_ip|
--- a/app/models/manageiq/providers/openstack/inventory/persister/definitions/cloud_collections.rb
+++ b/app/models/manageiq/providers/openstack/inventory/persister/definitions/cloud_collections.rb
@@ -124,8 +124,14 @@ module ManageIQ::Providers::Openstack::Inventory::Persister::Definitions::CloudC
       builder.add_properties(
         :model_class => ManageIQ::Providers::Openstack::CloudManager::AuthKeyPair,
       )
-
-      builder.add_default_values(:resource => manager) unless targeted?
+      # targeted refresh workaround-- always refresh the whole keypair collection
+      # regardless of whether this is a TargetCollection or not
+      # because OpenStack doesn't give us UUIDs of changed keypairs,
+      # we just get an event that one of them changed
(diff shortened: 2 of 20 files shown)


app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
```diff
@@ -309,7 +311,7 @@ def infer_related_vm_ems_refs_db!
 
       all_stacks.each { |s| add_simple_target!(:orchestration_stacks, s.ems_ref, :tenant_id => s.cloud_tenant.ems_ref) }
       vm.cloud_networks.collect(&:ems_ref).compact.each { |ems_ref| add_simple_target!(:cloud_networks, ems_ref) }
-      vm.floating_ips.collect(&:address).compact.each { |address| add_simple_target!(:floating_ips, address) }
+      vm.floating_ips.collect(&:ems_ref).compact.each { |address| add_simple_target!(:floating_ips, address) }
```
Ladas: probably rename also the other address to ems_ref

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
```diff
@@ -211,11 +213,11 @@ def cloud_volume_snapshots
...
 
   def cloud_volume_backups
+    # backup notifications from panko don't include ids, so
+    # we will just refresh all the backups if we get an event.
     return [] if references(:cloud_volume_backups).blank?
     return @cloud_volume_backups if @cloud_volume_backups.any?
-    @cloud_volume_backups = targets_by_association(:cloud_volume_backups).collect do |target|
-      scoped_get_backup(target.manager_ref[:ems_ref], target.options[:tenant_id])
-    end.compact
+    @cloud_volume_backups = cinder_service.handled_list(:list_backups_detailed, {:__request_body_index => "backups"}, cinder_admin?)
```
Ladas: we will need specs for this, I think this way it would lead to duplication of existing cloud_volume_backups

I think what we could do to make this work on persistor could be switching ```:targeted => true/false```. Given references will be [nil] or [], then ```targeted => true``` will not update or delete anything, since the targeted query will not return any records. With ```:targeted => false```, we will do a full refresh of the cloud_volume_backups IC.

Same for the key_pairs. But lets do specs to verify this will work.

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
mansam: :+1:

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
mansam: I'll write specs to make sure this behaves correctly. Why do you suspect it would result in duplicated objects?

Ladas: so the targeted refresh takes the references and it builds a query, so it would be:

```CloudVolumeBackups.where(:ems_ref => references(...))```

So if this returns always [], the refresh will create anything built by the parser. In this case that would be all cloud_volume_backups are created again. While the correct behavior should be that the query returns the existing ones, and we just update them.

Comment (mansam) [reply app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb]: Ah, gotcha

Comment (mansam) [reply app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb]: Local testing as well as my new specs suggest that this duplication does not occur.

Comment (Ladas) [reply app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb]: ah, I see, the DB scope is automatically inferred from the built data :-)

but there are still 2 things left, if we don't switch off the ```:targeted``` mode:
1. The scope is just for built data, so we will not delete entities that do not exist in the provider (you can try the stubbed spec with decreasing scaling)
2. The query for full refresh is more effective than the targeted one, if we are fetching all records

Comment (mansam) [reply app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb]: Ah, I see. `:targeted` seems to be at the persister level, not just the individual object collection, so how can I isolate one collection as not targeted while the rest are?

Comment (Ladas) [reply app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb]: Hm, you should be able to overwrite it per 1 Inventory Collection?

Comment (mansam) [reply app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb]: It seems like if I set the key_pair collection to `:targeted => false` that it still calls TargetedCollection's key pair collection method rather than CloudManager's, and entities aren't being deleted. Here's the persister definition:

```ruby
  def add_key_pairs(extra_properties = {})
    add_collection(cloud, :key_pairs, extra_properties, :shared_properties => shared_options.merge(:targeted => false)) do |builder|
      builder.add_properties(
        :model_class => ManageIQ::Providers::Openstack::CloudManager::AuthKeyPair
      )

      builder.add_default_values(:resource => manager) unless targeted?
    end
  end
```

Here's the test that fails because objects aren't removed 
```ruby
 it "works around keypair targeted refresh by refreshing all keypairs without creating duplicates" do
  @data_scaling = 1
  setup_mocked_collector
  2.times do
    keypair_target = ManagerRefresh::Target.new(:manager     => @ems,
                                                :association => :key_pairs,
                                                :manager_ref => {:ems_ref => nil})
    EmsRefresh.refresh(keypair_target)
    expect(AuthPrivateKey.count).to eq(test_counts(@data_scaling)[:key_pairs_count])
    @data_scaling -= 1
  end
end
```
I can see that the Inventory Collection's `@targeted` property is returning `false`, so I don't think there's an issue with it being passed incorrectly. Any ideas?

Ladas: try

```ruby

 def add_key_pairs(extra_properties = {})
    add_collection(cloud, :key_pairs, extra_properties) do |builder|
      builder.add_properties(
        :model_class => ManageIQ::Providers::Openstack::CloudManager::AuthKeyPair
      )

      builder.add_properties(
        :targeted => false
      ) unless references(:key_pairs).blank?

      builder.add_default_values(:resource => manager) unless targeted?
    end
  end

```

Ladas: btw the ```builder.add_default_values(:resource => manager) unless targeted?``` looks wrong? The :resource should be filled in any refresh?

Ladas: the ```that it still calls TargetedCollection's key pair collection method rather than CloudManager's```, this should be ok right? Since you have there ```unless references(:key_pairs).blank? -> collect all keypairs```

mansam: `builder.add_properties(:targeted => false)` doesn't make a difference. I'm not sure why `resource` is being set conditionally, I think that happened during the builder refactor. I'll investigate that later. And it's fine to call TargetedCollection's key pair method, I just was mistakenly believing I'd see a different call when `targeted => false`

Ladas: right, ```targeted => false``` should only change how we look at db. With targeted => true, we do ```ems.key_pairs.where(references(:vms))``` with false we do ``````ems.key_pairs```

So with false it loads all key_pairs and it will delete those you will not .build in parser. While with true, it just loads the ones you've built, so it will just update.

Ladas: in the spec, try also ems.key_pairs, since there might be the bug with :resource being nil

mansam: Doesn't seem to make a difference either way, though I'm continuing to investigate.

mansam: @Ladas Nevermind me, the spec wasn't written properly. >.>

app/models/manageiq/providers/openstack/inventory/persister/definitions/cloud_collections.rb
```diff
@@ -123,9 +123,10 @@ def add_key_pairs(extra_properties = {})
     add_collection(cloud, :key_pairs, extra_properties) do |builder|
       builder.add_properties(
         :model_class => ManageIQ::Providers::Openstack::CloudManager::AuthKeyPair,
+        # targeted refresh workaround-- always refresh the whole keypair collection
+        :targeted    => false
```
Ladas: I think we still need to conditionalize this , it should copy what we do in https://github.com/ManageIQ/manageiq-providers-openstack/pull/323/files#diff-0698b10bec7e7c31e8eabb30dc507213R140

So we fetch all entities only if there is something in references. 
---

Now, it the references would be blank, we would delete all keypairs

app/models/manageiq/providers/openstack/inventory/persister/definitions/storage_collections.rb
```diff
@@ -17,5 +16,22 @@ def initialize_storage_inventory_collections
...
+
+  def add_cloud_volume_backups(extra_properties = {})
+    add_collection(cloud, :cloud_volume_backups, extra_properties) do |builder|
+      if targeted?
+        builder.add_properties(:parent => manager.cinder_manager)
+        builder.add_default_values(:ems_id => manager.cinder_manager.try(:id))
+      else
+        builder.add_default_values(:ems_id => manager.id)
+      end
+      # targeted refresh workaround-- always refresh the whole backup collection
```
Ladas: I would probably write here why:

``Because OpenStack doesn't give us UUIDs of changed volume_backups, we just get an event that one of them changed```

Btw. is this a bug? :-D It's like super unhelpful event. :-) Or does it use ID we don't recognize?

mansam: Good call on the improved comment. I'm not sure whether this is officially a bug, but it is a really unhelpful event. There are some other IDs present like the request ID, but nothing that helps us as far as I know.

aufi:
Looks good! Merging.

PR #324, opened 2018-07-25
New Cloud provider: Fix event creds validation for AMQP

gildub:
Splits `#raw_connect` to target separately AMQP availability and ems default connect.

https://bugzilla.redhat.com/show_bug.cgi?id=1593663

### DIFF
--- a/app/models/manageiq/providers/openstack/manager_mixin.rb
+++ b/app/models/manageiq/providers/openstack/manager_mixin.rb
@@ -13,28 +13,38 @@ module ManageIQ::Providers::Openstack::ManagerMixin
   #
   module ClassMethods
     def raw_connect(password, params, service = "Compute")
-      ems = new
-      ems.name                   = params[:name].strip
-      ems.provider_region        = params[:provider_region]
-      ems.api_version            = params[:api_version].strip
-      ems.security_protocol      = params[:default_security_protocol].strip
-      ems.keystone_v3_domain_id  = params[:keystone_v3_domain_id]
+      if params[:amqp_hostname].present?
+        require 'manageiq/providers/openstack/legacy/events/openstack_rabbit_event_monitor'
+        OpenstackRabbitEventMonitor.available?(
+          :hostname => params[:amqp_hostname],
+          :username => params[:amqp_userid],
+          :password => params[:amqp_password],
+          :port     => params[:amqp_api_port]
+        )
+      else
+        ems = new
+        ems.name                   = params[:name].strip
+        ems.provider_region        = params[:provider_region]
+        ems.api_version            = params[:api_version].strip
+        ems.security_protocol      = params[:default_security_protocol].strip
+        ems.keystone_v3_domain_id  = params[:keystone_v3_domain_id]
 
-      user, hostname, port = params[:default_userid], params[:default_hostname].strip, params[:default_api_port].strip
+        user, hostname, port = params[:default_userid], params[:default_hostname].strip, params[:default_api_port].strip
 
-      endpoint = {:role => :default, :hostname => hostname, :port => port, :security_protocol => ems.security_protocol}
-      authentication = {:userid => user, :password => MiqPassword.try_decrypt(password), :save => false, :role => 'default', :authtype => 'default'}
-      ems.connection_configurations = [{:endpoint       => endpoint,
-                                        :authentication => authentication}]
+        endpoint = {:role => :default, :hostname => hostname, :port => port, :security_protocol => ems.security_protocol}
+        authentication = {:userid => user, :password => MiqPassword.try_decrypt(password), :save => false, :role => 'default', :authtype => 'default'}
+        ems.connection_configurations = [{:endpoint       => endpoint,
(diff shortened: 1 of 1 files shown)


gildub:
@miq-bot add_label bug, blocker, compute/cloud

gildub:
@miq-bot add_reviewer @aufi, @sseago, @andyvesel

gildub:
@miq-bot add_reviewer @sseago

app/models/manageiq/providers/openstack/manager_mixin.rb
```diff
@@ -13,28 +13,38 @@ module ManageIQ::Providers::Openstack::ManagerMixin
...
-
-      begin
-        ems.connect(:service => service)
-      rescue => err
-        miq_exception = translate_exception(err)
-        raise unless miq_exception
-
-        _log.error("Error Class=#{err.class.name}, Message=#{err.message}")
-        raise miq_exception
+      if params[:amqp_hostname].present?
```
aufi: I trust that this works, but I don't feel _raw_connect_ method as a right place for this AMQP connection testing. Could we add an AMQP test method somewhere level-up in OpenStack code?

sseago: @aufi Are you suggesting splitting this method in two, leaving raw_connect as-is, and add something like raw_amqp_connect with the new code, and having the calling method check for amqp_hostname, and call the appropriate raw_ method based on that?

gildub: @aufi, assuming you meant what @sseago detailed, it makes sense meanwhile we would have to test which raw_connect before, in the UI. Please see comment (unhide) https://github.com/ManageIQ/manageiq-ui-classic/pull/4239#pullrequestreview-136224564.

sseago: @gildub Looking at the comments on the UI commit you linked and at the use of raw_connect, I'm thinking that the controllers shouldn't be calling raw_connect at all. If I'm understanding how this currently works correctly, the issue is that there's no guarantee that the UI is running on the same appliance as the backend. If it's not, then any direct OpenStack API call from the UI will fail. This sort of thing needs to be done via the task infrastructure. Once that's done, then the caller of raw_connect won't be in the UI, it will be in the provider, and it can decide *which* raw_connect variant to call (amqp vs. default) without having to mix the two in the same raw_connect method, and at the same time without introducing provider-specific code into the UI or the base manageiq repo.

gildub: @sseago, yes because the UI calls `validate_credentials_task` which is a task wrapper of `raw_connect` so we are in asynchronous mode.

So what I mean is we'd need 2 separate task queues on the UI to determine which one to use depending if we're in AMQP case or not.

Since, the UI/core guys are saying "we call the provider raw_connect (via queued task) and the provider to figure it out", we have to determine the amqp case after. Which is the approach I've taken here.

sseago: @gildub I'd think that the right place for figuring out whether we're dealing with AMQP or not is in whatever method is called by the task queue. If raw_connect is the wrong place for this (as @aufi suggests above), then the answer would be to add a new layer that's called by validate_credentials_task which then calls raw_connect or something else. The other option is to leave it as-is, but split up the implementation of raw_connect into two private methods, one for the amqp case, and one for the non-amqp case, leaving raw_connect itself to just figuring out which method to call. I don't know if there's any significant difference between these approaches, but that depends on why @aufi objected to your raw_connect changes. If it's simply a matter of reducing the complexity of this method, then either approach here is fine. If there's another reason, then we'd need more information.

gildub: @sseago, @aufi,

I added `.raw_event_connect`, see description.

Thanks

agrare: Probably better to check the `:cred_type` here

JPrause:
@gildub as this is for a 5.10 blocker issue, did you have an update on this PR.

gildub:
@JPrause, I'm on it.

JPrause:
@gildub if this will be able to be backported, can you add the gaprindashvili/yes and hammer/yes labels

gildub:
@miq-bot add_label gaprindashvili/yes, hammer/yes

gildub:
@agrare would you mind review it again?

simaishi:
Hammer backport details:

```
$ git log -1
commit 2624415bbe72c8227ee8515b34e0aa77841e8273
Author: Samuel Lucidi <mansam@csh.rit.edu>
Date:   Fri Oct 26 10:37:24 2018 -0400

    Merge pull request #324 from gildub/fix-cloud-provider-event-amqp
    
    New Cloud provider: Fix event creds validation for AMQP
    
    (cherry picked from commit fb91f0eefea53973d4c3e6eba997a44610071c6c)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1593663
```

PR #325, opened 2018-07-26
Collect instance tags during inventory refresh

mansam:


mansam:
@aufi should we be pulling the tags from the server `tags` property, or the `metadata` property? It seems like either could do the job, but they're distinct from one another.

app/models/manageiq/providers/openstack/inventory/persister.rb
```diff
@@ -21,6 +24,13 @@ def initialize(manager, target = nil, collector = nil)
 
   protected
 
+  # TODO: this reads whole table ContainerLabelTagMapping.all.
+  #   Is this expensive for each targeted refresh?
```
aufi: Good question, maybe @Ladas or @agrare might know?  Otherwise I fine with merging.

mansam: That's a comment copied from @Ladas' original code ;)

aufi: Ah, copied from merged code :-)

PR #329, opened 2018-08-02
Flavors: always include private flavors

gildub:
By default private flavors are only accessible by the 'admin' user.
Meanwhile:
- The default policy ruling such access can be modified in OpenStack (policy.json),
- Flavors can also be made explicitly accessible to other projects as well (independently of the policy rule)

Therefore the collector and refresher need to always make sure both "public" and "private" flavors are handled.

https://bugzilla.redhat.com/show_bug.cgi?id=1609568


gildub:
@mansam thank you!

gildub:
@miq-bot add_label bug

PR #223, opened 2018-02-05
Fix Service Provisioning cloud_tenant issue

tinaafitz:
Call new cloud_tenant helper method to avoid provision error.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1532244

Provision error:  
[----] I, [2018-01-31T13:04:42.560968 #4325:f61134]  INFO -- : Q-task_id([miq_provision_1000000000024]) Starting Phase <provision_error>
[----] E, [2018-01-31T13:04:42.629955 #4325:f61134] ERROR -- : Q-task_id([miq_provision_1000000000024]) MIQ(ManageIQ::Providers::Openstack::CloudManager::Provision#provision_error) [[NoMethodError]: undefined method `[]' for nil:NilClass] encountered during phase [start_clone_task]
[----] E, [2018-01-31T13:04:42.630183 #4325:f61134] ERROR -- : Q-task_id([miq_provision_1000000000024]) /var/www/miq/vmdb/app/models/mixins/provider_object_mixin.rb:29:in `connection_source'
/var/www/miq/vmdb/app/models/mixins/provider_object_mixin.rb:5:in `with_provider_connection'
/var/www/miq/vmdb/app/models/manageiq/providers/openstack/cloud_manager/provision/cloning.rb:45:in `start_clone'
/var/www/miq/vmdb/app/models/manageiq/providers/cloud_manager/provision/state_machine.rb:37:in `start_clone_task'
/var/www/miq/vmdb/app/models/miq_request_task/state_machine.rb:21:in `signal'
/var/www/miq/vmdb/app/models/miq_provision/state_machine.rb:18:in `prepare_provision'
/var/www/miq/vmdb/app/models/miq_request_task/state_machine.rb:21:in `signal'
/var/www/miq/vmdb/app/models/manageiq/providers/cloud_manager/provision/state_machine.rb:28:in `poll_volumes_complete'
/var/www/miq/vmdb/app/models/miq_request_task/state_machine.rb:21:in `signal'
/var/www/miq/vmdb/app/models/manageiq/providers/cloud_manager/provision/state_machine.rb:15:in `prepare_volumes'
/var/www/miq/vmdb/app/models/miq_request_task/state_machine.rb:21:in `signal'

Service Dialog populates cloud_tenant as a string, code is expecting it to be an array.
----] I, [2018-01-31T13:04:42.543243 #4325:f61134]  INFO -- :
Q-task_id([miq_provision_1000000000024])
[...]

PR #331, opened 2018-08-08
Fix automatic cloud tenant placement

mansam:
Fixes issues left over from https://github.com/ManageIQ/manageiq-providers-openstack/pull/223
Also fixes https://bugzilla.redhat.com/show_bug.cgi?id=1610736 in combination with https://github.com/ManageIQ/manageiq/pull/17822


aufi:
Looks good to me, merging.

PR #332, opened 2018-08-14
Set tenant access when creating non public flavor

alexander-demichev:
Set tenant access when creating non public flavor. After a non-public flavor is created add tenant access. Array with tenants comes with `create_options ` from UI.
https://bugzilla.redhat.com/show_bug.cgi?id=1609564

@aufi

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/flavor.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/flavor.rb
@@ -3,7 +3,13 @@ class ManageIQ::Providers::Openstack::CloudManager::Flavor < ::Flavor
 
   def self.raw_create_flavor(ext_management_system, create_options)
     ext_management_system.with_provider_connection({:service => 'Compute'}) do |service|
-      service.flavors.create(create_options)
+      cloud_tenant_refs = create_options.delete("cloud_tenant_refs")
+      flavor = service.flavors.create(create_options)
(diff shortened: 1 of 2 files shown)


alexander-demichev:
https://github.com/ManageIQ/manageiq-ui-classic/pull/4469

app/models/manageiq/providers/openstack/cloud_manager/flavor.rb
```diff
@@ -3,7 +3,12 @@ class ManageIQ::Providers::Openstack::CloudManager::Flavor < ::Flavor
 
   def self.raw_create_flavor(ext_management_system, create_options)
     ext_management_system.with_provider_connection({:service => 'Compute'}) do |service|
-      service.flavors.create(create_options)
+      flavor = service.flavors.create(create_options)
```
mansam: Although Fog shouldn't care about `cloud_tenant_refs`being passed to it as part of the options hash, I think it would still be good practice to remove it from the hash before passing it to `create`. It's not a valid parameter for that method, and we have no guarantees that the implementation of `create` in Fog won't change in such a way that causes it to break later.

alexander-demichev:
@mansam I agree, thanks for advice. Is it better now? :)

aufi:
Looks good to me :+1:

mansam:
LGTM!

PR #335, opened 2018-08-16
Add missing ems_id to vnfd template

Ladas:
Add missing ems_id to vnfd template

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/persister/definitions/cloud_collections.rb
+++ b/app/models/manageiq/providers/openstack/inventory/persister/definitions/cloud_collections.rb
@@ -119,6 +119,7 @@ module ManageIQ::Providers::Openstack::Inventory::Persister::Definitions::CloudC
   def add_orchestration_templates
     add_collection(cloud, :orchestration_templates) do |builder|
       builder.add_properties(:model_class => ::OrchestrationTemplate)
+      builder.add_default_values(:ems_id => manager.id)
     end
   end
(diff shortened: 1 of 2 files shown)


Ladas:
@mansam looks like we are missing a spec for this, but if we add validation here https://github.com/ManageIQ/manageiq-providers-openstack/pull/160 it becomes an error

app/models/manageiq/providers/openstack/inventory/persister/definitions/cloud_collections.rb
```diff
@@ -119,6 +119,7 @@ def add_orchestration_stacks(extra_properties = {})
   def add_orchestration_templates
     add_collection(cloud, :orchestration_templates) do |builder|
       builder.add_properties(:model_class => ::OrchestrationTemplate)
+      builder.add_default_values(:ems_id => manager.id)
```
Ladas: hm, now I am not sure if adding it to all templates can cause problems

Ladas: maybe it will be better to add it to parser

mansam: Any further thoughts on this?

Ladas: I am not sure :-)

PR #339, opened 2018-08-28
Use nil as initialiser for @orchestration_stacks to prevent looping.

adamgoossens:
Hi everyone,

This PR resolves an issue I discovered where we will query Heat more than once per tenant during the NetworkManager refresh in environments where there are no orchestration stacks.

Where there are no orchestration stacks an empty list is always returned, which causes .any? to return False, and so doesn't stop another round of queries to Heat the next time orchestration_stacks is called.

The number of queries scales according to the number of cloud networks and security groups in the environment, as we attempt to match up their parent stack by calling orchestration_stacks...which triggers yet another batch of queries to Heat, one per tenant.

I did see a possible opportunity to adjust the initialisers for the other lists in Collector to also use the same logic, but given I'm not intimately familiar with the code I figured I'd narrow this PR to just orchestration_stacks and look at the others later, if there's interest.

Thanks!

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/collector.rb
+++ b/app/models/manageiq/providers/openstack/inventory/collector.rb
@@ -46,7 +46,7 @@ class ManageIQ::Providers::Openstack::Inventory::Collector < ManageIQ::Providers
     @host_aggregates           = []
     @key_pairs                 = []
     @images                    = []
-    @orchestration_stacks      = []
+    @orchestration_stacks      = nil
     @quotas                    = []
     @vms                       = []
     @vnfs                      = []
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/openstack/inventory/collector.rb
```diff
@@ -46,7 +46,7 @@ def initialize_inventory_sources
     @host_aggregates           = []
     @key_pairs                 = []
     @images                    = []
-    @orchestration_stacks      = []
+    @orchestration_stacks      = nil
```
Ladas: @mansam I wonder if we can just ditch these definitions and use the nil checks everywhere?

mansam: Yeah, these can probably be changed in the same way.

mansam:
Looks great, thanks!

gildub:

gildub:
This is going to help with
https://bugzilla.redhat.com/show_bug.cgi?id=1589065

PR #341, opened 2018-08-29
bug 1513086: add cores per socket to OpenStack cloud inventory parser

sseago:
https://bugzilla.redhat.com/show_bug.cgi?id=1513086

For OpenStack cloud, cpu_cores_per_socket was always set to 1 in the legacy refresh_parser
since OpenStack flavors only have a single "number of CPUs" param. The new inventory
parser was not setting this param at all, though, leaving it as nil.


sseago:
Note that if this PR needs to be backported to gaprindashvili it's not a clean backport. To work on gaprindashvili we also need to modify app/models/manageiq/providers/openstack/inventory_collection_default/cloud_manager.rb which has been deleted on master, so that change wasn't needed for this to work on master.

aufi:
Looks good to me, merging!

PR #342, opened 2018-09-04
Filter Keystone Projects by domain_id

aufi:
In order to connect from MIQ to OpenStack, MIQ lists projects
accessible for current User.

OpenStack provides list of all projects accessible for the User. But
MIQ allows to specify domain_id in Add provider form, so it is
expected see only project within such domain.

Adding filtering of projects available for the user by domain_id.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1593923


simaishi:
@aufi Can this be `gaprindashvili/yes`?

aufi:
Yes, please! The associated BZ is https://bugzilla.redhat.com/show_bug.cgi?id=1629125

PR #344, opened 2018-09-06
Use empty name for volume if only size provided

alexander-demichev:
Volumes that don't have name should be filtered from provisioning. When there is no name infinite loop may appear, so I added `index` variable.

https://bugzilla.redhat.com/show_bug.cgi?id=1625970

@aufi @mansam

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/provision_workflow.rb
@@ -54,10 +54,11 @@ class ManageIQ::Providers::Openstack::CloudManager::ProvisionWorkflow < ::MiqPro
         indexed_key = :"#{key}_#{volumes.length + 1}"
         new_volume[key.to_sym] = values[indexed_key] if values.key?(indexed_key)
       end
-      if new_volume.blank?
+      if new_volume.blank? || new_volume.values.all?(&:blank?)
         prepare_volumes = false
       else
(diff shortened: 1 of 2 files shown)


mansam:
If I understand the BZ correctly, the reporter wants to be able to create a volume without supplying a volume name. This patch looks like it would skip creating volumes without names, which wouldn't satisfy the reporter's request.

Loicavenel:
I think there are 2 things here: 

It the name is mandatory, then we should make sure this is indicated in the UI and More imporant don't let user summit if it is not filled like for the other mandatory field.

OR if it is not mandatory, then let it go through but don't break later... BUT does OpenStack will be happy and create a Name???

aufi:
From OpenStack point of view, name is not mandatory (can be empty or duplicate with other volumes), but in MIQ Instance provisioning form, we use non-empty name as an indication that the volume should be created.

I was in contact with reporter and I believe that skipping volumes without name makes sense (volume name doesn't have to be unique). Will follow-up on the BZ.

mansam:
Looks good!

simaishi:
@alexander-demichev Can this be `gaprindashvili/yes` ?

simaishi:
actually... this will conflict on backport. Please create a separate PR for Gaprindashvili branch.

simaishi:
Backported to Gaprindashvili via https://github.com/ManageIQ/manageiq-providers-openstack/pull/367/

PR #347, opened 2018-09-07
Add 'default_security_group' method to CloudTenant

mansam:
Adds a `default_security_group` instance method to CloudTenant that returns a security group named `default` if that CloudTenant has one, or else returns the security group with the highest population of VMs.


aufi:

kbrock:
This method is really bad. the large number of security groups and then the N+1 count for each of the vm counts is really bad.

Also, making this a virtual attribute means it is exposed for everyone - it probably needs to move to the core

It still is a bunch of work, but at least is done on the database:
`security_groups.order(:total_vms).last`

I'll ping @Ladas

mansam:
@ladas @kbrock Ordering via `security_groups.order(:total_vms)` throws an exception -- missing method 'asc' for NilClass. I'm working on writing a better join since this indeed isn't efficient enough.

kbrock:
@mansam thanks - will look into that

kbrock:
@mansam you are right - it looks like that is using a `:through` which currently does not handle this abstraction.

Will try and think up another way to avoid the N+1 queries

PR #348, opened 2018-09-10
Add filtering of image attributes

andyvesel:
Added filtering in `options` from unnecessary data, which doesn't belong to image attributes

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/template.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/template.rb
@@ -99,6 +99,8 @@ class ManageIQ::Providers::Openstack::CloudManager::Template < ManageIQ::Provide
 
   def raw_update_image(options)
     ext_management_system.with_provider_connection(:service => 'Image') do |service|
+      image_attrs = service.images.find_by_id(ems_ref).attributes.stringify_keys
+      options = options.select { |k| image_attrs.key?(k) }
       service.images.find_by_id(ems_ref).update(options)
     end
   rescue => err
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/openstack/cloud_manager/template.rb
```diff
@@ -99,6 +99,8 @@ def self.create_image(ext_management_system, create_options)
 
   def raw_update_image(options)
     ext_management_system.with_provider_connection(:service => 'Image') do |service|
+      image_attrs = self.attributes
```
aufi: This seems to fix issue with miq_queue keys in attributes, but to have it fully correct, it is needed check attributes against Fog Image instead of MIQ Template.

```self.attributes``` are attributes of Template class, ```service.images.find_by_id(ems_ref)``` returns Fog object - https://github.com/fog/fog-openstack/blob/master/lib/fog/image/openstack/v2/models/image.rb its attributes should be checked.

alexander-demichev:
Should this filtering also be added to create action?

app/models/manageiq/providers/openstack/cloud_manager/template.rb
andyvesel: Updated PR

mansam:
Looks fine, but why are invalid parameters making it into the update call in the first place?

mansam:
@andyvesel Alright, thanks for the explanation. This workaround is fine, but we should discover why these incorrect parameters are reaching this method.

andyvesel:
> @andyvesel Alright, thanks for the explanation. This workaround is fine, but we should discover why these incorrect parameters are reaching this method.

Agree with that, I'll figure this out

mansam:
@andyvesel It looks like those extra parameters are inserted by the MiqTask and MiqQueue classes for some reason.

https://github.com/ManageIQ/manageiq/pull/17015
https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_queue.rb#L557
https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_task.rb#L292

PR #352, opened 2018-09-11
Add StorageManager modules to fix CinderManager import paths

mansam:
One of the recent inventory refactor PRs moved the location of the CinderManager inventory modules, which causes problems in the console and in other situations because the `require_nested :CinderManager` calls don't work. This adds modules to fix the import path and updates the `require_nested` call.


Ladas:
@mansam looks like a core is failing, let me try to fix that

aufi:

PR #353, opened 2018-09-12
Replace inefficient default_security_group sort with join

mansam:
@kbrock @Ladas Please sanity check this, I don't write many queries these days.


mansam:
@Ladas there is a spec for this in the original PR: https://github.com/ManageIQ/manageiq-providers-openstack/pull/347/files

PR #357, opened 2018-09-18
[WIP] Download image data using url and upload it using fog

alexander-demichev:
This PR adds method `handle_upload ` for downloading an image using link, and the uploads it to openstack.
I think it's the easiest way to upload image iso using miq. I don't find `open` method suitable here, but I'm also not sure if we have some replacement in miq. 
Please share your opinions @aufi @andyvesel @mansam 
https://github.com/ManageIQ/manageiq-ui-classic/issues/3944

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/template.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/template.rb
@@ -76,7 +76,9 @@ class ManageIQ::Providers::Openstack::CloudManager::Template < ManageIQ::Provide
 
   def self.raw_create_image(ext_management_system, create_options)
     ext_management_system.with_provider_connection(:service => 'Image') do |service|
-      service.create_image(create_options)
+      url = create_options.delete(:url)
+      image = service.images.create(create_options)
+      ManageIQ::Providers::Openstack::CloudManager::UploadImageWorkflow.create_job(ext_management_system, image.id, url)
     end
   rescue => err
     _log.error("image=[#{name}], error=[#{err}]")
--- a/app/models/manageiq/providers/openstack/cloud_manager/upload_image_workflow.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/upload_image_workflow.rb
@@ -1,0 +1,91 @@
+class ManageIQ::Providers::Openstack::CloudManager::UploadImageWorkflow < Job
+  def self.create_job(ext_management_system_id, image_id, url, timeout: 1.hour, poll_interval: 1.minute)
+    options = {
+      :ext_management_system_id => ext_management_system_id,
+      :url                      => url,
+      :image_id                 => image_id,
+      :timeout                  => timeout,
+      :poll_interval            => poll_interval,
+    }
+
+    super(name, options)
+  end
+
+  def upload_image
+    ext_management_system_id, image_id, url = options.values_at(:ext_management_system, :image_id, :url)
+
+    EmsCloud.find(ext_management_system_id).with_provider_connection(:service => 'Image') do |service|
+      options[:uploading] = true
+      started_on = Time.now.utc
+      update_attributes!(:started_on => started_on)
+      miq_task.update_attributes!(:started_on => started_on)
+      queue_signal(:poll_runner)
+
+      image = service.images.find_by_id(image_id)
+      service.handle_upload(image, url)
+      options[:uploading] = false
+    end
+    rescue => err
+      _log.error("image=[#{name}], error=[#{err}]")
+      queue_signal(:abort, "Failed to upload image", "error")
+  end
+
+  def poll_runner
+    uploading = context[:uploading?]
+    if uploading
+      if started_on + options[:timeout] < Time.now.utc
+        queue_signal(:abort, "Playbook has been running longer than timeout", "error")
+      else
+        queue_signal(:poll_runner, :deliver_on => deliver_on)
+      end
+    else
+      queue_signal(:post_upload_image)
+    end
+  end
+
+  alias initializing dispatch_start
+  alias start        upload_image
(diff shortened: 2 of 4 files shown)


lib/manageiq/providers/openstack/legacy/openstack_handle/image_delegate.rb
```diff
@@ -45,5 +45,13 @@ def images_for_accessible_tenants
       @os_handle.accessor_for_accessible_tenants(
         SERVICE_NAME, :images_with_pagination_loop, :id)
     end
+
+    def handle_upload(image, url)
+      file = open(url) if url =~ URI::DEFAULT_PARSER.make_regexp
```
Ladas: hm, this will do it in memory right? We should probably fetch it to /tmp and never hold the image data in memory.

Also this will be done by generic worker, so it will be killed in 10 minutes, so we'll need a separate longrunning job wrapped into a state machine.

cc @agrare can you point to some state machine examples? And if we have some code for handling big files. ^

app/models/manageiq/providers/openstack/cloud_manager/template.rb
```diff
@@ -76,7 +76,9 @@ def requires_storage_for_scan?
 
   def self.raw_create_image(ext_management_system, create_options)
     ext_management_system.with_provider_connection(:service => 'Image') do |service|
-      service.create_image(create_options)
+      url = create_options.delete(:url)
+      image = service.images.create(create_options)
+      service.handle_upload(image, url)
```
Ladas: hm, what was issue with this approach? This was letting the OpenStack to fetch the url, so MIQ doesn't need to fetch all of the data (the images can be quite large easily 50GB and more)

alexander-demichev: Letting OpenStack to fetch url is supported only in Image v1, and deprecated in Image v2

lib/manageiq/providers/openstack/legacy/openstack_handle/image_delegate.rb
aufi: The ```open``` is expected to return TempFile for larger data files.

app/models/manageiq/providers/openstack/cloud_manager/template.rb
aufi: As Alexander wrote, passing URL doesn't work e.g. on OSP13.

lib/manageiq/providers/openstack/legacy/openstack_handle/image_delegate.rb
agrare: Hey sorry I missed this.  :+1: for not loading the entire file into memory :smile:

We have a state machine for ansible-runner which will exec a process to do something and the state machine monitors it, https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/ansible_runner_workflow.rb

You could do something similar with wget -> /tmp and use a state machine similar to ^ to wait for it to exit.

alexander-demichev:
@miq-bot add_label wip

app/models/manageiq/providers/openstack/cloud_manager/upload_image_workflow.rb
```diff
@@ -0,0 +1,76 @@
+class ManageIQ::Providers::Openstack::CloudManager::UploadImageWorkflow < Job
+  def self.create_job(ext_management_system_id, image_id, url)
```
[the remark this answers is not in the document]
agrare: Hey @alexander-demichev yeah this looks pretty close

app/models/manageiq/providers/openstack/cloud_manager/upload_image_workflow.rb
```diff
@@ -0,0 +1,76 @@
...
+      :ext_management_system_id => ext_management_system_id,
+      :url                      => url,
+      :image_id                 => image_id,
+    }
+
+    super(name, options)
+  end
+
+  def pre_upload_image
+    queue_signal(:upload_image)
```
agrare: Unless you intend to override this in a child class this doesn't look necessary, you can just have your `:start` signal `=> :upload_image`

app/models/manageiq/providers/openstack/cloud_manager/upload_image_workflow.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  def upload_image
+    ext_management_system_id, image_id, url = options.values_at(:ext_management_system, :image_id, :url)
+
+    EmsCloud.find(ext_management_system_id).with_provider_connection(:service => 'Image') do |service|
+      started_on = Time.now.utc
+      update_attributes!(:started_on => started_on)
+
+      image = service.images.find_by_id(image_id)
+      service.handle_upload(image, url)
```
agrare: Does this do the upload synchronously?  That's not going to work as this will likely timeout just like we were worried about without the state machine if it takes >10 mins.

The idea is that your `upload_image` state would start the upload, add some context key to the options hash, and queue the next state which check if the upload is finished.  If it is finished it signals finish and if not it requeues itself for e.g. 1 minute in the future.

This means that no queue method being handled by a worker is doing any long term blocking calls so the upload could theoretically take hours and we'd be just fine.

alexander-demichev:
@agrare Hi, I updated the PR, can you take a look at this again?

app/models/manageiq/providers/openstack/cloud_manager/upload_image_workflow.rb
agrare: In order for this to work you need to start the upload to run async in the background, queue the poll_upload method and return out of this method.

PR #359, opened 2018-09-24
Pass openstack_admin? flag to volume snapshot template collection

mansam:
Adds missing `openstack_admin?` to the Cloud Manager's volume snapshot templates collection.


aufi:
Looks good to me, merging!

PR #334, opened 2018-08-15
A fog-openstack with a more "version less" approach

gildub:
Changes necessary to use fog-openstack supporting "version less" endpoint (including identity) and which avoid tempering with URL path.

Tested: Cloud/Infra on Gaprindashvili and master

Requires https://github.com/ManageIQ/manageiq-gems-pending/pull/386

PR #360, opened 2018-09-25
Bump fog-openstack version to 0.3.x

gildub:


agrare:
Now that hammer is cut can we update fog-openstack to a version that uses a newer fog-core?

gildub:
@agrare, but we have to be able to merge https://github.com/ManageIQ/manageiq-providers-openstack/pull/334 for that to happen. The current PR is actually redundant but serves as a tracker.

PR #368, opened 2018-10-03
[HAMMER] Update i18n catalog for hammer

mzazrivec:


mzazrivec:
@agrare yup, I'll fix this in a separate PR though (seeing the fix is in code and needs to land in master as well).

mzazrivec:
@agrare here's the promised fix: https://github.com/ManageIQ/manageiq-providers-openstack/pull/369

agrare:
:+1:

PR #380, opened 2018-10-22
Check if host create event exists and assign host to it

alexander-demichev:
After creating host check if `compute.instance.create.end` event exists and assign host to it.
I don't know if there is a better way to assign a host, querying all events and then selecting hosts doesn't look very efficient, but I didn't came up with something better.

@mansam  Can you review it? :)

https://bugzilla.redhat.com/show_bug.cgi?id=1584770

### DIFF
--- a/app/models/manageiq/providers/openstack/infra_manager/host.rb
+++ b/app/models/manageiq/providers/openstack/infra_manager/host.rb
@@ -455,4 +455,30 @@ class ManageIQ::Providers::Openstack::InfraManager::Host < ::Host
   def self.display_name(number = 1)
     n_('Host (OpenStack)', 'Hosts (OpenStack)', number)
   end
+
+  def self.post_refresh_ems(ems_id, update_start_time)
+    ems = ExtManagementSystem.find(ems_id)
+    hosts = ems.hosts.where("created_on >= ?", update_start_time)
+    hosts.find_each(&:post_create_actions_queue)
+  end
+
+  def post_create_actions_queue
+    MiqQueue.submit_job(
+      :class_name  => self.class.name,
+      :instance_id => id,
+      :method_name => 'post_create_actions'
+    )
+  end
+
+  def post_create_actions
+    update_create_event
+  end
+
+  def update_create_event
+    parsed_ems_ref_obj = YAML.safe_load(ems_ref_obj)
+    create_event = ext_management_system.ems_events.find_by(:host_id    => nil,
+                                                            :event_type => "compute.instance.create.end",
+                                                            :host_name  => parsed_ems_ref_obj)
+    create_event&.update_attributes!(:host_id => id)
+  end
 end
--- a/app/models/manageiq/providers/openstack/infra_manager/event_parser.rb
+++ b/app/models/manageiq/providers/openstack/infra_manager/event_parser.rb
@@ -20,6 +20,7 @@ module ManageIQ::Providers::Openstack::InfraManager::EventParser
 
     if payload.key? "instance_id"
       event_hash[:host_id] = Host.find_by("ems_ref_obj" => YAML.dump(payload["instance_id"])).try(:id)
+      event_hash[:host_name] = payload["instance_id"]
     end
     event_hash[:message]                   = payload["message"]           if payload.key? "message"
     event_hash[:host_ems_ref]              = payload["node"]              if payload.key? "node"
--- a/app/models/manageiq/providers/openstack/infra_manager/refresher.rb
+++ b/app/models/manageiq/providers/openstack/infra_manager/refresher.rb
@@ -13,7 +13,7 @@ module ManageIQ
       end
 
       def post_process_refresh_classes
-        [::Vm]
+        [::Vm, ManageIQ::Providers::Openstack::InfraManager::Host]
       end
     end
   end
--- a/spec/models/manageiq/providers/openstack/infra_manager/host_spec.rb
+++ b/spec/models/manageiq/providers/openstack/infra_manager/host_spec.rb
@@ -578,4 +578,28 @@ describe ManageIQ::Providers::Openstack::InfraManager::Host do
       expect(MiqQueue.where(:method_name => "refresh").count).to eq(1)
     end
   end
+
+  describe 'after create callback' do
+    let(:ext_management_system) { FactoryGirl.create(:ems_openstack_infra) }
+
+    it "assigns scale out event fot host" do
+      payload = { "instance_id" => "openstack-perf-host-nova-instance" }
+
+      event_hash = {
+        :event_type => "compute.instance.create.end",
+        :source     => "OPENSTACK",
+        :message    => payload,
+        :timestamp  => "2016-03-13T16:59:01.760000",
+        :username   => "",
+        :full_data  => {:content => {'payload' => payload}},
+        :ems_id     => ext_management_system.id,
+        :host_name  => "openstack-perf-host-nova-instance"
+      }
+
+      EmsEvent.add(ext_management_system.id, event_hash)
+      host = FactoryGirl.create(:host_openstack_infra, :ext_management_system => ext_management_system)
+      host.update_create_event
+      expect(ext_management_system.ems_events.first.host_id).to eq(host.id)
+    end
+  end
 end


mansam:
Querying the list of events and then iterating through them doing comparisons on the message data is going to be pretty slow. Maybe something like `EmsEvent.where("message LIKE ? and event_type = 'compute.instance.create.end'", "%instance_id\"=>\"#{instance_id}\"%")` would be better, since it lets the database do the work.
`

mansam:
Actually, it looks like there's similar existing code that should be doing this already: https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/infra_manager/event_parser.rb#L21
Can you fix or refactor that to set the host id?

app/models/manageiq/providers/openstack/infra_manager/host.rb
```diff
@@ -439,4 +441,10 @@ def refresh_network_interfaces(ssu)
   def self.display_name(number = 1)
     n_('Host (OpenStack)', 'Hosts (OpenStack)', number)
   end
+
+  def update_create_event
+    events = ext_management_system.ems_events.where(:event_type => "compute.instance.create.end")
```
aufi: I'd suggest filter event_streams records with not null host_id, otherwise looks good to me and hopefully should be fast since it covers only undercloud Host and fields we're querying (ems_id, event_type and host_id) are indexed.

app/models/manageiq/providers/openstack/infra_manager/host.rb
```diff
@@ -439,4 +441,10 @@ def refresh_network_interfaces(ssu)
...
+    create_event = events.find { |event| event.full_data[:content]["payload"]["instance_id"] == YAML.load(ems_ref_obj) }
```
aufi: I'd also suggest run YAML.load just once.

agrare:
@alexander-demichev we already have a way of doing this sort of thing with post_refresh_ems hook.  We reconnect_events for the VmOrTemplate model.

There is a [`post_refresh_ems`](https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_or_template.rb#L1101-L1120) method defined on the class which finds [hosts added in the last refresh](https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_or_template.rb#L1106) and [queues a post_create_action](https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_or_template.rb#L1112) one of which is to [reconnect events](https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_or_template.rb#L1200).  Here is the full method: https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_or_template.rb#L834-L860

One of the nice performance gains here is that we are looking at the [vm_ems_ref column](https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_or_template.rb#L835) which we added recently.  I'd recommend adding a host_ems_ref to the schema for event_streams and use that instead of having to evaluate the context in ruby.  We can do something else for backport if needed like match on the host_name and only look at the `event.full_data[:content]["payload"]["instance_id"]` for hosts with duplicate names.

alexander-demichev:
@agrare Hi, can you review this again?

alexander-demichev:
I'm not sure about adding host_ems_ref to event streams, the lookup is done using host's ems_ref_obj, it's uuid that comes from ironic and it's different from ems_ref. I think there is no way to assign host_ems_ref in event parser, because host is not fetched yet.

agrare:
Okay the host's ems_ref isn't available in the event payload?  We could add a host_uuid instead of a host_ems_ref to event_streams.  I don't want to add a host_ems_ref_obj without a host_ems_ref but it sounds like the ems_ref_obj is just the uuid?

alexander-demichev:
I've just checked host using openstack cli, and it seems that ems_ref_obj refers to host's id as an object in openstack, and ems_ref refers to 'OS-EXT-SRV-ATTR:hypervisor_hostname' field. I don't know why we have it this way.
@Ladas Hi, do you know why it's like this?

Ladas:
@alexander-demichev from what I remember we needed a different id for the match and we needed to backport it, so that is why ems_ref_obj was used for the second id. But :ems_ref_obj is used just for connecting Overcloud to Undercloud

Ladas:
@alexander-demichev so talking to @agrare we could store the other id into Host<:uid_ems> and Event<:host_name>. I believe those are not being used for anything, correct? Then you should be able to use the code @agrare pasted (especially the query that finds the events that needs to be connected)

So that will solve any future events. 

Then we can talk about solving past events, I think it might be ok to just provide a ```rails c``` one liner into the BZ, that will take all OpenStack events of interest and populate the ```host_name```, then the next refresh should connect them.

alexander-demichev:
I updated `update_create_event ` method, if Event<:host_name> is not used for anything, this solution should work. @agrare @Ladas

app/models/manageiq/providers/openstack/infra_manager/host.rb
```diff
@@ -455,4 +455,31 @@ def refresh_network_interfaces(ssu)
   def self.display_name(number = 1)
     n_('Host (OpenStack)', 'Hosts (OpenStack)', number)
   end
+
+  def self.post_refresh_ems(ems_id, _update_start_time)
+    ems = ExtManagementSystem.find(ems_id)
+    hosts = ems.hosts
```
agrare: You should be able to limit this down to hosts created in the last refresh e.g. https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_or_template.rb#L1106

alexander-demichev: Is it going to query hosts if first refresh fails?

agrare: `post_refresh` is only called on successful refresh https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/base_manager/refresher.rb#L62-L63

app/models/manageiq/providers/openstack/infra_manager/host.rb
```diff
@@ -455,4 +455,31 @@ def refresh_network_interfaces(ssu)
   def self.display_name(number = 1)
     n_('Host (OpenStack)', 'Hosts (OpenStack)', number)
   end
+
+  def self.post_refresh_ems(ems_id, update_start_time)
+    ems = ExtManagementSystem.find(ems_id)
+    hosts = ems.hosts.where("created_on >= ?", update_start_time)
+    hosts.find_each(&:post_create_actions_queue) unless hosts.empty?
```
agrare: `.where` will return an empty relation so no need to worry about if `.empty?` since `[].find_each(&:....)` will just be a noop

app/models/manageiq/providers/openstack/infra_manager/host.rb
```diff
@@ -455,4 +455,31 @@ def refresh_network_interfaces(ssu)
...
+  def post_create_actions
+    update_create_event
+  end
+
+  def update_create_event
+    parsed_ems_ref_obj = YAML.safe_load(ems_ref_obj)
+    create_event = ext_management_system.ems_events.find_by(:host_id    => nil,
+                                                            :event_type => "compute.instance.create.end",
+                                                            :host_name  => parsed_ems_ref_obj,
+                                                            :ems_id     => ext_management_system.id)
```
agrare: If you're doing `ext_management_system.ems_events` you don't also need `:ems_id => ext_management_system.di`

app/models/manageiq/providers/openstack/infra_manager/host.rb
```diff
@@ -455,4 +455,31 @@ def refresh_network_interfaces(ssu)
...
+    create_event&.update(:host_id => id)
```
agrare: prefer `.update_attributes!`

agrare:
Looking pretty good @alexander-demichev just a few comments

agrare:
@aufi @Ladas @mansam can you give it another look?

alexander-demichev:

alexander-demichev:
@miq-bot add_label hammer/yes

simaishi:
Hammer backport details:

```
$ git log -1
commit 5d0d3b93a4a2d5d13dc6bebb5c3eef0dd6437db9
Date:   Mon Nov 26 08:30:05 2018 -0500

    Merge pull request #380 from alexander-demichev/add-host-create-event
    
    Check if host create event exists and assign host to it
    
    (cherry picked from commit 3b93ace71d28f031f09a7d35063c4e054ca22fcc)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1584770
```

PR #386, opened 2018-11-01
When doing targeted refresh of a volume, refresh any attached VMs

mansam:
If a volume that is attached to one or more VMs is subject to targeted refresh, but one of the VMs it is attached to hasn't yet been pulled into the inventory, then building the association between the volume and its attachments could fail. This PR collects any attached VMs as part of targeted refresh for a volume.

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
@@ -246,6 +246,8 @@ class ManageIQ::Providers::Openstack::Inventory::Collector::TargetCollection < M
         add_simple_target!(:cloud_tenants, t.ems_ref)
       when OrchestrationStack
         add_simple_target!(:orchestration_stacks, t.ems_ref)
+      when CloudVolume
+        add_simple_target!(:cloud_volumes, t.ems_ref)
       end
     end
   end
@@ -272,6 +274,10 @@ class ManageIQ::Providers::Openstack::Inventory::Collector::TargetCollection < M
       infer_related_orchestration_stacks_ems_refs_db!
       infer_related_orchestration_stacks_ems_refs_api!
     end
+    unless references(:cloud_volumes).blank?
+      infer_related_orchestration_stacks_ems_refs_db!
+      infer_related_cloud_volumes_ems_refs_api!
+    end
   end
 
   def infer_related_orchestration_stacks_ems_refs_db!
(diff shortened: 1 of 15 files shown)


Ladas:
:+1: make sense

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
```diff
@@ -272,6 +274,10 @@ def infer_related_ems_refs!
       infer_related_orchestration_stacks_ems_refs_db!
       infer_related_orchestration_stacks_ems_refs_api!
     end
+    unless references(:cloud_volumes).blank?
```
Ladas: maybe we should move this up before Vms? And call ```target.manager_refs_by_association_reset``` so we do full scan of the attached Vms too?

mansam: `target.manager_refs_by_association_reset` gets called once at the end of initialization after `parse_targets!` completes. Are you saying this would also need to be in the `parse_targets!` loop?

Ladas: Hm, lets probably keep it as you have it now. I think that VM will get all of its relations for it's own event.

aufi:
LGTM, merging!

aufi:
@fdupont-redhat Hi, just to be sure,   do we want backport this (Converted VM refresh fix) into Hammer? (if so, we need a BZ)

simaishi:
@aufi @fdupont-redhat We don't require a BZ for backport yet. Please do confirm if this should be backported to hammer or not.

fdupont-redhat:
@aufi @simaishi Yes, it should be backported to Hammer. It's useful for V2V.

simaishi:
@mansam Most of .yml are conflicting on backport. Please create a separate PR for hammer.

```
Changes to be committed:

	modified:   app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
	modified:   lib/vcr_recorder.rb
	modified:   spec/models/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty_spec.rb
	new file:   spec/vcr_cassettes/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty_volume_targeted_refresh.yml

Unmerged paths:
  (use "git add <file>..." to mark resolution)

	both modified:   spec/vcr_cassettes/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty.yml
	both modified:   spec/vcr_cassettes/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty_fast_refresh.yml
	both modified:   spec/vcr_cassettes/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty_keystone_v3.yml
	both modified:   spec/vcr_cassettes/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty_keystone_v3_fast_refresh.yml
	both modified:   spec/vcr_cassettes/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty_keystone_v3_legacy_fast_refresh.yml
	both modified:   spec/vcr_cassettes/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty_legacy_fast_refresh.yml
	both modified:   spec/vcr_cassettes/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty_network_targeted_refresh.yml
	both modified:   spec/vcr_cassettes/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty_port_targeted_refresh.yml
	both modified:   spec/vcr_cassettes/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty_router_targeted_refresh.yml
	both modified:   spec/vcr_cassettes/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty_stack_targeted_refresh.yml
	both modified:   spec/vcr_cassettes/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty_tenant_targeted_refresh.yml
	both modified:   spec/vcr_cassettes/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty_vm_targeted_refresh.yml
```

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
```diff
@@ -272,6 +274,10 @@ def infer_related_ems_refs!
...
+      infer_related_orchestration_stacks_ems_refs_db!
```
fdupont-redhat: Shouldn't it be `infer_related_cloud_volumes_ems_refs_db!` ?

aufi: Right, thanks fot the catch, fixed in https://github.com/ManageIQ/manageiq-providers-openstack/pull/400

PR #389, opened 2018-11-06
Refresh an attached undercloud when saving changes to an overcloud

mansam:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1445717 by kicking off a refresh when the overcloud is saved.

image

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager.rb
@@ -54,9 +54,17 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
                 :ensure_swift_managers
 
   before_update :ensure_managers_zone_and_provider_region
+  after_save :refresh_parent_infra_provider
 
   def hostname_required?
     enabled?
+  end
+
+  def refresh_parent_infra_provider
+    if provider
+      EmsRefresh.queue_refresh(provider)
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/openstack/cloud_manager.rb
```diff
@@ -54,11 +54,19 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
...
   before_update :ensure_managers_zone_and_provider_region
+  after_save :refresh_parent_infra_provider
 
   def hostname_required?
     enabled?
   end
 
+  def refresh_parent_infra_provider
+    if provider
+      EmsRefresh.queue_refresh(provider)
```
aufi: Is refresh on ```provider``` intentional or there should be something like ```provider.infra_manager```?

mansam: Provider is a property on the cloud manager that refers to the infra manager when they've been attached via the provider edit form.

mansam: At least, that's how it appeared to me. I'll double check to make sure there isn't anything funky going on with that association.

Ladas: I think it might be as @aufi  says

mansam: You're right, we want to check if the provider was changed and then refresh `provider.infra_ems`

PR #394, opened 2018-11-09
Add support for multiple amqp endpoints

alexander-demichev:
This PR adds support for multiple amqp endpoints. Currently, it's possible to use only one endpoint when adding a provider. If amqp server is not available CF should switch to another. Amqp is out of openstack HA proxy, so there is no other way to handle server fails. I replaced raising exceptions with returning false in case of fail for being able to iterate through hostnames.
We would need UI for this.

@aufi @mansam 

https://bugzilla.redhat.com/show_bug.cgi?id=1595558


mansam:
@alexander-demichev Have you been able to test this at all with your dev environment?

alexander-demichev:
@mansam Yes, I tested it, but it's not possible to add even one amqp from UI, it's broken, so after mocking parameters that come from UI everything works. I can give credentials to my env, if you want to test it.

PR #402, opened 2018-11-20
Collect Panko events for all tenants.

mansam:
It appears that by default Panko events are only returned for a single tenant at a time. There's a query parameter that returns all events for an admin user, which this PR enables. https://docs.openstack.org/panko/latest/webapi/v2.html

@fdupont-redhat

### DIFF
--- a/lib/manageiq/providers/openstack/legacy/events/openstack_ceilometer_event_monitor.rb
+++ b/lib/manageiq/providers/openstack/legacy/events/openstack_ceilometer_event_monitor.rb
@@ -95,6 +95,10 @@ class OpenstackCeilometerEventMonitor < OpenstackEventMonitor
       'field' => 'start_timestamp',
       'op'    => 'ge',
       'value' => latest_event_timestamp || ''
+    },
+    {
+      'field' => 'all_tenants',
+      'value' => 'True'
     }]
   end
 


bthurber:
https://bugzilla.redhat.com/show_bug.cgi?id=1651340

fdupont-redhat:
@mansam It should be labelled as `blocker`. I would also label it as `gaprindashvili/yes` as I have feedback from the field that it's an issue for customers using Gaprindashvili / CF 4.6.

Ladas:
@mansam this means we require an admin user? We should probably find a way for non admin users too?

mansam:
@Ladas Right, so without an admin user event collection will continue to work how it does now. For a non-admin user to get events for every tenant they have access to, it looks like we'd have to query the API in a loop, scoping to each tenant in turn. That seems pretty awful, so I'm open to ideas for other approaches. There does appear to be a Panko policy rule that can disable event isolation, but that doesn't seem like a great route to go down and the documentation is spotty.

https://github.com/openstack/panko/blob/master/panko/api/controllers/v2/events.py#L170

Ladas:
@mansam we'd need to do a thread pool, where each tread would manage X tenants (X = tenants_total/pool_size). But yeah, there's not better way than cycling through all tenants, if we need to switch scope for every API query

@agrare did we build something like this elsewhere?

agrare:
@Ladas hm no we had some prototypes but nothing merged.

aufi:
Looks good to me, merging!

simaishi:
Hammer backport details:

```
$ git log -1
commit b2c9895b69f91542d235ca5de9930168ed6e5b3d
Date:   Wed Nov 28 16:12:50 2018 +0100

    Merge pull request #402 from mansam/panko-events-for-all-tenants
    
    Collect Panko events for all tenants.
    
    (cherry picked from commit e68667ff737dd5c80d2e1a501e6e19ca017df4d0)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1651340
```

PR #405, opened 2018-11-28
Add a default parallel thread limit to the settings yaml

mansam:
This is a QoL improvement, should not be necessary for any functionality.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -6,6 +6,7 @@
       :is_global_admin: false
     :inventory_object_refresh: true
     :allow_targeted_refresh: true
+    :parallel_thread_limit: 0
   :openstack_network:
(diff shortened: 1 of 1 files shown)


config/settings.yml
```diff
@@ -6,6 +6,7 @@
       :is_global_admin: false
     :inventory_object_refresh: true
     :allow_targeted_refresh: true
+    :parallel_thread_limit: 0
```
aufi: Just for understanding, 0 means no parallelization (single 1 thread) or disable part of code providing parallel refresh functionality?

mansam: 0 means no parallelization.

aufi: Thanks, merging.

PR #410, opened 2018-12-07
Don't erase OS system info after SmartState Analysis

evgeny-chufarov:
After running SmartState Analysis on OpenStack instances or images, CloudForms shows the detailed information for operating system. However, on the next OpenStack provider refresh, the operating system field will revert to whatever CloudForms can get from the base image.
After adding this PR, CloudForms will check the presence of the OS information in the database. If exists, OS will not be changed.

https://bugzilla.redhat.com/show_bug.cgi?id=1646467

@aufi 
@mansam

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
@@ -173,6 +173,8 @@ class ManageIQ::Providers::Openstack::Inventory::Parser::CloudManager < ManageIQ
   end
 
   def miq_templates
+    @vm_or_template_hash ||= last_scan_hash
+
     collector.images.each do |i|
       parent_server_uid = parse_image_parent_id(i)
       image = persister.miq_templates.find_or_build(i.id)
@@ -201,10 +203,13 @@ class ManageIQ::Providers::Openstack::Inventory::Parser::CloudManager < ManageIQ
       hardware.virtualization_type = i.properties.try(:[], 'hypervisor_type') || i.attributes['hypervisor_type']
 
       operating_system = persister.operating_systems.find_or_build(image)
-      operating_system.vm_or_template = image
-      operating_system.product_name = guest_os
-      operating_system.distribution = i.try(:os_distro)
-      operating_system.version = i.try(:os_version)
+
+      unless find_scan_info(@vm_or_template_hash, i.id)
+        operating_system.vm_or_template = image
+        operating_system.product_name = guest_os
+        operating_system.distribution = i.try(:os_distro)
+        operating_system.version = i.try(:os_version)
+      end
     end
   end
 
@@ -292,6 +297,7 @@ class ManageIQ::Providers::Openstack::Inventory::Parser::CloudManager < ManageIQ
   def vms
     related_infra_ems = collector.manager.provider.try(:infra_ems)
     hosts = related_infra_ems.try(:hosts)
+    @vm_or_template_hash ||= last_scan_hash
 
     collector.vms.each do |s|
       if hosts && !s.os_ext_srv_attr_host.blank?
@@ -340,10 +346,13 @@ class ManageIQ::Providers::Openstack::Inventory::Parser::CloudManager < ManageIQ
       hardware.guest_os = persister.hardwares.lazy_find(miq_template_lazy, :key => :guest_os)
 
       operating_system = persister.operating_systems.find_or_build(server)
-      operating_system.vm_or_template = server
-      operating_system.product_name = persister.operating_systems.lazy_find(miq_template_lazy, :key => :product_name)
-      operating_system.distribution = persister.operating_systems.lazy_find(miq_template_lazy, :key => :distribution)
-      operating_system.version = persister.operating_systems.lazy_find(miq_template_lazy, :key => :version)
+
+      unless find_scan_info(@vm_or_template_hash, s.id)
+        operating_system.vm_or_template = server
+        operating_system.product_name = persister.operating_systems.lazy_find(miq_template_lazy, :key => :product_name)
+        operating_system.distribution = persister.operating_systems.lazy_find(miq_template_lazy, :key => :distribution)
+        operating_system.version = persister.operating_systems.lazy_find(miq_template_lazy, :key => :version)
+      end
 
       unless s.private_ip_address.blank?
         private_network = persister.networks.find_or_build_by(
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -200,11 +200,16 @@ def miq_templates
       hardware.size_on_disk = i.size
       hardware.virtualization_type = i.properties.try(:[], 'hypervisor_type') || i.attributes['hypervisor_type']
 
+      vm = VmOrTemplate.find_by(:ems_ref => i.id)
+      os_name = vm&.operating_system&.product_name
```
aufi: I'm not sure if this kind of lookup in refresh is okay, but let @mansam review!

mansam: This looks like it should be OK, it's one simple query per vm or image. @Ladas, do you have any opinions on this approach?

Ladas: Hm, we usually don't do this, since there can be hundreds of thousands of vms or images.

Btw. I think the approach should be another way. Refresh and smartstate should not be changing the same attributes, we should remove it from one or the other. Right now, refresh would keep rewriting what was set by smartstate.

Fryguy: Refresh and smartstate have always fought over this column,  because for different sources we may or may not get one or the other.  In the classic refresh, one of them was chosen as a winner (I believe smartstate always wins, but I'll check and find a link).

Fryguy: Oh, and we should not be querying in the collector code like this, especially once-per anything as it introduces nasty N+1 performance issues.

evgeny-chufarov: Ok, thank you for comments. I will try to find better solution for this issue ASAP.

Ladas: so we should be storing only ```operating_system.product_name = guest_os``` unless guest_os is unknown

Since it uses ```OperatingSystem.normalize_os_name(i.try(:os_distro) || 'unknown')```, smartstate should be storing there the same normalized value. And we could keep the rest for the smatstate. This approach should solve the BZ?

@Fryguy graph refresh doesn;t do any of ^ examples, since all of them are N+1 and going forward, collector will not able to even reach the DB. So best would be to have different schemas.

agrare: @Ladas we could check the Vm's `last_scan_on` to see if smartstate has been run for the Vm (we can do this direct to the DB for MIQ or using the API in the future for topological) and not set the attribute in the parser so as to not overwrite the value smartstate sets.

evgeny-chufarov: Well..I have deleted lines responsible for setting the OS. Now, there are no changes in OS information after refresh.

agrare: @evgeny-chufarov well we still want to set the OS info if smartstate wasn't run.

Fryguy: Not collecting it feels like the wrong approach as well.  SmartState is an opt-in thing, whereas refresh is not.  I would expect refresh to always collect at least something.

mansam: @Fryguy @agrare Am I understanding properly that the recommended course of action here is for us to query the DB once at collection time for all the instances associated with the manager, and then compare their last scan time?

agrare: @mansam yes that's correct

evgeny-chufarov: @aufi 
@mansam 
Now we are querying the database for information about the **last_scan time** of Smart State Analysis. If it is, the refresh skips this vm/template and checks the others. If no SSA scans were executed yet, refresh will try to collect any info that it can.

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -173,6 +173,9 @@ def quotas
   end
 
   def miq_templates
+
+    @vm_or_template_hash ||= database_query
```
mansam: Is there a reason why this is being made an instance variable instead of local?

evgeny-chufarov: It would be more proper way to use an instance variable because of the multiple using and in order to have an access to it from different places.

mansam: Ah, I see that you are caching the result so that you aren't repeating the query in miq_templates and vms. I think I would recommend putting this in the collector and using a similar caching pattern to the other collector methods, but this is fine.

mansam:
Thanks @evgeny-chufarov, looks pretty good. Could you write a test for this?

evgeny-chufarov:
Yes

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -528,4 +537,15 @@ def image_tenants(image)
...
+
+  def find_scan_info(vm_or_template_hash, ems_ref)
+    scan_time = vm_or_template_hash.select { |vm_or_template| vm_or_template["ems_ref"] == ems_ref }
+    scan_time[0]["last_scan_on"] unless scan_time.empty?
+  end
+
+  private
+
+  def last_scan_hash
+    VmOrTemplate.select(:ems_ref, :last_scan_on).where(:ems_id => collector.manager.id)
```
Ladas: lets do

```
VmOrTemplate.select(:ems_ref, :last_scan_on).where(:ems_id => collector.manager.id).where.not(:last_scan_on => nil)
```

Ladas: also use `.index_by` to turn this into hash `{"ems_ref" => "last_scan_on"}, then the lookup will be O(1)

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -528,4 +537,15 @@ def image_tenants(image)
     end
     tenants
   end
+
+  def find_scan_info(vm_or_template_hash, ems_ref)
+    scan_time = vm_or_template_hash.select { |vm_or_template| vm_or_template["ems_ref"] == ems_ref }
```
Ladas: so this is O(n) lookup, lets use hash instead

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
Ladas: and memoize it here

Fryguy:
@evgeny-chufarov I believe this can be closed now that https://github.com/ManageIQ/manageiq/pull/18477 has been merged

evgeny-chufarov:
Yes, I think so, ok.

PR #411, opened 2018-12-11
Improve handling of missing services

mansam:
The absence of volume, networking, or image services
from the service catalog should not cause refresh to
completely fail. This PR is intended to increase the
resiliency of the inventory collecting process when
optional services are missing.


aufi:

PR #416, opened 2018-12-19
bug 1614567: Allow user to select Network Port for provisioning

sseago:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1614567

Allow user to select Network Port or Cloud Network when provisioning an OpenStack cloud instance.

Related to 
https://github.com/ManageIQ/manageiq/pull/18303
https://github.com/ManageIQ/manageiq-ui-classic/pull/5103

### DIFF
--- a/content/miq_dialogs/miq_provision_openstack_dialogs_template.yaml
+++ b/content/miq_dialogs/miq_provision_openstack_dialogs_template.yaml
@@ -1,0 +1,409 @@
+---
+:name: miq_provision_openstack_dialogs_template
+:description: Sample Openstack Instance Provisioning Dialog
+:dialog_type: MiqProvisionWorkflow
+:content:
+  :buttons:
+  - :submit
+  - :cancel
+  :dialogs:
+    :requester:
+      :description: Request
+      :fields:
+        :owner_phone:
+          :description: Phone
+          :required: false
+          :display: :hide
+          :data_type: :string
+        :owner_country:
+          :description: Country/Region
+          :required: false
+          :display: :hide
+          :data_type: :string
+        :owner_phone_mobile:
+          :description: Mobile
+          :required: false
+          :display: :hide
+          :data_type: :string
+        :owner_title:
+          :description: Title
+          :required: false
+          :display: :hide
+          :data_type: :string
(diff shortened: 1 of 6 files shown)


sseago:
Regarding the codeclimate issue "Method validate_network_port has 5 arguments (exceeds 4 allowed)." -- this can't be changed because the args are an aspect of the UI/automate framework that this code works within

sseago:
Fixed the rubocop errors. Build/test failures are because the associated manageiq PR isn't pushed yet.

gmcculloug:
@sseago Since you moved the dialog, can you create separate commits for the dialog before the new fields (just the move of the original dialog to this repo) and then another commit showing the fields being added for this new feature.  Otherwise that detail is difficult to see.

The other files are all limited changes just for this feature.

content/miq_dialogs/miq_provision_openstack_dialogs_template.yaml
```diff
@@ -164,6 +164,15 @@
           :required: true
           :display: :edit
           :data_type: :integer
+        :cloud_network_selection_method:
+          :values:
+            network: Cloud Network
+            port: Network Port
+          :description: How to Specify Network
```
gmcculloug: @sseago Can you suggest an alternative description for this field?  It stood out to me when reviewing the screenshot in https://github.com/ManageIQ/manageiq-ui-classic/pull/5103 that it does not match the feel of any other field description.

Would this be better worded as: "Network Type" or "Network Selection Type"

sseago: @gmcculloug You're right -- it doesn't read very smoothly, although I think introducing "Type" into the label would make the wording worse, since it would imply that the selection would influence the actual object type of the resulting association. The wording there now is clear, but perhaps a bit clumsy. The back end param has been renamed cloud_network_selection_method -- maybe we could just use something based on this? I could change it to "Network Selection Method" -- i.e. with the choices of choose by "Cloud Network" vs choose by "Network Port", I think it would be as clear as the current wording, but less clumsy than the "How to..." verbiage.

PR #417, opened 2018-12-19
Get tenant for stack from parameters

alexander-demichev:
Get tenant for stack from parameters, not from service, because service seems to always be admin

@aufi @mansam 

https://bugzilla.redhat.com/show_bug.cgi?id=1597125

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
@@ -281,7 +281,13 @@ class ManageIQ::Providers::Openstack::Inventory::Parser::CloudManager < ManageIQ
       o.status_reason = stack.stack_status_reason
       o.parent = persister.orchestration_stacks.lazy_find(stack.parent)
       o.orchestration_template = orchestration_template(stack)
-      o.cloud_tenant = persister.cloud_tenants.lazy_find(stack.service.current_tenant["id"])
+      # stack parameters can miss tenant_id, so we make admin default tenant
+      tenant_id = if stack.parameters && stack.parameters["OS::project_id"]
+                    stack.parameters["OS::project_id"]
+                  else
+                    stack.service.current_tenant["id"]
+                  end
+      o.cloud_tenant = persister.cloud_tenants.lazy_find(tenant_id)
 
       orchestration_stack_resources(stack, o)
       orchestration_stack_outputs(stack, o)
--- a/spec/models/manageiq/providers/openstack/openstack_stubs.rb
+++ b/spec/models/manageiq/providers/openstack/openstack_stubs.rb
@@ -179,9 +179,8 @@ module OpenstackStubs
           'output_value' => "output_value_#{i}",
           'description'  => "output_description_#{i}"
         }],
-        :parameters          => [{
-          "parameter_#{i}" => "key_#{i}"
-        }],
+        :parameters          => {"OS::project_id" => "project_id_#{i}"},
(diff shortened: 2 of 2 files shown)


alexander-demichev:
@mansam Thanks, for review, I'll take a look at tests, just wanted to be sure, that this change is ok.

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -281,7 +281,11 @@ def orchestration_stacks
       o.status_reason = stack.stack_status_reason
       o.parent = persister.orchestration_stacks.lazy_find(stack.parent)
       o.orchestration_template = orchestration_template(stack)
-      o.cloud_tenant = persister.cloud_tenants.lazy_find(stack.service.current_tenant["id"])
```
aufi: I think we should keep the original tenant_id lookup.

URI parsing looks quite strange, it would be better write it more robust and use just in case stack.service.current_tenant["id"] doesn't work.

alexander-demichev:
Some more info:
I decided to use url for getting tenants because sometimes stack can miss tenant id in params. Like this: 
```
[1] pry(#<ManageIQ::Providers::Openstack::Inventory::Parser::CloudManager>)> stack
=>   <Fog::Orchestration::OpenStack::Stack
    id="091e1e54-e01c-4ec5-a0ab-b00bee4d425c",
    capabilities=nil,
    description="Heat WordPress template to ...",
    disable_rollback=nil,
    links=[{"href"=>"http://42.42.42.42:4242/v1/69f8f7205ade4aa59084c32c83e60b5a/stacks/stack1/stack_id", "rel"=>"self"}],
    notification_topics=nil,
    outputs=nil,
    parameters=nil,
    stack_name="stack1",
    stack_status="CREATE_COMPLETE",
    stack_status_reason="Stack CREATE completed successfully",
    template_description=nil,
    timeout_mins=nil,
    parent=nil,
    creation_time="2016-11-11T13:50:09",
    updated_time=nil,
    stack_user_project_id="8aa9524c3bfa41669b9e0837fc091e22",
    stack_owner=nil,
    files=nil
  >
```

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
mansam: Agreed, but I'd prefer to see `OS::project_id` checked first, and then if that's unavailable fall back to something else. `stack.service.current_tenant["id"]` is afaik always going to be the tenant used to connect to Heat, so that will usually not be correct.

mansam: @alexander-demichev does the url in `links` actually have the stack's project_id, or is it just the project_id of the user who made the API request?

alexander-demichev: @mansam I need to check how url is generated, but can we handle empty parameters other way?

mansam: @alexander-demichev I think the ultimate fallback if the project id can't be found would be to just use `stack.service.current_tenant["id"]` so that it's assigned to the admin.

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -281,7 +281,13 @@ def orchestration_stacks
...
+      # stack parameters can miss tenant_id, so we make admin default tenant
+      tenant_id = if stack.parameters
```
mansam: You should do something roughly like `if stack.parameters && stack.parameters["OS::project_id"]`, because this current conditional will return nil if `stack` has a value for `parameters`, but doesn't have a value for `OS::project_id`

PR #420, opened 2018-12-19
Don't collect volume attachments for missing servers

mansam:
This PR causes the creation of disks for volume attachments to be skipped when the instance the attachments refer to can't be found in the db. This is intended to help fix 1651340, where refresh of a migrated server fails with a referential integrity error while saving volume attachments, because `lazy_find` on a server id ultimately resolved to nil.

https://bugzilla.redhat.com/show_bug.cgi?id=1651340

@Ladas with the change to using `find`, the vcr specs fail because they get the wrong count of disks. (presumably the servers turn up later which lazy_find takes care of, but since find is immediate some get left out?). Changing the inventory strategy solves that problem, but I don't have much more than an intuition about why that is. Any thoughts on whether this makes sense?

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/parser/storage_manager/cinder_manager.rb
+++ b/app/models/manageiq/providers/openstack/inventory/parser/storage_manager/cinder_manager.rb
@@ -82,9 +82,12 @@ class ManageIQ::Providers::Openstack::Inventory::Parser::StorageManager::CinderM
 
       dev = File.basename(a['device'])
 
+      server = persister.vms.find(a["server_id"])
+      next unless server
+
       attachment_names = {'vda' => 'Root disk', 'vdb' => 'Ephemeral disk', 'vdc' => 'Swap disk'}
       persister.disks.find_or_build_by(
-        :hardware    => persister.hardwares.lazy_find(persister.vms.lazy_find(a["server_id"])),
+        :hardware    => persister.hardwares.lazy_find(server),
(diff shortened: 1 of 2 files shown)


Ladas:
@mansam hm this is strange, the current behavior should be the same. It should just throw warning in production mode and skip the saving of disk. 

Hm no, damn we've released it with this TODO: @agrare https://github.com/ManageIQ/inventory_refresh/blob/v0.1.1/lib/inventory_refresh/save_collection/saver/base.rb#L276

So we should fix the core or merge this PR as a hotfix now.

mansam:
It seems like fixing this in the core would be ideal, so that the result of a lazy_find can be used in a find_or_build (you don't know in advance whether lazy_find will find anything, so the inventory save code has to take care of that situation).

Ladas:
@mansam right, so for the Topology Inventory service, this is already solved by precreating it on the other end. But this requires unique indexes in the DB, so I can't add it to MIQ. 

So e.g. vms.lazy_find(ems_ref), will actually create a VM with just the ems_ref (and other basic info like ems_id), if it doesn't find it. Then some other event comes in and updates the Vm fully. And it's also save to run in parallel, since it does atomic upsert.

But for current MIQ, we will just ignore saving the disk record, if the core is fixed.

agrare:
@mansam https://github.com/ManageIQ/manageiq/pull/18328 should handle running those checks only in debug/test environments.

mansam:
@agrare Wonderful, thanks. Going to close this.

PR #421, opened 2019-01-03
Remove legacy network collection from CloudManager graph refresh

mansam:
This PR removes obsolete network collection code that was ported over from the [legacy refresh parser](https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/cloud_manager/refresh_parser.rb#L348) and which appears to predate CloudNetworks/Ports as a way to model Openstack networking.

This code, besides being obsolete, appears to be very slow since it uses [attributes on the fog-openstack Server model](https://github.com/fog/fog-openstack/blob/master/lib/fog/openstack/compute/models/server.rb#L169) that proxy into extremely inefficient calls to the deprecated Nova `os-floating-ip` API that are repeated per VM.


aufi:
Looks good to me, merging!

simaishi:
@mansam @aufi Can this be `hammer/yes` and `gaprindashvili/yes`?

mansam:
@miq-bot add_label hammer/yes

simaishi:
Adding `hammer/conflict` and `gaprindashvili/conflict` as separate PRs are needed.

simaishi:
Backported to Hammer via https://github.com/ManageIQ/manageiq-providers-openstack/pull/425

Backported to Gaprindashvili via https://github.com/ManageIQ/manageiq-providers-openstack/pull/424

PR #422, opened 2019-01-04
 Targeted refresh: Collect tenant ems references last

mansam:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1662126 by ensuring that the reference cache is refreshed before collecting tenants during a targeted refresh.

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
@@ -122,7 +122,6 @@ class ManageIQ::Providers::Openstack::Inventory::Collector::TargetCollection < M
 
   def tenants
     return [] if references(:cloud_tenants).blank?
-    return @tenants if @tenants.any?
     @tenants = references(:cloud_tenants).collect do |cloud_tenant_id|
       memoized_get_tenant(cloud_tenant_id)
     end.compact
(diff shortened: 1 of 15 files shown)


app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
```diff
@@ -278,6 +274,13 @@ def infer_related_ems_refs!
       infer_related_cloud_volumes_ems_refs_db!
       infer_related_cloud_volumes_ems_refs_api!
     end
+    unless references(:cloud_tenants).blank?
+      # need to reset the association cache so that tenants added by the
+      # previous infer methods get picked up
+      target.manager_refs_by_association_reset
+      infer_related_cloud_tenant_ems_refs_db!
+      infer_related_cloud_tenant_ems_refs_api!
```
Ladas: this is probably still not enough, since the infer_related_cloud_tenant_ems_refs_api! still ads to ```references(:cloud_tenants)```, so whatever is added by that would not be fetched by the API

i think it might be enough to just cache the API call here? Or do we call the ```tenants``` method often?

mansam: `infer_related_cloud_tenant_ems_refs_api!` is where `tenants` is used for the first time, so I think I should just need to move `target.manager_refs_by_association_reset` into there.

Ladas: right, but we'll need to also fetch the tenants again, since we are caching them here https://github.com/ManageIQ/manageiq-providers-openstack/blob/ac5abeab9386bb2cdcdfbf8e7d0ba9af0a09fb0e/app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb#L125

Ladas: so we should do both ```target.manager_refs_by_association_reset``` and ```@tenants = nil```, or maybe do a non cached call

Ladas: yes, I think we don't call the ```tenants``` much right? So it should be enough that the API call is cached

Ladas:
:+1:

simaishi:
@mansam This is needed in `hammer` branch, but spec files will conflict on backport. Please create a separate PR.

simaishi:
Backported to Hammer via https://github.com/ManageIQ/manageiq-providers-openstack/pull/427

PR #436, opened 2019-01-25
Bug 1664888: Warn, but still allow infra without ironic

sseago:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1664888

This commit changes the validation to warn but not fail for
infra providers without Ironic. In addition, refresh no longer
fails if ironic is missing -- it just logs a warning instead.


sseago:
UI validation warning display requires https://github.com/ManageIQ/manageiq-ui-classic/pull/5200 to show it, but there should be no build/run dependencies between the PRs

mansam:
Looks good to me, merging.

PR #445, opened 2019-03-04
[WIP] Collect Cinder AZs separately from Nova AZs

mansam:
This PR would cause Cinder availability zones to be split from Nova availability zones when using the modern inventory collection system. Currently both kinds of AZs are collected together and then merged under the CloudManager, which works when the AZs coincidentally share names but is broken the rest of the time, since they are not interchangeable. Splitting up the two kinds of AZs would allow the Create Volume form to list only the Cinder AZs instead of having to list a mix of valid and invalid options.

@Ladas or @aufi, do you have any thoughts about how I'm going about this, or know anything about why these were originally merged in the first place?

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager.rb
@@ -407,6 +407,10 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
     ManageIQ::Providers::Openstack::CloudManager::EventCatcher
   end
 
+  def volume_availability_zones
+    cinder_manager.availability_zones
+  end
+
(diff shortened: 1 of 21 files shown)


aufi:
Looks good even it is WIP :+1:

Ladas:
@mansam so I think they were considered as being the same thing. And I think we are actually counting on this in the UI. Since AZ shows just under the cloud. Might be easier to add a flags to AZ, so we can filter out just the storage, or the compute ones?

mansam:
@Ladas Adding a column to the AZ model does seem like a simpler approach.

mansam:
I've come up with a different way to do this with a "services" column that I like better. Going to close for now.

PR #447, opened 2019-03-06
Tag AvailabilityZones with the service(s) they belong to so that Cinder and Nova AZs can be identified

mansam:
@Ladas @aufi Based on our previous conversation, this PR uses a `services` column to indicate which Openstack services know about an AZ with that name. This seems like a much simpler way to go, and if this looks good I'll put up the other PRs it depends on.

https://bugzilla.redhat.com/show_bug.cgi?id=1549128

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
@@ -45,14 +45,23 @@ class ManageIQ::Providers::Openstack::Inventory::Parser::CloudManager < ManageIQ
 
   def availability_zones
     collector.availability_zones.each do |az|
-      availability_zone = persister.availability_zones.find_or_build(az.zoneName)
-      availability_zone.ems_ref = az.zoneName
-      availability_zone.name = az.zoneName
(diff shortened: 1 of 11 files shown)


app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -45,14 +45,23 @@ def volume_snapshot_templates
...
   def availability_zones
     collector.availability_zones.each do |az|
-      availability_zone = persister.availability_zones.find_or_build(az.zoneName)
-      availability_zone.ems_ref = az.zoneName
-      availability_zone.name = az.zoneName
+      availability_zone = persister.availability_zones.find_or_build(az)
+      availability_zone.ems_ref = az
+      availability_zone.name = az
+      availability_zone.services = []
+      if collector.availability_zones_compute.include?(az)
```
Ladas: probably turn the `collector.availability_zones` to_set, so the lookup is O(1)

mansam:
Depends on https://github.com/ManageIQ/manageiq-schema/pull/346

PR #448, opened 2019-03-14
Support Cinder's volume-extend action

mansam:
There's an RFE for supporting Cinder's volume extend action. I don't know if it'll be accepted, but I had time so why not be proactive about it. https://bugzilla.redhat.com/show_bug.cgi?id=1685063

Volume extension has to be done via a separate request from updating other fields. I've added it as part of `raw_update_volume` since the alternative would be having to add a new UI just for this purpose. The way this is set up, if updating the volume's attributes fail, the whole thing will fail. But if only extending fails, the rest of the update will succeed but you'll still get a message in the UI about the resize having failed. If that seems odd, we could probably set up unique notifications and exceptions for volume-extend so that you get a notification for the attribute update and a separate notification for the results of the extend.

@aufi, what do you think about that?


mansam:
Depends on https://github.com/ManageIQ/manageiq/pull/18560

aufi:
Looks good to me, Sam :+1:

Issue #455, opened 2019-03-26
Malformed error message on validation failure


djberg96:
When validation failed for an openstack provider, I got a rather gnarly error message:

![image](https://user-images.githubusercontent.com/78529/55015608-89c33700-4fc3-11e9-850e-3223273282a2.png)

Originally posted at https://github.com/ManageIQ/manageiq-ui-classic/issues/5320

I think specifically it's when you try to enable SSL but the server doesn't support, though I've not tried all the combinations.

agrare:
@djberg96 can you find a backtrace from the logs?

mansam:
That looks like a 404 page from the web server due to trying to hit the SSL endpoint that doesn't exist. You could get a message like that from putting in any number of hosts with listening services that aren't Openstack services. I'll look into cleaning up how we deal with HTTP errors that we don't expect.

djberg96:
Fixed via https://github.com/ManageIQ/manageiq-providers-openstack/pull/456. Thanks @mansam!

PR #456, opened 2019-03-26
Add case for 404 error to ManagerMixin:translate_exception

mansam:
image

Fixes the ugly 404 error demonstrated in https://github.com/ManageIQ/manageiq-providers-openstack/issues/455


aufi:
Thanks for the PR, looks good to me :+1:

djberg96:
👍

PR #473, opened 2019-05-24
Manage snapshots as part of the inventory refresh process

mansam:
Instance "snapshots" in Openstack are a collection of an image
and a set of volume snapshots. In Horizon these are managed
individually, but ManageIQ abstracts them into a "snapshot"
object with its own UI. These snapshots were not part of the
inventory refresh process and could get out of sync with
the Glance images they represented. This patch causes these
helper objects to be managed by the inventory refresh process.

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
@@ -205,7 +205,29 @@ class ManageIQ::Providers::Openstack::Inventory::Parser::CloudManager < ManageIQ
       operating_system.product_name = guest_os
       operating_system.distribution = i.try(:os_distro)
       operating_system.version = i.try(:os_version)
+
+      if snapshot?(i) && parent_server_uid
+        snapshot = persister.snapshots.find_or_build(i.id)
+        snapshot.name = i.name
+        snapshot.uid  = i.id
+        snapshot.uid_ems = i.id
+        snapshot.ems_ref = i.id
+        snapshot.create_time = i.created_at
+        snapshot.description = i.attributes[:description]
+        snapshot.vm_or_template = persister.vms.lazy_find(parent_server_uid)
+      end
     end
+  end
+
+  def snapshot?(image)
+    return true if image.image_type == 'snapshot'
+
+    block_device_mapping = image.attributes[:block_device_mapping]
+    source_type = JSON.parse(block_device_mapping)&.dig('source_type') if block_device_mapping
+
+    source_type == 'snapshot'
+  rescue JSON::ParserError
+    false
   end
 
   def orchestration_stack_resources(stack, stack_inventory_object)
--- a/app/models/manageiq/providers/openstack/inventory/persister/definitions/cloud_collections.rb
+++ b/app/models/manageiq/providers/openstack/inventory/persister/definitions/cloud_collections.rb
@@ -45,6 +45,8 @@ module ManageIQ::Providers::Openstack::Inventory::Persister::Definitions::CloudC
 
     add_orchestration_stack_ancestry
 
+    add_snapshots
+
     add_vm_and_template_labels
     add_vm_and_template_taggings
   end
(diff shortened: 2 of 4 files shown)


mansam:
https://bugzilla.redhat.com/show_bug.cgi?id=1690954

aufi:
Looks as right approach to me :+1:

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -205,9 +205,29 @@ def miq_templates
...
+        snapshot.vm_or_template = persister.vms.lazy_find(parent_server_uid)
+      end
     end
   end
 
+  def snapshot?(image)
+    block_device_mapping = image.attributes[:block_device_mapping]
+    source_type = JSON.parse(block_device_mapping).try(:first).try(:[], 'source_type') if block_device_mapping
+
+    (image.image_type == 'snapshot') || (source_type == 'snapshot')
```
agrare: How about `return true if image.image_type == 'snapshot'` early to avoid having to  `JSON.parse()` if the first condition of your `||` is true?

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -205,9 +205,29 @@ def miq_templates
...
+        snapshot.create_time = i.created_at
+        snapshot.description = i.attributes[:description]
+        snapshot.vm_or_template = persister.vms.lazy_find(parent_server_uid)
+      end
     end
   end
 
+  def snapshot?(image)
+    block_device_mapping = image.attributes[:block_device_mapping]
+    source_type = JSON.parse(block_device_mapping).try(:first).try(:[], 'source_type') if block_device_mapping
```
agrare: Is the `.try(:first)` necessary?  I don't think it is possible for `JSON.parse()` to not throw a parser error and also return nil.  Or is this if it returns something that doesn't respond_to `:first` ?

mansam: Good catch, that should be removed. It's an artifact of me originally using `JSON.load` there.

agrare:
@mansam :heart: this   much better than creating the record when creating the snapshot

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
agrare: Ah okay, `&.dig("source_type")` possibly a little cleaner, idk why but the `.try(:[], 'source_type')` always looked a little strange to me

mansam: Heh agreed, .try(:[], _) is a pretty weird construct.

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -205,9 +205,31 @@ def miq_templates
...
     end
   end
 
+  def snapshot?(image)
+    return true if image.image_type == 'snapshot'
+
+    block_device_mapping = image.attributes[:block_device_mapping]
+    source_type = JSON.parse(block_device_mapping)&.dig('source_type') if block_device_mapping
+
+    (source_type == 'snapshot')
```
agrare: Minor but don't need the parens now

app/models/manageiq/providers/openstack/inventory/persister/definitions/cloud_collections.rb
```diff
@@ -144,6 +146,13 @@ def add_orchestration_stack_ancestry
     end
   end
 
+  def add_snapshots
```
agrare: We have this collection in the base InfraManager [ref](https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/inventory/persister/builder/infra_manager.rb#L51-L56) so we can just move this to shared [ref](https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/inventory/persister/builder/shared.rb) so you don't have to re-define it.

Trying to get all of the provider persisters as close to core as possible

agrare: I should say I'm good with this for now just an opportunity to clean things up in a follow-up pr

aufi:
Looks great, merging :+1:

PR #480, opened 2019-06-28
The leaf class should be tested

d-m-u:
Found here: https://github.com/ManageIQ/manageiq-providers-amazon/pull/542

One of these specs is pointless now since the leaf will always create three. Unless Adam disagrees. I'm sure I'll hear about it if he does. 

Broken originally by https://github.com/ManageIQ/manageiq/pull/18842

### DIFF
--- a/spec/models/manageiq/providers/openstack/cloud_manager/refresh_worker_spec.rb
+++ b/spec/models/manageiq/providers/openstack/cloud_manager/refresh_worker_spec.rb
@@ -1,28 +1,11 @@
 describe ManageIQ::Providers::Openstack::CloudManager::RefreshWorker do
   context "EMS with children" do
-    let!(:network_manager) { FactoryBot.create(:ems_network) }
-    let!(:storage_manager) { FactoryBot.create(:ems_storage) }
-    let(:ems) do
-      FactoryBot.create(:ems_cloud).tap do |ems|
-        network_manager.update_attributes(:parent_ems_id => ems.id)
-        storage_manager.update_attributes(:parent_ems_id => ems.id)
-      end
-    end
+    let(:ems) { FactoryBot.create(:ems_openstack) }
 
     it ".queue_name_for_ems" do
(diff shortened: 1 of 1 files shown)


d-m-u:
@agrare

d-m-u:
@miq-bot add_label test

d-m-u:
@agrare you okay with the removal of the superfluous spec here?

d-m-u:
I'm sure it's gonna break. I'm not sure it's gonna break for reason I know anything about. I'm not at all sure it's gonna break for anything this time that has anything to do with me. One can hope.

d-m-u:
I don't think anything that failed is related to this, or ems changes done last week at all.

agrare:
Going to merge this red since it fixes some of the failing tests

spec/models/manageiq/providers/openstack/cloud_manager/refresh_worker_spec.rb
```diff
@@ -1,28 +1,11 @@
...
-        network_manager.update_attributes(:parent_ems_id => ems.id)
-        storage_manager.update_attributes(:parent_ems_id => ems.id)
-      end
-    end
+    let(:ems) { FactoryBot.create(:ems_openstack) }
 
     it ".queue_name_for_ems" do
       queue_name = described_class.queue_name_for_ems(ems)
-      expect(queue_name.count).to eq(3)
+      expect(queue_name.count).to eq(4)
```
aufi: I'm looking for good reason why this should be 4 (Cloud+Network+Storage is 3)..

d-m-u: Because we should be testing these the way we use them. Since this is now a leaf class with the correct inheritance set by the model, the counts all increase by one to reflect the correct inheritance. It makes no sense to use an intermediate class here since we would never actually do that in the product.

PR #487, opened 2019-07-24
[WIP] Fixing targeted refresh for nested stacks and their resources

Ladas:
First the stack target was not extracted from the event, because
it goes as resource_id, not stack_id.

Second we need to always add all parent stacks as targets (so the
chain of relations gets connected) and we need to scan resources
to connect all instances to the stack.

(Issue found and fixed as part of hackaton)

WIP: 
- have to add other resources tied to stack (networks, security_groups, ...)

Fixes:
https://bugzilla.redhat.com/show_bug.cgi?id=1732861

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/collector.rb
+++ b/app/models/manageiq/providers/openstack/inventory/collector.rb
@@ -100,4 +100,21 @@ class ManageIQ::Providers::Openstack::Inventory::Collector < ManageIQ::Providers
   def orchestration_service
     @orchestration_service ||= manager.openstack_handle.detect_orchestration_service
   end
+
+  def all_orchestration_stacks
+    return [] unless orchestration_service
+    # TODO(lsmola) We need a support of GET /{tenant_id}/stacks/detail in FOG, it was implemented here
+    # https://review.openstack.org/#/c/35034/, but never documented in API reference, so right now we
+    # can't get list of detailed stacks in one API call.
+    return @all_orchestration_stacks unless @all_orchestration_stacks.nil?
+    @all_orchestration_stacks = if openstack_heat_global_admin?
+                                  orchestration_service.handled_list(:stacks, {:show_nested => true, :global_tenant => true}, true).collect(&:details)
+                                else
+                                  orchestration_service.handled_list(:stacks, :show_nested => true).collect(&:details)
+                                end
+  rescue Excon::Errors::Forbidden
+    # Orchestration service is detected but not open to the user
+    $log.warn("Skip refreshing stacks because the user cannot access the orchestration service")
+    []
+  end
 end
--- a/app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
@@ -118,10 +118,16 @@ class ManageIQ::Providers::Openstack::Inventory::Collector::TargetCollection < M
 
   def vms
     return [] if references(:vms).blank?
-    return @vms if @vms.any?
-    @vms = references(:vms).collect do |vm_id|
-      safe_get { compute_service.servers.get(vm_id) }
+    references(:vms).collect do |vm_id|
+      get_vm(vm_id)
     end.compact
+  end
+
+  def get_vm(uuid)
+    @indexes_vms ||= {}
+    return @indexes_vms[uuid] if @indexes_vms[uuid]
+
+    @indexes_vms[uuid] = safe_get { compute_service.servers.get(uuid) }
   end
 
   def flavors
@@ -319,7 +325,6 @@ class ManageIQ::Providers::Openstack::Inventory::Collector::TargetCollection < M
       orchestration_resources(stack).each do |resource|
         case resource.resource_type
         when "OS::Nova::Server"
-          # TODO(lsmola) we need to cache vms also on individual level, so the calling vms again has also these vms
           add_simple_target!(:vms, resource.physical_resource_id)
         end
       end
--- a/app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
@@ -94,8 +94,9 @@ class ManageIQ::Providers::Openstack::Inventory::Collector::TargetCollection < M
   def orchestration_stacks
     return [] unless orchestration_service
     return [] if targets_by_association(:orchestration_stacks).blank?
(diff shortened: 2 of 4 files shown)


app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
```diff
@@ -307,7 +314,26 @@ def infer_related_orchestration_stacks_ems_refs_db!
...
   def infer_related_orchestration_stacks_ems_refs_api!
     orchestration_stacks.each do |stack|
-      add_simple_target!(:orchestration_stacks, stack.parent, :tenant_id => stack.service.current_tenant["id"]) unless stack.parent.blank?
+      # Scan resources for VMs and add them as target, so the stack connects to vm, otherwise they don't connect on
+      # targeted refresh
+      orchestration_resources(stack).each do |resource|
+        case resource.resource_type
+        when "OS::Nova::Server"
+          # TODO(lsmola) we need to cache vms also on individual level, so the calling vms again has also these vms
+          add_simple_target!(:vms, resource.physical_resource_id)
```
Ladas: @mansam where do we use the `tenant_id` that I am seeing here. Cause I see it missing for other collections

Ladas: @aufi ^

app/models/manageiq/providers/openstack/inventory/collector.rb
```diff
@@ -100,4 +100,21 @@ def volume_service
   def orchestration_service
     @orchestration_service ||= manager.openstack_handle.detect_orchestration_service
   end
+
+  def all_orchestration_stacks
+    return [] unless orchestration_service
+    # TODO(lsmola) We need a support of GET /{tenant_id}/stacks/detail in FOG, it was implemented here
+    # https://review.openstack.org/#/c/35034/, but never documented in API reference, so right now we
+    # can't get list of detailed stacks in one API call.
```
aufi: I tested Fog and also Heat API, and OpenStack still doesn't support list stacks with details. Did I missed something or we need add get_stack with details iterated queries?


https://github.com/openstack/heat/blob/master/heat/api/openstack/v1/stacks.py

Ladas: ah, this TODO is unrelated. It's just copied when I moved the method. (it was not implemented for 5 years or so? :-))

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
```diff
@@ -104,19 +105,31 @@ def orchestration_stacks
...
 
-  def get_orchestration_stack(stack_id, tenant_id)
-    tenant = memoized_get_tenant(tenant_id)
-    safe_get { @os_handle.detect_orchestration_service(tenant.try(:name)).stacks.get(stack_id) }
+  def indexed_all_orchestration_stacks
+    @indexed_all_orchestration_stacks ||= all_orchestration_stacks.index_by(&:id)
+  end
+
+  def get_orchestration_stack(stack_id, _tenant_id = nil)
+    # TODO fog needs to implement /v1/{tenant_id}/stacks/{stack_identity} call, right now the only supported call
```
Ladas: @aufi this is the TODO you should focus on. So we are able to fetch orchestration stack with :ems_ref and it searches in nested stack too. So adding `/v1/{tenant_id}/stacks/{stack_identity} ` to fog

Ladas: the workaround now is that we need to fetch all nested orchestration stacks, instead of fetching the specific ones involved in targeted refresh. So this one call is making targeted refresh much slower (in reality, it would be quicker if we have e.g. 1000 stack targets, which is logic I have implemented in Azure, doing list vs. get based on number of targets)

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
```diff
@@ -104,19 +105,31 @@ def orchestration_stacks
     []
   end
 
-  def get_orchestration_stack(stack_id, tenant_id)
-    tenant = memoized_get_tenant(tenant_id)
-    safe_get { @os_handle.detect_orchestration_service(tenant.try(:name)).stacks.get(stack_id) }
```
Ladas: @aufi it looks like this wasn't fetching stack `details`, so I wonder if this worked correctly (fetching all the needed data). 
So check if `/v1/{tenant_id}/stacks/{stack_identity}` gives you the detail, or we need to do 2 calls

app/models/manageiq/providers/openstack/inventory/collector.rb
Ladas: you can move this method back to `app/models/manageiq/providers/openstack/inventory/collector/cloud_manager.rb` once it is not used in targeted refresh

aufi: Right, it has not been implemented since :-)

Okay, thanks.

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
aufi: Got it, thanks!

app/models/manageiq/providers/openstack/inventory/collector/target_collection.rb
mansam: @Ladas It's used here: https://github.com/ManageIQ/manageiq-providers-openstack/blob/3209e0ffff8f60d9f311d58774f07827dd4cb0c1/app/models/manageiq/providers/openstack/cloud_manager/event_target_parser.rb#L68

If I remember correctly, with other collections, you can get the tenant from the event payload. I couldn't get the tenant from the event payload for stacks, so I did it this way,

Ladas: @mansam ok, can you check if you can add it? Also I've seen it being nil for some other targets, can you check that out with the OpenStack env I've send to @aufi ?

(feel free to just take this PR and continue :-))

andyvesel:
@Ladas @aufi Some relevant BZ https://bugzilla.redhat.com/show_bug.cgi?id=1728624 has been tested with this PR and it works fine. Can we merge the PR it its current state?

aufi:
At least CI should pass and PR should not be WIP. It would be nice have some tests too, but I'd rely on Ladas view.

Ladas:
@aufi @andyvesel yeah, you will need to re-record the VCRs. Then this can be merged. (not sure if you fixed the ineffective loading of stacks?)

PR #500, opened 2019-08-08
Collect all security groups in response to a security group event.

mansam:
Security group events from Panko don't include the ID of the security group they refer to, so targeted refresh can't work. This patch causes a security group event to kick off a refresh of the entire security group inventory.


aufi:
LGTM :+1:

PR #502, opened 2019-08-21
Set :complete=>false when doing targeted snapshot refresh

mansam:
I noticed that targeted refresh of snapshots seems to wipe out all the snapshots except for the target, suggesting that the inventory system is treating this as a complete refresh. Manually setting the :complete flag in the collection definition solves the problem, but I'm puzzled why this is necessary here but not for other collections.

@agrare @Ladas any thoughts on why this would be needed for just this collection? I hunted around through the inventory_refresh repo but didn't discover any clues.

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/persister/definitions/cloud_collections.rb
+++ b/app/models/manageiq/providers/openstack/inventory/persister/definitions/cloud_collections.rb
@@ -150,6 +150,7 @@ module ManageIQ::Providers::Openstack::Inventory::Persister::Definitions::CloudC
     add_collection(cloud, :snapshots) do |builder|
       builder.add_properties(:model_class => ::Snapshot)
       builder.add_properties(:parent_inventory_collections => %i[vms miq_templates])
+      builder.add_properties(:complete => !targeted?)
     end
   end
 


app/models/manageiq/providers/openstack/inventory/persister/definitions/cloud_collections.rb
```diff
@@ -150,6 +150,7 @@ def add_snapshots
     add_collection(cloud, :snapshots) do |builder|
       builder.add_properties(:model_class => ::Snapshot)
       builder.add_properties(:parent_inventory_collections => %i[vms miq_templates])
```
Ladas: @mansam so we might have a bad definition here. The parent_inventory collection says the snapshots are owned by VMs and images and we can't really do targeted refresh of those in MIQ. 

It's only possible to do a targeted refresh of Vm, which expects you will send all snapshots there, otherwise it will delete the snapshots you didn't send. (the query is fetching all snapshots of targeted VMs)

So the snapshot should really have ems_id and not be owned by a Vm or image (after all snapshot can live without VM or image), then you can ditch the parent_inventory_collections and targeted refresh should work as expected.

In the meantime, the `:complete => false` will prevent Vm targeted refresh to delete all snapshots that are not being sent for refresh, so I think it's a good workaround for now.

mansam: @Ladas Ahh, gotcha. Thanks for the thorough explanation.

PR #507, opened 2019-09-11
Completely remove gettext catalogs

mzazrivec:
* gettext catalogs are no longer needed
* no need for the plugin to initialize its own catalogs
* remove `zanata.xml`
* remove i18n placeholders test


aufi:
LGTM :+1:

PR #513, opened 2019-09-27
Specify CloudManager::Template in Cloud persister definition

mansam:
Since the Cloud persister miq_template collection definition was using the 'MiqTemplate' model, all MiqTemplates were getting scooped up when attempting to reconnect archived templates. This PR specifies that only objects of the CloudManager's own Template type should be included.

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/base_template.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/base_template.rb
@@ -1,0 +1,2 @@
+class ManageIQ::Providers::Openstack::CloudManager::BaseTemplate < ManageIQ::Providers::CloudManager::Template
+end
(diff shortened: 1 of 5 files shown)


mansam:
@aufi I was able to confirm that creating an Infra provider, removing it, and then adding the same Openstack as a cloud provider triggers the bug we saw w/ infra templates. This fix seems to resolve the issue for me.

agrare:
Do you even need to specify a model class? This should get built automatically

mansam:
@agrare It seems like I do, archived Infra templates seem to still get pulled in if I don't specify the model class.

agrare:
Hm okay the persister builder should set the model class to what you have there automatically if you don't pass anything.  Can you debug it and see what it does set it to if you don't pass anything?

agrare:
Oh it might be because the association is miq_templates and the class name is Template

mansam:
I'll double check what it's setting. Should we be using a different association other than `miq_templates` for this collection?

agrare:
If we alias :templates to :miq_templates it might  just work but I'm good with this in the meantime

mansam:
Added a CloudManager::BaseTemplate class in the last commit which Template, VolumeTemplate, and VolumeSnapshotTemplate can all inherit from to tighten this up a bit since they all previously inherited from the base ManageIQ::Providers::CloudManager::Template. Should prevent any weird stuff with kidnapping archived templates from other CloudProviders due to using that base class in the persister definition.

PR #514, opened 2019-10-02
Raise notifications when creating, restoring, and deleting backups

mansam:
I expect this to be a partial fix (along with associated core PR) for https://bugzilla.redhat.com/show_bug.cgi?id=1748918, where a lack of notifications around Volume Backup actions is obscuring the reason for backup deletion failing.

Depends on https://github.com/ManageIQ/manageiq/pull/19357


aufi:
Travis fails on small issue with test matchers.

```
Using the `raise_error` matcher without providing a specific
                   error or message risks false positives, since `raise_error`
                   will match when Ruby raises a `NoMethodError`, `NameError` or
                   `ArgumentError`, potentially allowing the expectation to pass
                   without even executing the method you are intending to call.
     
                   Instead, provide a specific error class or message.
```

mansam:
Thanks, will fix

aufi:
Thanks :+1: 
LGTM, merging.

PR #517, opened 2019-10-18
Skip Volumes without mountpoint in VM refresh

aufi:
If volume is mounted to a VM, but it doesn't have mountpoint to the VM, something
is wrong in OpenStack (e.g. failures with limits in migration), but our code should not fail
the refresh.

Adding skip volumes without mountpoint to VM parsing refresh code.

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
@@ -413,6 +413,9 @@ class ManageIQ::Providers::Openstack::Inventory::Parser::CloudManager < ManageIQ
       # will take care of that.
       if s.attributes.fetch("os-extended-volumes:volumes_attached", []).length > 0
         s.volume_attachments.each do |attachment|
+          # Skip Volume mounts without mount point
+          next if attachment['device'].blank?
+
           dev = File.basename(attachment['device'])
           persister.disks.find_or_build_by(
             :hardware    => hardware,


app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -413,6 +413,9 @@ def vms
       # will take care of that.
       if s.attributes.fetch("os-extended-volumes:volumes_attached", []).length > 0
         s.volume_attachments.each do |attachment|
+          # Skip Volume mounts without mount point
+          next unless attachment['device']
```
mansam: will `attachment['device']` be nil if the mountpoint is missing, or can it contain the empty string? Wondering if it might be better to test with `.blank?

aufi: Good point, updated, thanks :+1:

simaishi:
@aufi @mansam can this be `ivanchuk/yes`?

simaishi:
@agrare ^

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 8f981d72ecda5c9d43a48909c119f0c9a774bcdf
Author: Samuel Lucidi <slucidi@redhat.com>
Date:   Mon Oct 21 09:57:30 2019 -0400

    Merge pull request #517 from aufi/skip_vm_volumes_wo_mountpoint

    Skip Volumes without mountpoint in VM refresh

    (cherry picked from commit 96a1aead04b721a5cdf9854f552b696fe236ba37)

    https://bugzilla.redhat.com/show_bug.cgi?id=1848265
```

PR #518, opened 2019-10-23
Renamed subclass from ManageIQ::Providers::Openstack::InfraManager::EmsCluster to ManageIQ::Providers::Openstack::InfraManager::Cluster

chessbyte:


agrare:
Need to update the RefreshParser to set the correct type [here](https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/infra_manager/refresh_parser.rb#L387)

chessbyte:
@agrare I think I addressed all your concerns.  Please double-check

PR #523, opened 2019-10-31
Add an exception handler to event monitoring

djberg96:
This adds an exception handler to the `EventCatcherMixin#monitor_events` method. If it fails, it will properly log the issue instead of spewing a backtrace into the log. This occurred recently when connecting to an Openstack instance where Panko was missing.

Should it also call `stop_event_monitor` on an exception there? If you think so I can update the PR.


Fryguy:
Calling stop may also blow up.  I'm not sure what the right answer is for that.  reset ends up just setting it to nil, so technically it would still be running until the GC comes along.

agrare:
@aufi if we hit this exception would calling `.stop` on the event monitor handle be needed?

aufi:
I don't think calling ```.stop``` is needed, LGTM, merging.

PR #524, opened 2019-11-01
Move raw_ vm operations to providers

agrare:


app/models/manageiq/providers/openstack/cloud_manager.rb
```diff
@@ -230,60 +230,6 @@ def allow_targeted_refresh?
   # Operations
   #
 
-  def vm_start(vm, _options = {})
```
gtanzillo: Are these methods moving to a common place now? I searched a couple of them and see that they exist in each provider repo

agrare: So the code in core has `VmOrTemplate#start` calling `run_command_via_parent` which essentially does `vm.ext_management_system.vm_start(vm)`.  All of the cloud providers including openstack overrode these vm operations methods however (e.g. https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/cloud_manager/vm/operations/power.rb) so the core `run_command_via_parent` was never called and thus these were never used.

PR #525, opened 2019-11-04
Add proxy support to Openstack

djberg96:
At the moment it doesn't appear that the Openstack cloud provider is honoring default or openstack-specific proxy settings. As far as I can tell they're completely ignored.

~~WIP for now until I figure out why it's failing.~~

Steps to reproduce:

* Create a simple proxy server locally.
* Add openstack cloud provider.
* Update proxy settings in the advanced settings screen to use your local proxy settings.
* Refresh the provider and/or update authentication.

A basic proxy you can use to test:
```
#!/usr/bin/env ruby

require 'webrick'
require 'webrick/httpproxy'

proxy = WEBrick::HTTPProxyServer.new(Port: 8080) # Adjust port as desired.

trap 'INT'  do proxy.shutdown end
trap 'TERM' do proxy.shutdown end

proxy.start
```

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1649903

Regarding the extensive nature of the PR itself, this was largely due to the heavy use of delegation in combination with the fog client that was already present when I got here. This required, as far as I can tell, the addition of a proxy setting to each delegate class individually, which the fog internals would then consume.

### DIFF
--- a/app/models/manageiq/providers/openstack/helper_methods.rb
+++ b/app/models/manageiq/providers/openstack/helper_methods.rb
@@ -1,8 +1,9 @@
 module ManageIQ::Providers::Openstack::HelperMethods
   extend ActiveSupport::Concern
 
+  # Return the openstack proxy, if any, as a string. Otherwise return nil.
   def openstack_proxy
-    VMDB::Util.http_proxy_uri(:openstack) || VMDB::Util.http_proxy_uri(:default)
+    ManageIQ::Providers::Openstack::CloudManager.http_proxy_uri&.to_s
   end
 
   def parse_error_message_from_fog_response(exception)
@@ -19,7 +20,7 @@ module ManageIQ::Providers::Openstack::HelperMethods
 
   module ClassMethods
     def openstack_proxy
-      VMDB::Util.http_proxy_uri(:openstack) || VMDB::Util.http_proxy_uri(:default)
+      ManageIQ::Providers::Openstack::CloudManager.http_proxy_uri&.to_s
     end
 
     def parse_error_message_from_fog_response(exception)
--- a/app/models/manageiq/providers/openstack/helper_methods.rb
+++ b/app/models/manageiq/providers/openstack/helper_methods.rb
@@ -1,6 +1,10 @@
 module ManageIQ::Providers::Openstack::HelperMethods
   extend ActiveSupport::Concern
 
+  def openstack_proxy
+    VMDB::Util.http_proxy_uri(:openstack) || VMDB::Util.http_proxy_uri(:default)
+  end
+
   def parse_error_message_from_fog_response(exception)
     self.class.parse_error_message_from_fog_response(exception)
   end
--- a/app/models/manageiq/providers/openstack/helper_methods.rb
+++ b/app/models/manageiq/providers/openstack/helper_methods.rb
@@ -18,6 +18,10 @@ module ManageIQ::Providers::Openstack::HelperMethods
   end
 
   module ClassMethods
+    def openstack_proxy
+      VMDB::Util.http_proxy_uri(:openstack) || VMDB::Util.http_proxy_uri(:default)
+    end
+
     def parse_error_message_from_fog_response(exception)
       exception_string = exception.to_s
       matched_message = exception_string.match(/message\\\": \\\"(.*)\\\", /)
--- a/app/models/manageiq/providers/openstack/manager_mixin.rb
+++ b/app/models/manageiq/providers/openstack/manager_mixin.rb
@@ -115,7 +115,7 @@ module ManageIQ::Providers::Openstack::ManagerMixin
     end
 
     def raw_connect(password, params, service = "Compute")
-      params[:proxy] = openstack_proxy
+      params[:proxy] = openstack_proxy if openstack_proxy
 
       case params[:event_stream_selection]
       when "amqp"
@@ -171,7 +171,7 @@ module ManageIQ::Providers::Openstack::ManagerMixin
       extra_options[:region]            = provider_region if provider_region.present?
       extra_options[:omit_default_port] = ::Settings.ems.ems_openstack.excon.omit_default_port
       extra_options[:read_timeout]      = ::Settings.ems.ems_openstack.excon.read_timeout
-      extra_options[:proxy]             = openstack_proxy
+      extra_options[:proxy]             = openstack_proxy if openstack_proxy
 
       osh = OpenstackHandle::Handle.new(username, password, address, port, api_version, security_protocol, extra_options)
       osh.connection_options = {:instrumentor => $fog_log}
--- a/app/models/manageiq/providers/openstack/manager_mixin.rb
+++ b/app/models/manageiq/providers/openstack/manager_mixin.rb
@@ -40,6 +40,7 @@ module ManageIQ::Providers::Openstack::ManagerMixin
       ems.api_version            = params[:api_version].strip
       ems.security_protocol      = params[:default_security_protocol].strip
       ems.keystone_v3_domain_id  = params[:keystone_v3_domain_id]
+      ems.http_proxy_uri         = parmas[:proxy].strip if params[:proxy]
 
       user, hostname, port = params[:default_userid], params[:default_hostname].strip, params[:default_api_port].try(:strip)
 
@@ -115,6 +116,9 @@ module ManageIQ::Providers::Openstack::ManagerMixin
     end
 
     def raw_connect(password, params, service = "Compute")
+      proxy = VMDB::Util.http_proxy_uri(:openstack) || VMDB::Util.http_proxy_uri(:default)
+      params[:proxy] = proxy.to_s if proxy
+
       case params[:event_stream_selection]
       when "amqp"
         amqp_available?(password, params)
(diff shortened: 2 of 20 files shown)


djberg96:
Interestingly, it seems to validate successfully if you don't specify an api-version, and let it use a default. I was able to successfully add a provider this way, though event collection is still failing.

djberg96:
The UI issue was a bit of a red herring, and the current changes work at least for verification. Now on to refresh...

aufi:
I think we need setup proxy for provider validation also on https://github.com/ManageIQ/manageiq-providers-openstack/pull/525/files#diff-6f9b687ce51fd31aa102fa9305ef2501R33

djberg96:
@aufi With the current PR I can see the proxy being passed to `OpenstackHandle::Handle#raw_connect` as expected, but I don't see `ManagerMixin#ems_connect?` (or amqp_connect, presumably) even being called.

djberg96:
@aufi Actually, I do see it called, but only with an OSP11 provider.

djberg96:
@aufi Alright, updated. Could use a little help with live testing if/when you have some time.

djberg96:
Ah shoot, seeing this failure:
```
[----] E, [2020-02-21T08:32:49.060764 #98876:3fd13f432e34] ERROR -- : MIQ(MiqQueue#deliver) Message id: [296], Error: [undefined method `http_proxy_uri=' for #<ManageIQ::Providers::Openstack::CloudManager:0x00007fa25eca4308>
Did you mean?  http_proxy_uri]
[----] E, [2020-02-21T08:32:49.060995 #98876:3fd13f432e34] ERROR -- : [NoMethodError]: undefined method `http_proxy_uri=' for #<ManageIQ::Providers::Openstack::CloudManager:0x00007fa25eca4308>
Did you mean?  http_proxy_uri  Method:[block (2 levels) in <class:LogProxy>]
[----] E, [2020-02-21T08:32:49.061067 #98876:3fd13f432e34] ERROR -- : /Users/dberger/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/activemodel-5.1.7/lib/active_model/attribute_methods.rb:432:in `method_missing'
/Users/dberger/Dev/manageiq-providers-openstack-djberg96/app/models/manageiq/providers/openstack/manager_mixin.rb:33:in `ems_connect?'
/Users/dberger/Dev/manageiq-providers-openstack-djberg96/app/models/manageiq/providers/openstack/manager_mixin.rb:195:in `raw_connect'
/Users/dberger/Dev/manageiq-djberg96/app/models/ext_management_system.rb:327:in `raw_connect?'
/Users/dberger/Dev/manageiq-djberg96/app/models/miq_queue.rb:480:in `block in dispatch_method'
/Users/dberger/.rbenv/versions/2.5.7/lib/ruby/2.5.0/timeout.rb:93:in `block in timeout'
/Users/dberger/.rbenv/versions/2.5.7/lib/ruby/2.5.0/timeout.rb:33:in `block in catch'
/Users/dberger/.rbenv/versions/2.5.7/lib/ruby/2.5.0/timeout.rb:33:in `catch'
/Users/dberger/.rbenv/versions/2.5.7/lib/ruby/2.5.0/timeout.rb:33:in `catch'
/Users/dberger/.rbenv/versions/2.5.7/lib/ruby/2.5.0/timeout.rb:108:in `timeout'
/Users/dberger/Dev/manageiq-djberg96/app/models/miq_queue.rb:478:in `dispatch_method'
/Users/dberger/Dev/manageiq-djberg96/app/models/miq_queue.rb:455:in `block in deliver'
/Users/dberger/Dev/manageiq-djberg96/app/models/user.rb:302:in `with_user_group'
/Users/dberger/Dev/manageiq-djberg96/app/models/miq_queue.rb:455:in `deliver'
/Users/dberger/Dev/manageiq-djberg96/app/models/miq_queue_worker_base/runner.rb:99:in `deliver_queue_message'
/Users/dberger/Dev/manageiq-djberg96/app/models/miq_queue_worker_base/runner.rb:123:in `deliver_message'
/Users/dberger/Dev/manageiq-djberg96/app/models/miq_queue_worker_base/runner.rb:142:in `block in do_work'
/Users/dberger/Dev/manageiq-djberg96/app/models/miq_queue_worker_base/runner.rb:137:in `loop'
/Users/dberger/Dev/manageiq-djberg96/app/models/miq_queue_worker_base/runner.rb:137:in `do_work'
/Users/dberger/Dev/manageiq-djberg96/app/models/miq_worker/runner.rb:256:in `block in do_work_loop'
/Users/dberger/Dev/manageiq-djberg96/app/models/miq_worker/runner.rb:253:in `loop'
/Users/dberger/Dev/manageiq-djberg96/app/models/miq_worker/runner.rb:253:in `do_work_loop'
/Users/dberger/Dev/manageiq-djberg96/app/models/miq_worker/runner.rb:113:in `run'
/Users/dberger/Dev/manageiq-djberg96/app/models/miq_worker/runner.rb:95:in `start'
/Users/dberger/Dev/manageiq-djberg96/lib/workers/bin/run_single_worker.rb:113:in `<main>'
```

djberg96:
Latest issue: https://gist.github.com/djberg96/9b118978a2c34f431afdb18c8e462e73

djberg96:
Latest issue was that I needed an explicit `Accept-Encoding` to deal with the fact that the proxy was setting the `Content-Length` incorrectly because it was (also incorrectly) claiming the `Content-Type` was "gzip". So, this code explicitly sets it if the proxy value is also set.

I've tested this against a V2 provider and it works fine. I've not been able to test it against a V3 provider, still looking for a dev environmet.

djberg96:
Tested against v3, worked fine.

Fryguy:
@djberg96 I'm seeing `openstack_proxy.to_s` over and over and wondering if the openstack_proxy method should just return a string directly...something like

```ruby
(VMDB::Util.http_proxy_uri(:openstack) || VMDB::Util.http_proxy_uri(:default))&.to_s
```

Fryguy:
@aufi @agrare Does this look good to you?

app/models/manageiq/providers/openstack/helper_methods.rb
```diff
@@ -1,6 +1,10 @@
 module ManageIQ::Providers::Openstack::HelperMethods
   extend ActiveSupport::Concern
 
+  def openstack_proxy
+    VMDB::Util.http_proxy_uri(:openstack) || VMDB::Util.http_proxy_uri(:default)
```
agrare, line 5: Isn't this basically just `ManageIQ::Providers::BaseManager#http_proxy_uri` ?  Are you not able to use that method where this is being called from?  (Not sure where `HelperMethods` is included)

lib/manageiq/providers/openstack/legacy/openstack_handle/handle.rb
```diff
@@ -135,6 +135,13 @@ def excon_options
       @excon_options[:omit_default_port] = @extra_options[:omit_default_port] unless
                                            @extra_options[:omit_default_port].blank?
       @excon_options[:read_timeout]      = @extra_options[:read_timeout] unless @extra_options[:read_timeout].blank?
+
+      # Beware of proxies that lie about gzip encoding and content length.
```
agrare, line 139: :laughing:

agrare:
Other than it seeming like we copied `VMDB::Util.http_proxy_uri(:openstack) || VMDB::Util.http_proxy_uri(:default)` a bunch of times, looks good to me

app/models/manageiq/providers/openstack/helper_methods.rb
agrare: I think you can use the one from BaseManager when this is called by the ManagerMixin, maybe not from the legacy/openstack_handle code though

djberg96: I guess I can use the singleton method if nothing else. As for all the `to_s` calls, I don't think Excon takes a URI object, but let me check.

djberg96:
@agrare, @Fryguy Trying to remember now, but I think the issue was that `ManageIQ::Providers::BaseManager.http_proxy_uri` bombs if neither the default nor provider-specific proxy are set:

```
# With no providers or proxy settings set:
irb(main):005:0> VMDB::Util.http_proxy_uri(:openstack) || VMDB::Util.http_proxy_uri(:default)
=> nil
irb(main):006:0> ManageIQ::Providers::BaseManager.http_proxy_uri
Traceback (most recent call last):
       14: from bin/rails:4:in `<main>'
       13: from bin/rails:4:in `require'
       12: from railties (5.2.4.4) lib/rails/commands.rb:18:in `<top (required)>'
       11: from railties (5.2.4.4) lib/rails/command.rb:46:in `invoke'
       10: from railties (5.2.4.4) lib/rails/command/base.rb:69:in `perform'
        9: from thor (1.0.1) lib/thor.rb:392:in `dispatch'
        8: from thor (1.0.1) lib/thor/invocation.rb:127:in `invoke_command'
        7: from thor (1.0.1) lib/thor/command.rb:27:in `run'
        6: from railties (5.2.4.4) lib/rails/commands/console/console_command.rb:96:in `perform'
        5: from railties (5.2.4.4) lib/rails/commands/console/console_command.rb:19:in `start'
        4: from railties (5.2.4.4) lib/rails/commands/console/console_command.rb:64:in `start'
        3: from (irb):6
        2: from app/models/manageiq/providers/base_manager.rb:33:in `http_proxy_uri'
        1: from lib/vmdb/util.rb:4:in `http_proxy_uri'
NoMethodError (undefined method `to_hash' for nil:NilClass)
```

agrare:
@djberg96 that is because you're calling it directly on `BaseManager`, 
```
>> ManageIQ::Providers::Openstack::CloudManager.http_proxy_uri
=> nil
```

djberg96:
@agrare, @Fryguy ok, used your suggestions and refactored the helper method to use `ManageIQ::Providers::Openstack::CloudManager.http_proxy_uri&.to_s` so that it returns nil if it isn't set, or a string if it is. This eliminated most of the redundancy that was bothering you I think.

Tested it against a local OSP 10 instance, seemed to work fine.

app/models/manageiq/providers/openstack/manager_mixin.rb
```diff
@@ -115,7 +115,7 @@ def verify_credentials(args)
     end
 
     def raw_connect(password, params, service = "Compute")
-      params[:proxy] = openstack_proxy.to_s if openstack_proxy
+      params[:proxy] = openstack_proxy
```
agrare, line 118: Travis failures look related, I think you might still need the `if openstack_proxy`

djberg96: Ah, so explicit nils are causing problems? I was afraid of that.

agrare: Just guessing because that looks like the major difference between the commits

agrare: Nice looks like that did the trick :+1:

agrare (approved):

PR #529, opened 2019-11-08
Subclass require_snapshot_for_scan? method

chessbyte:


app/models/manageiq/providers/openstack/cloud_manager/vm_or_template_shared/scanning.rb
```diff
@@ -0,0 +1,16 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    supports :smartstate_analysis do
+      feature_supported, reason = check_feature_support('smartstate_analysis')
+      unless feature_supported
+        unsupported_reason_add(:smartstate_analysis, reason)
+      end
+    end
+  end
```
agrare: @chessbyte if we're moving this to vm_or_template_shared can we remove it from the [vm](https://github.com/ManageIQ/manageiq-providers-openstack/blob/78b451bfcfeed466a0ea5f37cf1032d8a566834d/app/models/manageiq/providers/openstack/cloud_manager/vm.rb#L12-L17) and [miq_template](https://github.com/ManageIQ/manageiq-providers-openstack/blob/78b451bfcfeed466a0ea5f37cf1032d8a566834d/app/models/manageiq/providers/openstack/cloud_manager/template.rb#L7-L12) classes?

PR #532, opened 2019-11-13
Add params_for_create and verify_credentials

agrare:
https://github.com/ManageIQ/manageiq/issues/18818


agrare:
In a follow-up we need to enable additional information to be entered if the amqp event type is selected

agrare:
@gtanzillo PTAL

PR #536, opened 2019-11-14
Remove ems_ref_obj

agrare:
We are looking to remove ems_ref_obj since it was originally added to store the "full" VMware object reference

### DIFF
--- a/app/models/manageiq/providers/openstack/infra_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/openstack/infra_manager/refresh_parser.rb
@@ -203,9 +203,8 @@ module ManageIQ
         new_result = {
           :name                 => host_name,
           :type                 => 'ManageIQ::Providers::Openstack::InfraManager::Host',
-          :uid_ems              => uid,
+          :uid_ems              => host.instance_uuid,
           :ems_ref              => uid,
-          :ems_ref_obj          => host.instance_uuid,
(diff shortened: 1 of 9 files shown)


aufi:
:+1: for removing ```ems_ref_obj``` attribute (when CI will pass)

app/models/manageiq/providers/openstack/infra_manager/refresh_parser.rb
```diff
@@ -203,9 +203,8 @@ def parse_host(host, indexed_servers, indexed_resources, cloud_hosts_attributes)
         new_result = {
           :name                 => host_name,
           :type                 => 'ManageIQ::Providers::Openstack::InfraManager::Host',
-          :uid_ems              => uid,
+          :uid_ems              => host.instance_uuid,
```
Fryguy: Will this require a data migration?

agrare: We don't use uid_ems for save_hosts_inventory so the first refresh would update existing host records.

We could parse the ems_ref_obj yaml and update uid_ems in a migration though if you want

agrare: https://github.com/ManageIQ/manageiq-schema/pull/426/commits/d57f12e7d0df25484eed2ec2c29b3e3ed93be8c5

PR #538, opened 2019-11-15
Remove unused sub-manager refresh worker

carbonin:
Refresh of the network manager is handled by the parent manager
refresh worker

see also: https://github.com/ManageIQ/manageiq/pull/19524


aufi:
If the parent manager is Cloud manager, we should also update settings and events filtering to allow network-related events to Cloud
https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/config/settings.yml#L173-L184

agrare:
@aufi yes unless the child-managers use a completely different API then I think most workers that currently run on a child-manager could be combined with the parent.  (Event and Metrics capture workers both come to mind)

PR #542, opened 2019-12-03
Add SecurityGroupRules to Targeted refresh

aufi:
Updating Security Groups Targeted Refresh to be triggered by change of
Security Group Rule which is a child object of Security Group.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1776365


aufi:
Depends on https://github.com/ManageIQ/manageiq-content/pull/612

aufi:
@mansam Hi, current version of the PR doesn't trigger refresh, could you check what I've missed?

mansam:
@aufi I believe you'll need to add entries to the automate domain in order for those events to kick off targeted refresh. Examples here: https://github.com/ManageIQ/manageiq-content/tree/master/content/automate/ManageIQ/System/Event/EmsEvent/OpenStack.class

aufi:
@mansam Right, that's part of https://github.com/ManageIQ/manageiq-content/pull/612

mansam:
Sorry, missed that you already had submitted that!

aufi:
@mansam I reached point where refresh is triggered and mostly works, but it duplicates existing Security Group Rules on a Security Group. It looks to me that it is caused by persister collection setup. Few question to refresh generally:

Do we know what is https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/inventory/persister/definitions/network_collections.rb#L55 used for?

Following line looks to be connected to Security Groups refresh. Should be a similar line in add_firewall_rules method? 
https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/inventory/persister/definitions/network_collections.rb#L98

mansam:
@aufi If I remember correctly, `add_firewall_rules` sets `manager_ref` because the base function synthesizes a `manager_ref` from the rule's properties, but firewall rules from OSP have an ID we can use which is simpler. https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/inventory/persister/builder/network_manager.rb#L60

And yes, if we don't reliably get IDs in events we probably need the same workaround for the firewall rules as for the security groups.

app/models/manageiq/providers/openstack/network_manager/event_target_parser.rb
```diff
@@ -34,7 +34,8 @@ def parse_ems_event_targets(ems_event)
                     :cloud_networks
                   elsif ems_event.event_type.start_with?("subnet.")
                     :cloud_subnets
-                  elsif ems_event.event_type.start_with?("security_group.")
+                  elsif ems_event.event_type.start_with?("security_group")
```
agrare: Minor but I'd prefer being explicit e.g. `event_type.start_with?("security_group.") || event_type.start_with?("security_group_rules..")` just to prevent possibly hitting on something unintended

aufi: Yes, updated.

aufi:
I chose different approach than triggering Security Groups refresh when SecurityGroupRule was changed. And that's handling SecurityGroupRules (FirewallRules in MIQ) as standalone collection.

Working on tests/VCR update..

aufi:
Rubocop - code style follows the rest of the file.

aufi:
@borod108 Hi Boris, this PR changes refresh of network provider, so it might affect ovirt provider VCRs propably. (I'm still working on getting working OpenStack Infra/undercloud provider for infra VCRs, but hopefully this PR could get in in few days)

aufi:
VCRs were finally reloaded and test updated - CI is green.

@mansam May I ask for review?

aufi:
Thanks for merge! 

This is hammer/yes, ivanchuk/yes. Conflicts in VCRs are expected, I'm going to prepare backport PRs.

aufi:
@simaishi Hi Satoe, is there a chance backporting https://github.com/ManageIQ/manageiq-providers-openstack/pull/550 and https://github.com/ManageIQ/manageiq-providers-openstack/pull/551 first? (tests would conflict with this PR backport)

simaishi:
@aufi Merged https://github.com/ManageIQ/manageiq-providers-openstack/pull/550 for hammer branch, but https://github.com/ManageIQ/manageiq-providers-openstack/pull/551 for ivanchuk will need to wait. Please get started with hammer first.

aufi:
@simaishi Sure, thanks!

simaishi:
@aufi Plan changed.. https://github.com/ManageIQ/manageiq-providers-openstack/pull/551 is now merged as well.

simaishi:
Backported to hammer via https://github.com/ManageIQ/manageiq-providers-openstack/pull/553

simaishi:
Backported to ivanchuk via https://github.com/ManageIQ/manageiq-providers-openstack/pull/554

PR #543, opened 2019-12-06
Subclass VmScan job as ManageIQ::Providers::Openstack::CloudManager::Scanning::Job

chessbyte:
State Machine diagram for this subclass of VmScan.

manageiq-providers-openstack-cloud_manager-scanning-job


chessbyte:
@aufi if you have the environment to test SmartState for OpenStack with this and the [base](https://github.com/ManageIQ/manageiq/pull/19607) PR, it would be much appreciated.

PR #547, opened 2020-01-03
Subclass CloudManager metrics capture

kbrock:
openstack is a cloud metrics capture class. subclassing the proper parent

part of ManageIQ/manageiq#19684

NOTE: this one is not open/shut like the others.

The hierarchy is a little odd. Especially since there is a `NetworkManager` with network capture
Wondering if this is good enough, or if we need to get into modules and the like

### DIFF
--- a/app/models/manageiq/providers/openstack/base_metrics_capture.rb
+++ b/app/models/manageiq/providers/openstack/base_metrics_capture.rb
@@ -1,4 +1,4 @@
-class ManageIQ::Providers::Openstack::BaseMetricsCapture < ManageIQ::Providers::BaseManager::MetricsCapture
+module ManageIQ::Providers::Openstack::BaseMetricsCapture
   def perf_collect_metrics(interval_name, start_time = nil, end_time = nil)
     log_header = "[#{interval_name}] for: [#{target.class.name}], [#{target.id}], [#{target.name}]"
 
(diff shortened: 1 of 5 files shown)


app/models/manageiq/providers/openstack/base_metrics_capture.rb
```diff
@@ -1,4 +1,4 @@
-class ManageIQ::Providers::Openstack::BaseMetricsCapture < ManageIQ::Providers::BaseManager::MetricsCapture
+class ManageIQ::Providers::Openstack::BaseMetricsCapture < ManageIQ::Providers::CloudManager::MetricsCapture
```
agrare: Hm actually just noticed this class is used by the Cloud, Infra, and NetworkManagers.  I think this should become a mixin and those classes should then inherit from the relevant base classes.

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 94142a7c79d673463d4c489a677a2b8cb3f2f0a6
Date:   Tue Jan 7 11:42:10 2020 -0500

    Merge pull request #547 from kbrock/metrics_capture_hierarchy

    Subclass CloudManager metrics capture

    (cherry picked from commit 13d2b674c182c51cf70f82aa3fd2162bd66b5149)
```

simaishi:
Reverted from ivanchuk

```
commit 40439e75a8315b01326d392353d973b8369d244c
Date:   Tue Sep 8 11:18:46 2020 -0400

    Revert "Merge pull request #547 from kbrock/metrics_capture_hierarchy"

    This reverts commit 94142a7c79d673463d4c489a677a2b8cb3f2f0a6.
```

PR #549, opened 2020-01-14
Create MIQ snapshot record

andyvesel:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1771605


aufi:
As @mansam suggested, the right way of fixing this issues is to continue on changes from https://github.com/ManageIQ/manageiq-providers-openstack/pull/473

PR #555, opened 2020-01-30
Check for infra_ems before calculating block_storage_disk_capacity

mansam:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1741237


chessbyte:
Can we add a spec for this?

PR #556, opened 2020-01-31
Add STF Event monitor backend

aufi:
Service Telemetry Framework is a new way how Events can be provided by OpenStack. This support was added in OSP16.

Adding backend support for this technology.

## Technology notes
The STF is based on QPID Proton, Ruby bindings: https://github.com/apache/qpid-proton/tree/master/ruby/
That requires having qpid-proton-c package on RPM-based systems or similar for DEB-based. In ManageIQ, it is already used by NUAGE provider.

## Testing steps
1. Have >=OSP16.1 with enabled STF for CloudForms Events
2. Add OSP provider with STF Event source
3. After initial refresh, perform an action in OpenStack (like image upload, provision an Instance or create network)
4. Check that targeted refresh was triggered and events are visible on Provider - Timelines page.
In not clear situations refer to evm.log and search for "STF" keyword.

Related to: https://bugzilla.redhat.com/show_bug.cgi?id=1660842


aufi:
The official name of OpenStack service assurance framework is Service Telemetry Framework (STF), updated name.

.travis.yml
```diff
@@ -13,6 +13,7 @@ env:
   - RUBY_GC_HEAP_GROWTH_FACTOR=1.25
 addons:
   postgresql: '10'
+before_install: sudo apt-get install -y libqpid-proton2 libqpid-proton2-dev
```
agrare, line 16: @aufi we should use the same qpid-proton version as the nuage provider (0.30.0), the version in xenial looks like 0.10.0.

We use 0.30.0 because that is the version of the C library that we can install on the appliance, see https://github.com/ManageIQ/manageiq-providers-nuage/blob/master/tools/ci/before_install.sh and https://github.com/ManageIQ/manageiq-providers-nuage/blob/master/.travis.yml#L27

aufi: Thanks, updated!

app/models/manageiq/providers/openstack/manager_mixin.rb
```diff
@@ -304,6 +314,8 @@ def verify_credentials(args)
     def raw_connect(password, params, service = "Compute")
       if params[:event_stream_selection] == 'amqp'
         amqp_available?(password, params)
+      elsif params[:event_stream_selection] == 'stf'
```
Fryguy, line 317: does it make sense to use a case statement now that it's more than a single if/else?

```ruby
case params[:event_stream_selection]
when "amqp" then amqp_available?(password, params)
when "stf"  then stf_available?(password, params)
else             ems_connect?(password, params, service)
end
```

tools/ci/before_install.sh
```diff
@@ -0,0 +1,30 @@
...
+make all
+
+# Install system libraries
+sudo make install
+
+# Enable the qpid_proton bundler group
+[ -z "$BUNDLE_WITH" ] && bundle config with qpid_proton
+
+popd
+set +v
```
Fryguy, line 30: Minor, but trailing newline is missing.

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
```diff
@@ -0,0 +1,112 @@
+require 'manageiq/providers/openstack/legacy/openstack_event_monitor'
+require 'manageiq/providers/openstack/legacy/events/openstack_event'
+require 'manageiq/providers/openstack/legacy/events/openstack_stf_event_converter'
+require 'manageiq/providers/openstack/legacy/events/openstack_stf_event_receiver'
+require 'manageiq/providers/openstack/legacy/events/openstack_stf_event_test_receiver'
+require 'qpid_proton'
```
Fryguy, line 6: will this all be eager-loaded even if not needed?

tools/ci/before_install.sh
aufi: Thanks, fixed.

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
aufi: This will be loaded when an OpenStack Event monitor is needed (required by manager_mixin methods https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/manager_mixin.rb)

app/models/manageiq/providers/openstack/manager_mixin.rb
aufi: The ```case``` statement makes more sense :+1:  I prefer refactor this method in follow-up PR.

aufi:
Rubocop issues look to me to fail only in cases which fit to existing code in files solving similar tasks. Please ping me if there is some inacceptable format issue.

aufi:
Screencast with quick demo of capturing Events from OpenStack STF using this PR
[STF Events in MIQ quick demo youtube video](https://youtu.be/ICBmlmHW2yQ "STF Events in MIQ quick demo")

aufi:
@agrare Hi Adam, may I ask for review? It is more complicated setup OpenStack with STF Events, so I can provide machine if needed or show demo on my dev setup - https://youtu.be/ICBmlmHW2yQ

config/settings.yml
```diff
@@ -144,6 +144,9 @@
         :amqp_vhost: '/'
         :ceilometer:
           :event_types_regex: '\A(?!firewall|floatingip|gateway|net|port|router|subnet|security_group|vpn|volume|snapshot|backup)'
+        :stf:
+          :topic_name: anycast/ceilometer/event.sample
```
agrare, line 148: Is the topic name really `event.sample?`

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_converter.rb
```diff
@@ -0,0 +1,30 @@
...
+    }
+  end
+
+  private
+
+  def hashize_traits(traits_list)
+    output = {}
+    traits_list.each do |trait_elem|
+      output[trait_elem.first] = trait_elem.last
+    end
```
agrare, line 27: @aufi could this just be `Hash[@event_payload.fetch(:traits)]` ?

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
```diff
@@ -0,0 +1,112 @@
...
+    @recv_block = ->(event) { @events << event }
+
+    @qdr_receiver = Qpid::Proton::Container.new(OpenStackStfEventReceiver.new(self.class.build_qdr_client_url(@options), @config[:topic_name] || DEFAULT_TOPIC_NAME, @recv_block, @events_mutex))
+  end
+
+  def start
+    return if @qdr_receiver.running > 0
+
+    $log.info("STF QDR client START..") if $log
+    @handler = Thread.start { @qdr_receiver.run }
```
agrare, line 56: Is there anything that would restart the handler if `@qdr_receiver` failed?  Thinking maybe `each_batch` can ensure the thread is running and restart it if not.

Also probably good to log any exceptions from the handler thread otherwise if it fails it will do so silently and just look like no events are coming through

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
```diff
@@ -0,0 +1,112 @@
...
+  def stop
+    $log.info("STF QDR client STOP..") if $log
+    @qdr_receiver&.stop
+    @handler&.terminate!
+  end
+
+  def each_batch
+    @collecting_events = true
+    while @collecting_events
+      @events_mutex.synchronize do
```
agrare, line 68: Every time I see a mutex I wonder if we could do this without a mutex haha but we can look at that in a followup

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
```diff
@@ -0,0 +1,112 @@
...
+  def each
+    each_batch do |events|
+      events.each { |e| yield e }
+    end
+  end
+
+  def unserialize_event(raw_event)
+    msg = JSON.parse(raw_event)
+    event = JSON.parse(msg["request"]["oslo.message"])
+    event["payload"] = event["payload"].first
```
agrare, line 99: Is payload always an array of one?

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
```diff
@@ -0,0 +1,112 @@
...
+    event
+  rescue => ex  # need cover more exceptions than JSON::ParserError
+    $log.error("MIQ(#{self.class.name}) Error unserializing STF Event #{ex}") if $log
+    {}
+  end
+
+  def filter_event_types(events)
+    $log.debug("Received new OpenStack STF events: (before filtering)") if $log && events.any?
+    $log.debug(events.inspect) if $log && events.any?
```
agrare, line 108: If logging to debug something relatively large I like to use `$log.debug {}` with a block, so that the block won't even get invoked unless the loglevel is debug.  This prevents `events.inspect` from being evaluated if the loglevel isn't debug.

config/settings.yml
aufi: Yes, it is. (the topic name it has been changed recently on OSP side, but the suffix ```event.sample``` is there..)

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_converter.rb
aufi: The ```:traits``` value is array of 3 element arrays, so it doesn't work. (TBH I don't know why is there the middle element "1", but it is there..)
```
"traits"=>[
            ["service", 1, "network.controller-0.redhat.local"],
            ["request_id", 1, "req-2520570f-7549-414c-98b2-d64ad7e0ffa0"],
            ["project_id", 1, "9609072e9f3b401dac5f7ace13672627"],
            ["user_id", 1, "e4c58a343ce845bbb61fc0b435c5b434"],
            ["tenant_id", 1, "9609072e9f3b401dac5f7ace13672627"],
            ["name", 1, "stfnet7"]
           ],
```

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
aufi: I know this looks strange, the whole process of wrapping internal RabbitMQ Event by Ceilometer backend for distribution via QPID message bus is not well suited for this use case.

QPID/STF Event is generated when a RabbitMQ Event is received, no buffering, no aggregations, it is just wrapped into payload and sent. Confirmed with OSP telemetry team.

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
aufi: Okay :-) Do you mean using other thread data sharing construct or refactor the code to way that the sync would not be needed?

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
aufi: Good point! updating to raise failure when connection dies to let MIQ restart the Event monitor.

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
aufi: Thanks, updated.

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_converter.rb
agrare: Ah okay :+1:

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
aufi: Actually this logging seems to be redundant (events before filtering are logged to debug when received one-by-one).

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
aufi: Thanks for brining this up, this worked bad in previous solution. Updated to raise error which is logged and EventMonitor attempts to reconnect.

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
agrare: Using an already concurrent safe [queue](https://ruby-doc.org/core-2.5.0/Queue.html) would allow us to push from the collector thread and consume from the main thread without implementing our own locking

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
agrare: :+1: nice

app/models/manageiq/providers/openstack/manager_mixin.rb
aufi: https://github.com/ManageIQ/manageiq-providers-openstack/pull/590

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
aufi: Thanks for suggestion, updated by https://github.com/ManageIQ/manageiq-providers-openstack/pull/589

simaishi:
Changed to `jansa/yes?` to let this go through triage process.

chessbyte:
We need a list of PRs to backport to support this feature.  In the meantime, marking as `jansa/no`

aufi:
Thanks for comment, the feature consist of following PRs which needs to be backported:
https://github.com/ManageIQ/manageiq-providers-openstack/pull/556 (openstack backend - this one)
https://github.com/ManageIQ/manageiq-providers-openstack/pull/587 (require qpid update, @agrare please correct me if there is something else related to qpid support)
https://github.com/ManageIQ/manageiq-providers-openstack/pull/592 (UI provider form update, uses DDF, @mzazrivec should create backport PR version once backend PRs get merged)

Qpid-proton itself was bring to MIQ by Nuage provider, so big rubygem updates should not be necessary.

agrare:
Hey @aufi jansa/ivanchuk won't have DDF forms so it will require a new UI PR for those branches

aufi:
> Hey @aufi jansa/ivanchuk won't have DDF forms so it will require a new UI PR for those branches

Correct, it was discussed with UI team, https://github.com/ManageIQ/manageiq-ui-classic/issues/7000

agrare:
Backport with
* https://github.com/ManageIQ/manageiq-providers-openstack/pull/587
* https://github.com/ManageIQ/manageiq-providers-openstack/pull/590

simaishi:
@aufi `rescue` in app/models/manageiq/providers/openstack/event_catcher_mixin.rb (which was added in #523) doesn't exist in ivanchuk branch. Should I also backport #523 or just skip that part of the change when backporting this PR?

simaishi:
Jansa backport details:

```
$ git log -1
commit 9e708534ebc926c26309a353754975b20eba8018
Date:   Fri May 22 08:41:19 2020 -0400

    Merge pull request #556 from aufi/saf_event_monitor

    Add STF Event monitor backend

    (cherry picked from commit 8228ffc1fcf390ee9db56b5b8a3274cc3d16066c)
```

aufi:
@simaishi https://github.com/ManageIQ/manageiq-providers-openstack/pull/523 is not needed for STF, it handles Panko vs. Ceilometer service availability which is different event monitor type than STF.

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit e92177f81ead656ad294af0759b84c368e2689c4
Date:   Fri May 22 08:41:19 2020 -0400

    Merge pull request #556 from aufi/saf_event_monitor

    Add STF Event monitor backend

    (cherry picked from commit 8228ffc1fcf390ee9db56b5b8a3274cc3d16066c)

    https://bugzilla.redhat.com/show_bug.cgi?id=1757128
```

PR #557, opened 2020-02-10
Make the DDF fields more similar to the current form's layout + i18n

skateman:
Updating the initial DDF schema to look more similar to the current form's layout. Here are the catches:
* Keystone v2 as default API version - suggested by @aufi 
* Events type selection changed from `radio` to `select` because DDF doesn't do inline radios
* SSL enforcement dropdown doesn't allow to select an option - reported [here](https://github.com/ManageIQ/manageiq-ui-classic/issues/6656#issuecomment-584624556)
* No AMQP fallback hostnames - will be added after [this bug](https://bugzilla.redhat.com/show_bug.cgi?id=1801583) gets fixed
* Made API port for AMQP a required field
* RSA Private Key field is set to password as DDF doesn't do file uploads [yet](https://github.com/data-driven-forms/react-forms/issues/307)

**Old Default Endpoint:**
Screenshot from 2020-02-12 10-07-43
**Old AMQP Endpoint:**
Screenshot from 2020-02-12 10-07-32
**Old RSA Endpoint:**
Screenshot from 2020-02-12 10-07-54

**New Default Endpoint:**
Screenshot from 2020-02-12 10-30-36

**New AMQP Endpoint:**
Screenshot from 2020-02-12 10-31-00

**New RSA Endpoint:**
Screenshot from 2020-02-12 10-32-27

The `verify_credentials` has been also updated in order to support validation for AMQP.

### DIFF
--- a/app/models/manageiq/providers/openstack/manager_mixin.rb
+++ b/app/models/manageiq/providers/openstack/manager_mixin.rb
@@ -31,7 +31,7 @@ module ManageIQ::Providers::Openstack::ManagerMixin
       ems.security_protocol      = params[:default_security_protocol].strip
       ems.keystone_v3_domain_id  = params[:keystone_v3_domain_id]
 
-      user, hostname, port = params[:default_userid], params[:default_hostname].strip, params[:default_api_port].strip
+      user, hostname, port = params[:default_userid], params[:default_hostname].strip, params[:default_api_port].try(:strip)
 
       endpoint = {:role => :default, :hostname => hostname, :port => port, :security_protocol => ems.security_protocol}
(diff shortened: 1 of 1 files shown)


aufi:
Looks good to me :+1: 

The "No AMQP fallback hostnames" doesn't seem to be an issue of OpenStack provider. The provider backend code counts with these attributes, but current UI doesn't fill it into database.

aufi:
@skateman I checked out this locally I cannot see the change showed on attached screenshots. Is there any other dependent PR?

skateman:
@aufi yeah, there's a [WIP PR](https://github.com/ManageIQ/manageiq-ui-classic/pull/6402), you will need to rebase and run `bin/update`, however, it only implements validation for now, not saving.

skateman:
Can we get this in please? I would like to work on this further.

PR #560, opened 2020-03-02
Remove invalid partials from specs

djberg96:
This removes a few invalid partials that aren't actually implemented, which will cause failures if strict partial validation is enabled.

Removing these had no effect.


djberg96:
@miq-bot add_label test, refactoring

PR #561, opened 2020-03-05
Override inventory object refresh

d-m-u:
we need to do this for removal of the legacy ems_refresh stuff
it's adam's circus

### DIFF
--- a/app/models/manageiq/providers/openstack/infra_manager/refresher.rb
+++ b/app/models/manageiq/providers/openstack/infra_manager/refresher.rb
@@ -12,6 +12,21 @@ module ManageIQ
         EmsRefresh.queue_refresh(ems.network_manager)
       end
 
+      def collect_inventory_for_targets(ems, _targets)
+        [[ems, nil]]
+      end
+
+      def parse_targeted_inventory(ems, _target, _inventory)
+        log_header = format_ems_for_logging(ems)
+        _log.debug("#{log_header} Parsing inventory...")
+        hashes = ems.class::RefreshParser.ems_inv_to_hashes(ems, refresher_options)
+        _log.debug("#{log_header} Parsing inventory...Complete")
+        hashes
+      end
+
+      def preprocess_targets_manager_refresh
+      end
+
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/openstack/infra_manager.rb
```diff
@@ -34,6 +34,10 @@ def cloud_tenants
     self.class.none
   end
 
+  def inventory_object_refresh?
+    true
+  end
```
agrare: OpenStack Infra doesn't have graph refresh yet so we need to make sure that the [`Openstack::InfraManager::Refresher`](https://github.com/ManageIQ/manageiq-providers-openstack/blob/30e0ac3dc8ca4c80bb7ab3fdf7af7323a993a191/app/models/manageiq/providers/openstack/infra_manager/refresher.rb) overrides any methods from the `BaseManager` refresher that call out to `ems.inventory_object_refresh?` aren't being used if we are going to remove that check from core.

d-m-u: sorry that was silly

app/models/manageiq/providers/openstack/infra_manager/refresher.rb
```diff
@@ -12,6 +12,35 @@ def save_inventory(ems, target, hashes)
...
+        _log.debug("#{log_header} Parsing inventory...")
+        hashes = parse_legacy_inventory(ems)
+        _log.debug("#{log_header} Parsing inventory...Complete")
+        hashes
+      end
+
+      def preprocess_targets_manager_refresh
+        @targets_by_ems_id.each do |ems_id, targets|
+          ems = @ems_by_ems_id[ems_id]
+          next unless ems.inventory_object_refresh?
```
agrare: @d-m-u since this doesn't support inventory_object_refresh think this method should just be empty

agrare:
Thanks @d-m-u, now that https://github.com/ManageIQ/manageiq-providers-openstack/pull/564 is merged we no longer need this

PR #563, opened 2020-03-12
Add graph refresh to the OpenStack InfraManager

agrare:
This adds the option for graph refresh to the Openstack::InfraManager.

Depends:
- [x] https://github.com/ManageIQ/manageiq/pull/19958
- [x] https://github.com/ManageIQ/manageiq/pull/19960

Core issue: https://github.com/ManageIQ/manageiq/issues/19888

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/collector/infra_manager.rb
+++ b/app/models/manageiq/providers/openstack/inventory/collector/infra_manager.rb
@@ -1,3 +1,9 @@
 class ManageIQ::Providers::Openstack::Inventory::Collector::InfraManager < ManageIQ::Providers::Openstack::Inventory::Collector
   include ManageIQ::Providers::Openstack::Inventory::Collector::HelperMethods
+
+  def images
+    return [] unless image_service
+    return @images if @images.any?
+    @images = uniques(image_service.handled_list(:images))
+  end
 end
--- a/app/models/manageiq/providers/openstack/inventory/collector/infra_manager.rb
+++ b/app/models/manageiq/providers/openstack/inventory/collector/infra_manager.rb
@@ -1,0 +1,3 @@
+class ManageIQ::Providers::Openstack::Inventory::Collector::InfraManager < ManageIQ::Providers::Openstack::Inventory::Collector
+  include ManageIQ::Providers::Openstack::Inventory::Collector::HelperMethods
+end
--- a/app/models/manageiq/providers/openstack/inventory/collector/infra_manager.rb
+++ b/app/models/manageiq/providers/openstack/inventory/collector/infra_manager.rb
@@ -26,4 +26,11 @@ class ManageIQ::Providers::Openstack::Inventory::Collector::InfraManager < Manag
 
     @servers = uniques(compute_service.handled_list(:servers))
   end
+
+  def hosts
+    return [] unless baremetal_service
+    return @hosts if @hosts.any?
+
+    @hosts = uniques(baremetal_service.handled_list(:nodes))
+  end
 end
--- a/app/models/manageiq/providers/openstack/inventory/collector/infra_manager.rb
+++ b/app/models/manageiq/providers/openstack/inventory/collector/infra_manager.rb
@@ -4,6 +4,7 @@ class ManageIQ::Providers::Openstack::Inventory::Collector::InfraManager < Manag
   def images
     return [] unless image_service
     return @images if @images.any?
+
     @images = uniques(image_service.handled_list(:images))
   end
 end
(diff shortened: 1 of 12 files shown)


app/models/manageiq/providers/openstack/inventory/collector/infra_manager.rb
```diff
@@ -1,3 +1,9 @@
 class ManageIQ::Providers::Openstack::Inventory::Collector::InfraManager < ManageIQ::Providers::Openstack::Inventory::Collector
   include ManageIQ::Providers::Openstack::Inventory::Collector::HelperMethods
+
+  def images
+    return [] unless image_service
+    return @images if @images.any?
+    @images = uniques(image_service.handled_list(:images))
```
chessbyte: why not the more standard `@images ||= uniques(image_service.handled_list(:images))` ?

agrare: Because `@images` is initialized with `[]` in the collector initializer so `||=` wouldn't actually fetch anything

spec/models/manageiq/providers/openstack/infra_manager/refresher_rhos_juno_spec.rb
```diff
@@ -10,101 +12,116 @@
...
+  it "graph refresh the same as classic refresh" do
+    stub_settings_merge(:ems_refresh => {:openstack_infra => {:inventory_object_refresh => false}})
+    full_refresh
+    before = serialize_inventory
+
+    stub_settings_merge(:ems_refresh => {:openstack_infra => {:inventory_object_refresh => true}})
+    full_refresh
+    after = serialize_inventory
+
+    assert_inventory_not_changed(before, after)
```
agrare: This does a full database comparison between classic and graph refresh

agrare:
@aufi can you review please?  I started by copying the existing RefreshParser then updating it to work with graph refresh, then moving the collection parts into the Collector.

It passes with the same VCR cassettes, the same specs, and additionally I did a full db comparison of classic and graph refresh.

app/models/manageiq/providers/openstack/inventory/parser/infra_manager.rb
```diff
@@ -0,0 +1,366 @@
...
+
+  def lookup_connection_state(power_state_input)
+    case power_state_input
+    when "power on"               then "connected"
+    when "power off", "rebooting" then "disconnected"
+    else                               "disconnected"
+    end
+  end
+
+  def clusters_and_host_mapping
```
agrare: NOTE: I'd like to move this into the Collector as `collector.clusters` but currently it uses the parsed orchestration stacks (`@data_index[:orchestration-stacks]`).  Once we move this to use the collected stacks instead then we can drop all of the `process_collection`/`@data`/`@data_index` stuff.

That said, this works the same way as it does in classic refresh and I'd like to do this change in a follow-up PR

app/models/manageiq/providers/openstack/inventory/parser/infra_manager.rb
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    persister.clusters.build(new_result)
+
+    return uid, new_result
+  end
+
+  def set_relationship_on_hosts(hosts, cluster_host_mapping)
+    hosts.each do |host|
+      host.ems_cluster = persister.clusters.lazy_find(cluster_host_mapping[host[:uid_ems]])
```
agrare: This I'd also like to move to the collector, something like `cluster_by_host_uuid` so that it can be set by the host parser which would allow us to do targeted refresh in the future

gtanzillo:
@aufi Please take a look

app/models/manageiq/providers/openstack/inventory/parser/infra_manager.rb
agrare: Nevermind I was able to get this working pretty easily

app/models/manageiq/providers/openstack/inventory/parser/infra_manager.rb
agrare: Also fixed, we set the link to the cluster in the host parser now

aufi:
Hi, first I'd like thank you for continue work on this feature!

The code looks good and works for me (tested with lab OSP10 undercloud) :+1: 

Adding just 3 notes which don't block merging:
* I'm not sure if a Storage manager for the Infra provider would be neeed (e.g. information about disks is not part of the Host IIUC)
* targeted refresh for infra could be enabled, events are there..
* refreshing of orchestration stack is not very fast (this is not new issue, fog-openstack changes would be needed to list orch.stacks with details)

agrare:
> I'm not sure if a Storage manager for the Infra provider would be neeed (e.g. information about disks is not part of the Host IIUC)

Yes you are correct, I wanted to get this passing with the same VCRs.  In a follow-up I want to remove this and remove it from the VCR cassettes

aufi:
OK, merging!

agrare:
Thanks @aufi !

simaishi:
Jansa backport details:

```
$ git log -1
commit d8867844884b420e6222fc28ef13ccaae819b5e1
Author: Marek Aufart <maufart@redhat.com>
Date:   Tue Mar 17 17:06:45 2020 +0100

    Merge pull request #563 from agrare/add_graph_refresh_infra_manager

    Add graph refresh to the OpenStack InfraManager

    (cherry picked from commit 77fd88c701707a19f4134fd28882dad5820a4463)
```

PR #564, opened 2020-03-17
Remove classic refresh from Openstack InfraManager

agrare:
The old style of refresh is going to be removed in a future version so we should have graph refresh be the only option.

https://github.com/ManageIQ/manageiq/issues/19888


app/models/manageiq/providers/openstack/inventory/parser.rb
```diff
@@ -38,4 +38,10 @@ def orchestration_template(stack)
       o
     end
   end
+
+  private
+
+  def compose_ems_ref(*keys)
+    keys.join('_')
+  end
```
agrare: This was the only method left being used from the `ManageIQ::Providers::Openstack::RefreshParserCommon::OrchestrationStacks` module allowing us to delete it

aufi:
Looks and works good for me :+1:

simaishi:
Jansa backport details:

```
$ git log -1
commit ccca8d4fbbbffe65632254949d8e649dfac4268d
Author: Marek Aufart <maufart@redhat.com>
Date:   Wed Mar 18 13:18:47 2020 +0100

    Merge pull request #564 from agrare/remove_classic_refresh_infra_manager

    Remove classic refresh from Openstack InfraManager

    (cherry picked from commit cc7898fdbdb6288ce73c0c5eaeddacbd9a5927bc)
```

PR #565, opened 2020-03-18
Don't collect storage/objects in the InfraManager

agrare:
We don't need to connect to the storage service nor collect objects in the InfraManager

Depends on: ~~https://github.com/ManageIQ/manageiq-providers-openstack/pull/564~~


aufi:
Just for clarification - @agrare why the Infra Images are not needed? On OpenStack Undercloud there are Images for provisioning baremetals and such images are not part of Storage AFAIK and should be visible in MIQ.

agrare:
Ah sorry not images, I meant objects thanks @aufi I'll correct the title

aufi:
Thanks, looks and works good, merging :+1:

agrare:
Thanks @aufi !

PR #568, opened 2020-03-19
Adjusted DDF schema for create/edit and validation

skateman:
Parent issue: https://github.com/ManageIQ/manageiq/issues/18818

There's no file upload component yet in the current DDF version. After the v2 update, I can replace the secret text field with a file upload.

### DIFF
--- a/app/models/manageiq/providers/openstack/manager_mixin.rb
+++ b/app/models/manageiq/providers/openstack/manager_mixin.rb
@@ -59,10 +59,21 @@ module ManageIQ::Providers::Openstack::ManagerMixin
             :label     => _("Provider Region"),
           },
           {
+            :component => "select-field",
+            :name      => "provider_id",
+            :label     => _("Openstack Infra Provider"),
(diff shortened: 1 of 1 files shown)


Fryguy:
Did you want this in before or after the new `service` column goes in?

skateman:
@Fryguy not adjusted to the new column yet, that's why the WIP

app/models/manageiq/providers/openstack/manager_mixin.rb
```diff
@@ -495,5 +546,14 @@ def orchestration_template_validate(template)
     raise MiqException::MiqOrchestrationValidationError, err.to_s, err.backtrace
   end
 
+  def authentication_class(authtype)
```
agrare, line 549: Can this be in the `InfraManager` instead of the `ManagerMixin`?  That way it isn't accidentally returning `ManageIQ::Providers::Openstack::InfraManager::AuthKeyPair` for the Openstack Cloud/Network/Storage managers

PR #572, opened 2020-03-29
Remove unused ems_inv_to_hashes

agrare:
The RefreshParser.ems_inv_to_hashes method was used by legacy refresh
and no longer needed


aufi:
Looks good to me, merging.

PR #573, opened 2020-03-31
If we fail to find a tenant don't return an Array

agrare:
If we fail to find a tenant that works the default_tenant_name method was return the list of tenant_names to the caller.


aufi:
Makes sense :+1:

PR #575, opened 2020-04-02
Don't connect to Volume service on VM provisioning without Volumes

aufi:
VM provisioning code connected to Volume (Cinder) service before was iterating volumes hat
should be created. This connection was not needed if the Volumes array was empty.

Adding conditional skipping the Volume service connection.


agrare:
@aufi I think this might already be done by the core state machine: https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/cloud_manager/provision/state_machine.rb#L13-L16

aufi:
Good point! moved to main repo.

PR #579, opened 2020-04-06
Handle Keystone service without a region

agrare:
If the keystone identity service is deployed without a region then passing `:openstack_region` will cause a missing endpoint failure.


aufi:
The change looks good. It is important mention that this is not issue when using Red Hat OpenStack or TripleO / Packstack, but Canonical's Microstack-only as far we know.

Merging since the change seems to be working. If it would turn out that the change breaks some of existing test scenarios during QE process, it should be reverted..

PR #581, opened 2020-05-12
Implemented authentication_class selection logic based on authtype

skateman:
This is required for selecting the correct STI subclass in https://github.com/ManageIQ/manageiq/pull/20157

Related issue: https://github.com/ManageIQ/manageiq/issues/18818

@miq-bot add_label enhancement

### DIFF
--- a/app/models/manageiq/providers/openstack/infra_manager.rb
+++ b/app/models/manageiq/providers/openstack/infra_manager.rb
@@ -217,4 +217,10 @@ class ManageIQ::Providers::Openstack::InfraManager < ManageIQ::Providers::InfraM
       end
     end
   end
+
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/openstack/manager_mixin.rb
```diff
@@ -495,5 +495,14 @@ def orchestration_template_validate(template)
     raise MiqException::MiqOrchestrationValidationError, err.to_s, err.backtrace
   end
 
+  def authentication_class(attributes)
```
agrare, line 498: This should be in the `Openstack::InfraManager` not the `ManagerMixin`

PR #584, opened 2020-05-14
Ensure that the AMQP port gets properly to the validation in DDF

skateman:
The `Endpoint` model has a `port` field, but in the AMQP vaidation it's referred as `api_port` which is not very consistent. The DDF schema should and uses the `port` version, as the data from the API arrives so. In order to make sure that the validation will get the port, I had to add a fallback.

I'm not sure if the string conversion is necessary, but the old forms were sending the value as a string, so I wanted to play it safe.

Parent issue: https://github.com/ManageIQ/manageiq/issues/18818

@miq-bot assign @agrare 
@miq-bot add_label bug

### DIFF
--- a/app/models/manageiq/providers/openstack/manager_mixin.rb
+++ b/app/models/manageiq/providers/openstack/manager_mixin.rb
@@ -18,7 +18,7 @@ module ManageIQ::Providers::Openstack::ManagerMixin
         :hostname => params[:amqp_hostname],
         :username => params[:amqp_userid],
         :password => ManageIQ::Password.try_decrypt(password),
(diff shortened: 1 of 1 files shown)


aufi:
Hi, ok :+1:  Will update STF backend for ```stf_port``` instead of ```stf_api_port``` key.

aufi:
Updated https://github.com/ManageIQ/manageiq-providers-openstack/pull/556/files#diff-6f9b687ce51fd31aa102fa9305ef2501

aufi:
Looks good to me :+1:

PR #585, opened 2020-05-18
Fix Openstack service status section for infra node

andyvesel:
FIX: https://bugzilla.redhat.com/show_bug.cgi?id=1723864
Updated the bash condition which manageiq-ssh-util gem couldn't handle without errors.

PR #586, opened 2020-05-19
Remove sudo prepend to bash condition

andyvesel:
Set `@passwordless_sudo` to false in order to run bash `if` condition without `sudo`.
Fixed https://bugzilla.redhat.com/show_bug.cgi?id=1723864
Related to https://github.com/ManageIQ/manageiq-providers-openstack/pull/585

### DIFF
--- a/app/models/manageiq/providers/openstack/infra_manager/host.rb
+++ b/app/models/manageiq/providers/openstack/infra_manager/host.rb
@@ -149,6 +149,7 @@ class ManageIQ::Providers::Openstack::InfraManager::Host < ::Host
 
   def refresh_openstack_services(ssu)
     openstack_status = ssu.shell_exec("systemctl -la --plain | awk '/openstack/ {gsub(/ +/, \" \"); gsub(\".service\", \":\"); gsub(\"not-found\",\"(disabled)\"); split($0,s,\" \"); print s[1],s[3],s[2]}' | sed \"s/ loaded//g\"")
+    ssu.instance_variable_set(:@passwordless_sudo, false)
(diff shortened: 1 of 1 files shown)


agrare:
@andyvesel Not a fan of the instance_variable_set, also not sure why you'd want to disable running the whole command as root when you want do `sudo podman ps`.  Is it because you don't want to run the `docker ps` as root?

It isn't clear from the BZ or this description what the actual problem is and how this fixes it.

andyvesel:
@agrare the issue was that `ssu.shell_exec(list_all_service_containers_cmd)` prepends `sudo` to bash `if`condition, according to [this comment](https://github.com/ManageIQ/manageiq-ssh-util/blob/master/lib/manageiq/ssh/util.rb#L123). Eventually it looks like 
`sudo if [ -e /usr/bin/podman ]; then sudo podman ps --format "{{.Names}} {{.Status}}"; \
    elif [ -e /usr/bin/docker ]; then docker ps --format "table {{.Names}}\t{{.Status}}" | tail -n +2; fi'` which complains on a wrong syntax and causes an error.
I simply set `@passwordless_sudo` to false in this case, but maybe yes, updating instance variable is not a good idea here.

agrare:
Yeah aside from `#instance_variable_set` being a bit of a code smell, it also would leave it as false for anything that uses `ssu` to run commands after this.

You could use `sudo bash -c 'if []; then ...; elif []; then; fi` or do this in two calls.

agrare:
@andyvesel what is the status of this?  I understand this is broken right now

PR #589, opened 2020-05-25
Refactor STF Event Monitor - replace Mutex with Queue

aufi:
Replacing Mutex-synced Array for storing captured Events with a Queue object.

Based on Adam's comment on https://github.com/ManageIQ/manageiq-providers-openstack/pull/556

### DIFF
--- a/lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
+++ b/lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
@@ -6,8 +6,9 @@ require 'manageiq/providers/openstack/legacy/events/openstack_stf_event_test_rec
 require 'qpid_proton'
 
 class OpenstackStfEventMonitor < OpenstackEventMonitor
-  DEFAULT_AMQP_PORT  = 5666
-  DEFAULT_TOPIC_NAME = 'anycast/ceilometer/event.sample'.freeze
+  DEFAULT_AMQP_PORT       = 5666
+  DEFAULT_TOPIC_NAME      = 'anycast/ceilometer/event.sample'.freeze
+  EVENTS_BATCH_MAX_SIZE   = 200
 
   def self.available?(options = {})
     $log.info("Testing connection to STF..") if $log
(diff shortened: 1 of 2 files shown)


lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
```diff
@@ -66,24 +65,20 @@ def stop
...
-
-        $log.info("MIQ(#{self.class.name}) STF Yielding #{filtered_events.size} events to"\
-        " event_catcher: #{filtered_events.map { |e| e }}") if $log
-
-        yield filtered_events
-
-        $log.info("MIQ(#{self.class.name}) Clearing events") if $log && @events.any? && filtered_events.any?
-        @events.clear
+      captured_events = []
+      while !@events.empty? && captured_events.length < EVENTS_BATCH_MAX_SIZE
```
agrare, line 69: If the queue is empty this won't block, what if we do something like this:

```
captured_events = []
captured_events << @events.pop # this will block until there is something in the queue
captured_events << @events.pop until @events.empty? || captured_events.length >= EVENTS_BATCH_MAX_SIZE # this will dequeue everything off the queue without blocking

yield filtered_events
```

This way we can get rid of the `sleep 5`

aufi: Hi Adam, thanks for the comment!

> If the queue is empty this won't block

That was actually intentional, at least in my view. I planned to keep some level of similarity with other OpenStack Event monitors which work with Events in batches (passing events with ```yield events_array```).

I don't think that pop-ing a next Event just after one was received will return something very often. Switching a thread and/or skipping on next line in Ruby process should be faster than Event generation and its delivery from OpenStack via network (even using AMQP) IMO. So it would end up in processing event-by-event. That that is be possible too..

agrare: Ah okay so you want to give a chance for more events to fill up, maybe `queue.pop; sleep N; queue.pop until queue.empty` ?

If you just want to stick with what you have here that's fine, it just does some unnecessary operations every few seconds

lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
```diff
@@ -66,24 +65,20 @@ def stop
...
+      end
+      converted_events = captured_events.map do |raw_event|
+        unserialized_event = unserialize_event(raw_event)
+        converted_event = OpenstackStfEventConverter.new(unserialized_event)
+        $log.debug("Processing a new OpenStack STF Event: #{unserialized_event.inspect}") if $log
+        openstack_event(nil, converted_event.metadata, converted_event.payload)
       end
+      filtered_events = filter_event_types(converted_events)
+      $log.info("MIQ(#{self.class.name}) STF Yielding #{filtered_events.size} events to"\
+      " event_catcher: #{filtered_events.map { |e| e }}") if $log
```
agrare, line 80: @aufi so this will log an empty array every 5 seconds?  Also didn't notice this before but what is `filtered_events.map { |e| e }` doing?

PR #591, opened 2020-05-26
Handle timeouts for unsupported services, as well as possible Neutron failure

djberg96:
When performing a refresh in our dev environment, I ran into an issue where it would hit a timeout on certain services, causing the whole refresh to fail. Specifically, stacks. It would also fail if Neutron had died with a connection refusal.

Interestingly, the Openstack UI seems to have a similar issue, and attempting to view stack information caused it to spin infinitely.

I've updated the code to handle both a timeout and a socket connection failure. The latter is unlikely in practice, but it's still better to catch it than die IMO.


djberg96:
The Hakiri warning/failure is unrelated to this PR.

djberg96:
@aufi Please review

aufi:
LGTM, merging :+1:

PR #592, opened 2020-05-26
Update DDF form with the STF event monitor and adjust validation

skateman:
This is the UI part of the STF event monitor, after https://github.com/ManageIQ/manageiq-ui-classic/pull/6698 gets merged, any form change can be addressed in the provider repo by editing the `params_for_create` method.

Unfortunately I have no access to a live STF provider and I had issues with the `qpid_proton` gem so I couldn't test this properly, @aufi can you please try this?

Screenshot from 2020-05-26 17-25-29

Fixes https://github.com/ManageIQ/manageiq-ui-classic/issues/7000

@miq-bot assign @aufi


djberg96:
The Hakiri failure is unrelated.

PR #593, opened 2020-05-26
Update activesupport requirement

djberg96:
This updates the activesupport gem from 5.0.0 to 5.2.4. This addresses CVE-2020-8165, which apparently suffers from a potential issue with MemCacheStore and RedisCacheStore.

https://groups.google.com/forum/#!topic/rubyonrails-security/bv6fW4S0Y1c

### DIFF
--- a/manageiq-providers-openstack.gemspec
+++ b/manageiq-providers-openstack.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,config,lib}/**/*"]
 
-  s.add_runtime_dependency "activesupport",        ">= 5.2.4.3", "~> 5.2.4"
+  s.add_runtime_dependency "activesupport",        "~> 5.2.4.3"
   s.add_runtime_dependency "bunny",                "~> 2.1.0"
   s.add_runtime_dependency "excon",                "~> 0.71"
   s.add_runtime_dependency "fog-openstack",        ">= 0.3.10"
--- a/manageiq-providers-openstack.gemspec
+++ b/manageiq-providers-openstack.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,config,lib}/**/*"]
 
-  s.add_runtime_dependency "activesupport",        ">= 5.0", "< 5.3"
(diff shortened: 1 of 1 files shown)


manageiq-providers-openstack.gemspec
```diff
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,config,lib}/**/*"]
 
-  s.add_runtime_dependency "activesupport",        ">= 5.0", "< 5.3"
+  s.add_runtime_dependency "activesupport",        "~>5.2.4", ">= 5.2.4.3", "< 5.3"
```
Fryguy, line 16: You can remove the "< 5.3" because that can't happen anyway with "~>5.2.4"

djberg96: Updated.

djberg96:
@Fryguy ok, Hakiri is saying "Versions Affected: rails < 5.2.5, rails < 6.0.4", so even with this change it's getting the error.

manageiq-providers-openstack.gemspec
```diff
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,config,lib}/**/*"]
 
-  s.add_runtime_dependency "activesupport",        ">= 5.0", "< 5.3"
+  s.add_runtime_dependency "activesupport",        "~> 5.2.4", ">= 5.2.4.3"
```
agrare, line 16: @djberg96 try `"activesupport", ">= 5.2.4.3", "~> 5.2.4"`  I've seen hakiri not process all the versions passed the same way that bundler does and helps to have the unaffected version first.

agrare:
:crossed_fingers:

agrare:
@Fryguy hakiri isn't green with anything less than `"~> 5.2.4.3"`, even `>= 5.2.4.3 ~> 5.2.4`

What do you want to do here?  I'm tempted to say lets set the version "properly" (aka >= 5.2.4.3 ~> 5.2.4) even if hakiri doesn't recognize it as fixed.

Fryguy:
Version it properly and I'll send a bug report to them

PR #600, opened 2020-05-29
Cache metrics api call

andyvesel:
https://bugzilla.redhat.com/show_bug.cgi?id=1840945

### DIFF
--- a/app/models/manageiq/providers/openstack/base_metrics_capture.rb
+++ b/app/models/manageiq/providers/openstack/base_metrics_capture.rb
@@ -1,4 +1,5 @@
 module ManageIQ::Providers::Openstack::BaseMetricsCapture
+  extend ActiveSupport::Concern
   def perf_collect_metrics(interval_name, start_time = nil, end_time = nil)
     log_header = "[#{interval_name}] for: [#{target.class.name}], [#{target.id}], [#{target.name}]"
 
@@ -92,30 +93,36 @@ module ManageIQ::Providers::Openstack::BaseMetricsCapture
   def meter_names(metric_capture_module)
     @meter_names ||= metric_capture_module::COUNTER_NAMES
   end
+  
+  included do
+    cache_with_timeout(:counters_by_vm, 60.minutes) { Hash.new }
+  end
 
   def find_meter_counters(metric_capture_module, resource_filter, metadata_filter, log_header)
-    counters = list_resource_meters(resource_filter, log_header) + list_metadata_meters(metadata_filter, log_header)
-
-    # With Gnocchi, the network metrics are not associated with the instance's resource id
-    # but with the instance's network interface resource id. Here we fetch the counters
-    # for the network interface, so that the network metrics can be fetched.
-    if target.respond_to?(:network_ports)
-      target.network_ports.each do |port|
-        # fetch the list of resources and use the original_resource_id and type to find
-        # the network interface's resource
-        if port.ems_ref
-          original_resource_id = "#{target.ems_ref}-tap#{port.ems_ref[0..10]}"
-          resources = @perf_ems.list_resources.body
-          resources.each do |r|
-            if r["type"].to_s == "instance_network_interface" && r["original_resource_id"].include?(original_resource_id)
-              resource_filter = {"field" => "resource_id", "value" => r["id"]}
-              counters = counters + list_resource_meters(resource_filter, log_header)
+    counters = self.class.counters_by_vm.dig(ems.id, target.ems_ref)
+    if counters.nil?
+      counters = list_resource_meters(resource_filter, log_header) + list_metadata_meters(metadata_filter, log_header)
+      # With Gnocchi, the network metrics are not associated with the instance's resource id
+      # but with the instance's network interface resource id. Here we fetch the counters
+      # for the network interface, so that the network metrics can be fetched.
+      if target.respond_to?(:network_ports)
+        target.network_ports.each do |port|
+          # fetch the list of resources and use the original_resource_id and type to find
+          # the network interface's resource
+          if port.ems_ref
+            original_resource_id = "#{target.ems_ref}-tap#{port.ems_ref[0..10]}"
+            resources = @perf_ems.list_resources.body
+            resources.each do |r|
+              if r["type"].to_s == "instance_network_interface" && r["original_resource_id"].include?(original_resource_id)
+                resource_filter = {"field" => "resource_id", "value" => r["id"]}
+                counters = counters + list_resource_meters(resource_filter, log_header)
+              end
             end
           end
         end
       end
+      self.class.counters_by_vm.store_path(ems.id, target.ems_ref, counters)
     end
-
     # Select only allowed counters, with unique names
     counters.select { |c| meter_names(metric_capture_module).include?(c["name"]) }.uniq { |x| x['name'] }
   end


andyvesel:
@aufi @agrare opening WIP to have a discussion here

agrare:
Hey @andyvesel 

Since cache_with_timeout saves this at the class level, and since the MetricsCollectorWorkers aren't per-ems you have to store these per-ems.  What I was thinking was to cache the counters [here](https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/base_metrics_capture.rb#L97) and have e.g.:

```
cache_with_timeout(:counters) { Hash.new }
```

Then your `find_meter_counters` method can check for e.g. `counters[ems.id]` and if they don't exist go and fetch them.

This block https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/base_metrics_capture.rb#L102-L116 looks like it might also be expensive, but also not sure if we want to be caching that for ever vm target.

aufi:
Thanks for comment @agrare!

The lines https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/base_metrics_capture.rb#L102-L116 produce at least the same API calls amount as the line 97 (each VM has at least one Network interface normally), so it would be useful cache it too.

I'm not sure I understand the creation of the caching key fully in this method, but the cached counters hash should be indexed maybe by EMS_ID, but also by VM ID (or ems_ref) since the method is iterated with each VM and this identification is part of ```resource_filter``` parameter.

app/models/manageiq/providers/openstack/base_metrics_capture.rb
```diff
@@ -94,22 +94,26 @@ def meter_names(metric_capture_module)
   end
 
   def find_meter_counters(metric_capture_module, resource_filter, metadata_filter, log_header)
-    counters = list_resource_meters(resource_filter, log_header) + list_metadata_meters(metadata_filter, log_header)
-
+    # cache counters to prevent too frequent calls to Gnocchi API
+    cache_with_timeout(:counters, 60.minutes) do
+      counters = list_resource_meters(resource_filter, log_header) + list_metadata_meters(metadata_filter, log_header)
+    end
```
agrare, line 100: The `cache_with_timeout` method defines a class-level singleton so this should be pulled out of the method and moved to the class level.  Something like `cache_with_timeout(:counters_by_vm, 60.minutes) { Hash.new }` then you should use the ems.id and vm.ems_ref as keys into the hash so that one metrics capture worker across multiple EMSs won't have hash collisions.

app/models/manageiq/providers/openstack/base_metrics_capture.rb
```diff
@@ -92,9 +93,17 @@ def list_meters(filter)
...
+  included do
+    cache_with_timeout(:counters_by_vm, 30.seconds) { Hash.new }
+  end
 
   def find_meter_counters(metric_capture_module, resource_filter, metadata_filter, log_header)
-    counters = list_resource_meters(resource_filter, log_header) + list_metadata_meters(metadata_filter, log_header)
+    counters = []
+    if self.class.counters_by_vm.dig(ems.id, target.ems_ref).nil?
+      counters = list_resource_meters(resource_filter, log_header) + list_metadata_meters(metadata_filter, log_header)
+      self.class.counters_by_vm.store_path(ems.id, target.ems_ref, counters)
```
agrare, line 105: Hey so this looks like it is still caching the counter values rather than the meter names, this is the API call that is done every metrics capture call for every VM https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/lib/manageiq/providers/openstack/legacy/openstack_handle/metric_delegate.rb#L17-L31 so I was expecting that we'd cache that.

Caching the counter values for 30 seconds doesn't really help us

agrare: Actually its possible I'm not understanding what is actually stored in `counters`, but it shouldn't be the value of a counter since those are expected to change constantly, it should be the name of a counter which doesn't change very often if at all.

app/models/manageiq/providers/openstack/base_metrics_capture.rb
```diff
@@ -92,9 +93,17 @@ def list_meters(filter)
...
   end
+  
+  included do
+    cache_with_timeout(:counters_by_vm, 30.seconds) { Hash.new }
+  end
 
   def find_meter_counters(metric_capture_module, resource_filter, metadata_filter, log_header)
-    counters = list_resource_meters(resource_filter, log_header) + list_metadata_meters(metadata_filter, log_header)
+    counters = []
+    if self.class.counters_by_vm.dig(ems.id, target.ems_ref).nil?
```
agrare, line 103: We probably should use the cache if it is present
```
counters = self.class.counters_by_vm.dig(ems.id, target.ems_ref)
if counters.nil?
  counters = list_resource_meters(resource_filter, log_header) + list_metadata_meters(metadata_filter, log_header)
  self.class.counters_by_vm.store_path(ems.id, target.ems_ref, counters)
end
```

app/models/manageiq/providers/openstack/base_metrics_capture.rb
```diff
@@ -92,9 +93,17 @@ def list_meters(filter)
   def meter_names(metric_capture_module)
     @meter_names ||= metric_capture_module::COUNTER_NAMES
   end
+  
+  included do
+    cache_with_timeout(:counters_by_vm, 30.seconds) { Hash.new }
```
agrare, line 98: If these counter definitions don't change often we should cache them longer than 30 seconds to get the most benefit out of the cache.  We cache the vmware counter definitions for 1 hour for example.

agrare:
Looks good to me, I want to test this on an openstack system that is returning values for these just to confirm

simaishi:
Jansa backport details:

```
$ git log -1
commit 866a3d28183058fc70473d34dd1a571922aa3d0d
Date:   Thu Jul 23 08:16:23 2020 -0400

    Merge pull request #600 from andyvesel/metrics_api_call_caching

    Cache metrics api call

    (cherry picked from commit cac8776a4ebf03754e88e777cc1a7a0bf5aae5b6)
```

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 48d95750cf1b9e1a3e4472b35f2d5c253fba1ba2
Date:   Thu Jul 23 08:16:23 2020 -0400

    Merge pull request #600 from andyvesel/metrics_api_call_caching

    Cache metrics api call

    (cherry picked from commit cac8776a4ebf03754e88e777cc1a7a0bf5aae5b6)

    https://bugzilla.redhat.com/show_bug.cgi?id=1840945
```

simaishi:
@andyvesel Please take a look at Travis errors in ivanchuk branch:
https://travis-ci.com/github/ManageIQ/manageiq-providers-openstack/builds/180533999

FYI, this PR did conflict during backport due to https://github.com/ManageIQ/manageiq-providers-openstack/pull/547 not in ivanchuk branch.

simaishi:
@andyvesel ping ^

simaishi:
Reverted from ivanchuk:

```
commit b3b229220e79921e1e885f54c943bc0e72072e7b
Date:   Tue Sep 8 11:21:59 2020 -0400

    Revert "Merge pull request #600 from andyvesel/metrics_api_call_caching"

    This reverts commit 48d95750cf1b9e1a3e4472b35f2d5c253fba1ba2.

    https://bugzilla.redhat.com/show_bug.cgi?id=1840945
```

simaishi:
Backported to ivanchuk via https://github.com/ManageIQ/manageiq-providers-openstack/pull/633

Issue #601, opened 2020-05-29
qpid_proton is being forced to be required


Fryguy:
qpid_proton is an optional depedency in core, and if someone doesn't have it installed, the current code fails when doing this line:

https://github.com/ManageIQ/manageiq-providers-openstack/blob/daa0965840695d6e9b47a177d66949c99158148c/lib/manageiq/providers/openstack/legacy/openstack_event_monitor.rb#L130-L131

which in turn loads all of:

https://github.com/ManageIQ/manageiq-providers-openstack/blob/daa0965840695d6e9b47a177d66949c99158148c/lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb#L6

https://github.com/ManageIQ/manageiq-providers-openstack/blob/daa0965840695d6e9b47a177d66949c99158148c/lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_receiver.rb#L1

https://github.com/ManageIQ/manageiq-providers-openstack/blob/daa0965840695d6e9b47a177d66949c99158148c/lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_test_receiver.rb#L1

Need to delay load qpid_proton on first use (and probably have a clean exception when we detect it's not installed).

Fryguy:
cc @aufi @agrare

PR #602, opened 2020-05-29
Delay load qpid_proton via extracted method

chessbyte:
Fixes https://github.com/ManageIQ/manageiq-providers-openstack/issues/601

### DIFF
--- a/lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
+++ b/lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor.rb
@@ -1,8 +1,6 @@
 require 'manageiq/providers/openstack/legacy/openstack_event_monitor'
 require 'manageiq/providers/openstack/legacy/events/openstack_event'
 require 'manageiq/providers/openstack/legacy/events/openstack_stf_event_converter'
-require 'manageiq/providers/openstack/legacy/events/openstack_stf_event_receiver'
-require 'manageiq/providers/openstack/legacy/events/openstack_stf_event_test_receiver'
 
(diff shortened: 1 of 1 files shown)


Fryguy:
There are 3 places that require qpid_proton and they'd all have to be changed...I updated the issue with them.

Fryguy:
Though I think maybe you can just remove the requires in those files as I see nothing qpid_proton specific (unless specs need it?)

Fryguy:
wonder if we can test this somehow... like a cross repo test, but somehow set BUNDLE_WITHOUT qpid-proton ?

chessbyte:
The [OpenStackStfEventReceiver](https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/lib/manageiq/providers/openstack/legacy/events/openstack_stf_event_receiver.rb) class inherits from `Qpid::Proton::MessagingHandler`, so probably needs the require (although maybe it can be narrower).

agrare:
If you delay require these two:
```
require 'manageiq/providers/openstack/legacy/events/openstack_stf_event_receiver'
require 'manageiq/providers/openstack/legacy/events/openstack_stf_event_test_receiver'
```
that looks like it fixes this

chessbyte:
@agrare should I do the same delay load for the [bunny](https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/lib/manageiq/providers/openstack/legacy/events/openstack_rabbit_event_monitor.rb#L3) gem?

agrare:
@chessbyte Lets leave it for now so we can get this fixed

aufi:
Thanks for the PR :+1:

PR #613, opened 2020-07-06
Move supports :create to Openstack provider repo

agrare:
Currently the supports :create/:delete features are enabled at the base Flavor model, however only Openstack implements these operations

Cross Repo Tests: https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/149
Dependent PR: https://github.com/ManageIQ/manageiq/pull/20332

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/flavor.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/flavor.rb
@@ -1,6 +1,9 @@
 class ManageIQ::Providers::Openstack::CloudManager::Flavor < ::Flavor
   include ManageIQ::Providers::Openstack::HelperMethods
 
+  supports :create
+  supports :delete
+
   def self.raw_create_flavor(ext_management_system, create_options)
     ext_management_system.with_provider_connection({:service => 'Compute'}) do |service|
       cloud_tenant_refs = create_options.delete("cloud_tenant_refs")


app/models/manageiq/providers/openstack/cloud_manager/flavor.rb
```diff
@@ -1,6 +1,11 @@
 class ManageIQ::Providers::Openstack::CloudManager::Flavor < ::Flavor
   include ManageIQ::Providers::Openstack::HelperMethods
 
+  supports :create
+  supports :delete do
+    unsupported_reason_add(:delete, _("The Flavor is not connected to an active Provider")) if ext_management_system.nil?
```
chessbyte, line 6: shouldn't it be checking whether `ext_management_system.nil?` on `:create` as well?

agrare: > shouldn't it be checking whether ext_management_system.nil? on :create as well

There is no flavor instance to check on create since we're trying to create it.  The supports feature blocks are only invoked on the instance level not the class level.

>  in fact, the original code in this file seems to have been checking only on :create and not on :delete

Yeah I don't think those validations make any sense (see https://github.com/ManageIQ/manageiq/pull/20332#issuecomment-654343352).  We're validating that when we create a flavor that it has an EMS but it isn't created yet.

agrare: Going to simplify this for this bugfix and leave the validates_* methods and just move the `supports *` calls

simaishi:
Jansa backport details:

```
$ git log -1
commit c5b6c0ac22d25203c3d0969dd9e0e7cb90140661
Date:   Tue Jul 7 12:23:58 2020 -0400

    Merge pull request #613 from agrare/move_supports_create_flavor

    Move supports :create to Openstack provider repo

    (cherry picked from commit 78cdb0b1641dc1e1d3a19cfcd07cddbaed7bcf9e)
```

PR #614, opened 2020-07-08
Parse oslo.message in all events

Fryguy:
@aufi @agrare Please review.

Some notes:

- This only handles the events in the CloudManager.  I was going to bump this higher as there's an EventTargetParser for the other managers as well, but I didn't know if that made sense.  If you think I should, I'll add to this PR.
- I didn't reuse the message_content method on purpose, because it does an EMS lookup.  Ultimately, I'd like to remove that lookup and just always process `oslo.message` if it's there, because a hash lookup is relatively cheap compared to an EMS lookup and if `oslo.message` is present it needs to be parsed anyway.  Since this would require a signature change, I opted for not doing that in this PR, and can do that in a follow up.  I did, however put the message_payload method next to the message_content method so that the oslo.message code would stay close together.

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/event_parser.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/event_parser.rb
@@ -36,4 +36,20 @@ module ManageIQ::Providers::Openstack::CloudManager::EventParser
     end
     event[:content]
   end
+
+  def self.extract_content(ems_event)
+    if (oslo_message = ems_event.full_data.fetch_path(:content, 'oslo.message'))
+      begin
+        JSON.parse(oslo_message)
+      rescue JSON::ParserError
+        {}
+      end
+    else
+      ems_event.full_data.fetch(:content, {})
+    end
+  end
+
+  def self.extract_payload(ems_event)
+    extract_content(ems_event).fetch('payload', {})
+  end
 end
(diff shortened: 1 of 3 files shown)


Fryguy:
Oh also, I'd like to add tests for this particular case, but I'm not sure how?  Any suggestions?

agrare:
@aufi can these `oslo.message` contents show up for the network and storage type events as well?

agrare:
@Fryguy you could add a test to the `spec/models/manageiq/providers/openstack/cloud_manager/event_target_parser_spec.rb` that includes an oslo.message in the full_data

Fryguy:
@agrare Oh nice...I hadn't seen that file and was expecting to find a VCR spec...yeah I can add these.

Fryguy:
Updated with specs...basically I changed the specs to run _all_ of them with and without the oslo.message format.  It's easier to review with whitespace removed.

agrare:
:+1: nice, still not sure if we need the oslo message parsing on the network and storage managers also (@aufi or @andyvesel ?) but given that the `parse_oslo_message` was only in the `CloudManager::EventTargetParser` I think it is fine to only do this for the cloud manager and if we want to we can do a follow-up to add this to the others

simaishi:
Jansa backport details:

```
$ git log -1
commit acfa4890b80bf62250cd0eec096fe5c295c1f57e
Date:   Wed Jul 8 12:35:54 2020 -0400

    Merge pull request #614 from Fryguy/oslo_message

    Parse oslo.message in all events

    (cherry picked from commit 21fb383db6731f0dccb1e884ab8ad581b830d703)
```

PR #616, opened 2020-07-08
Consolidate event parsing

Fryguy:
Consolidates the duplicated event parsing and oslo.message handling for all EventParsers and EventTargetParsers.

- Removes the unnecessary lookup and check of the ems and it's "amqp" role
- Handles oslo.message on events coming from anywhere, so tests were added for NetworkManager and CinderManager EventTargetParsers
- Makes it clearer that the NetworkManager and CinderManager event parser are exactly the same as the CloudManager

@agrare Please review.

### DIFF
--- a/app/models/manageiq/providers/openstack/infra_manager/event_parser.rb
+++ b/app/models/manageiq/providers/openstack/infra_manager/event_parser.rb
@@ -1,43 +1,14 @@
 module ManageIQ::Providers::Openstack::InfraManager::EventParser
-  def self.event_to_hash(event, ems_id)
-    content = message_content(event, ems_id)
-    event_type = content["event_type"]
-    payload = content["payload"] || {}
-
-    log_header = "ems_id: [#{ems_id}] " unless ems_id.nil?
-    _log.debug("#{log_header}event: [#{event_type}]") if $log && $log.debug?
-
-    # attributes that are common to all notifications
-    event_hash = {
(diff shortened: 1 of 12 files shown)


Fryguy:
Marking it jansa/no because it's not required for backport, being a refactoring cleanup.

app/models/manageiq/providers/openstack/infra_manager/event_parser.rb
```diff
@@ -1,43 +1,14 @@
...
-    unless ems_id.nil?
-      ems = ExtManagementSystem.find_by_id(ems_id)
-      if ems.connection_configuration_by_role("amqp")
-        if event[:content].key?("oslo.message")
-          return JSON.parse(event[:content]["oslo.message"])
-        end
+ def self.event_to_hash(event, ems_id)
+    ManageIQ::Providers::Openstack::EventParserCommon.event_to_hash(event, ems_id) do |event_hash, payload|
+      if payload.key? "instance_id"
+        event_hash[:host_id] = Host.find_by(:uid_ems => payload["instance_id"]).try(:id)
```
agrare, line 5: Not related to this change but we should setting the `event_hash[:host_uid_ems]` and then doing this lookup in the event handler worker like we do with the [vm uid_ems](https://github.com/ManageIQ/manageiq/blob/master/app/models/ems_event.rb#L80)

PR #617, opened 2020-07-09
Use event host_uid_ems instead of doing a db query

agrare:
Ideally the EventCatcher shouldn't hit the database to do queries, linking up events to inventory should be done by the event handler.

Depends on: https://github.com/ManageIQ/manageiq/pull/20348


agrare:
cc @Fryguy

PR #633, opened 2020-09-08
[IVANCHUK] Cache metrics api call

agrare:
Backport of https://github.com/ManageIQ/manageiq-providers-openstack/pull/600

### DIFF
--- a/app/models/manageiq/providers/openstack/base_metrics_capture.rb
+++ b/app/models/manageiq/providers/openstack/base_metrics_capture.rb
@@ -92,30 +92,34 @@ class ManageIQ::Providers::Openstack::BaseMetricsCapture < ManageIQ::Providers::
   def meter_names(metric_capture_module)
     @meter_names ||= metric_capture_module::COUNTER_NAMES
   end
+  
+  cache_with_timeout(:counters_by_vm, 60.minutes) { Hash.new }
 
   def find_meter_counters(metric_capture_module, resource_filter, metadata_filter, log_header)
-    counters = list_resource_meters(resource_filter, log_header) + list_metadata_meters(metadata_filter, log_header)
-
-    # With Gnocchi, the network metrics are not associated with the instance's resource id
-    # but with the instance's network interface resource id. Here we fetch the counters
-    # for the network interface, so that the network metrics can be fetched.
-    if target.respond_to?(:network_ports)
-      target.network_ports.each do |port|
-        # fetch the list of resources and use the original_resource_id and type to find
-        # the network interface's resource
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/openstack/base_metrics_capture.rb
```diff
@@ -92,30 +92,34 @@ def list_meters(filter)
   def meter_names(metric_capture_module)
     @meter_names ||= metric_capture_module::COUNTER_NAMES
   end
+  
+  cache_with_timeout(:counters_by_vm, 60.minutes) { Hash.new }
```
agrare, line 96: I just moved this out of the `included do; end` block and removed the `extend ActiveSupport::Concern` because this is a class not a module on ivanchuk

app/models/manageiq/providers/openstack/base_metrics_capture.rb
```diff
@@ -92,30 +92,34 @@ def list_meters(filter)
...
-        # fetch the list of resources and use the original_resource_id and type to find
-        # the network interface's resource
-        if port.ems_ref
-          original_resource_id = "#{target.ems_ref}-tap#{port.ems_ref[0..10]}"
-          resources = @perf_ems.list_resources("instance_network_interface").body
-          resources.each do |r|
-            if r["original_resource_id"].include?(original_resource_id)
-              resource_filter = {"field" => "resource_id", "value" => r["id"]}
-              counters = counters + list_resource_meters(resource_filter, log_header)
+    counters = self.class.counters_by_vm.dig(target.ems_id, target.ems_ref)
```
agrare, line 99: This was `ems.id` but since on ivanchuk we don't have an `@ems` ivar I just used the `target.ems_id`

app/models/manageiq/providers/openstack/base_metrics_capture.rb
```diff
@@ -92,30 +92,34 @@ def list_meters(filter)
...
+            resources.each do |r|
+              if r["type"].to_s == "instance_network_interface" && r["original_resource_id"].include?(original_resource_id)
+                resource_filter = {"field" => "resource_id", "value" => r["id"]}
+                counters = counters + list_resource_meters(resource_filter, log_header)
+              end
             end
           end
         end
       end
+      self.class.counters_by_vm.store_path(target.ems_id, target.ems_ref, counters)
```
agrare, line 121: Same here

simaishi:
@gtanzillo can you please review? I'd like to include in today's build.

Fryguy (approved):

PR #634, opened 2020-09-08
Fix bug where new providers are not reflected in the list

Fryguy:
Since we were memoizing a dynamic value, subsequent changes were not reflected.

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager.rb
@@ -55,8 +55,20 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
   before_update :ensure_managers_zone_and_provider_region
   after_save :refresh_parent_infra_manager
 
+  private_class_method def self.provider_id_options
+    (t = ManageIQ::Providers::Openstack::Provider)
+      .order(t.arel_table[:name].lower)
+      .pluck(:name, :id)
+      .map do |name, id|
+        {
+          :label => name,
+          :value => id.to_s,
+        }
+      end
+  end
+
   def self.params_for_create
-    @params_for_create ||= {
+    {
       :fields => [
         {
           :component => "text-field",
@@ -70,12 +82,7 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
           :name        => "provider_id",
           :label       => _("Openstack Infra Provider"),
           :isClearable => true,
-          :options     => ManageIQ::Providers::Openstack::Provider.pluck(:name, :id).map do |name, id|
-            {
-              :label => name,
-              :value => id.to_s,
-            }
-          end
+          :options     => provider_id_options
         },
         {
           :component    => "select",
--- a/app/models/manageiq/providers/openstack/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager.rb
@@ -56,8 +56,9 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
   after_save :refresh_parent_infra_manager
 
   private_class_method def self.provider_id_options
-    (t = ManageIQ::Providers::Openstack::Provider)
-      .order(t.arel_table[:name].lower)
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/openstack/cloud_manager.rb
```diff
@@ -56,7 +56,25 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
...
   def self.params_for_create
-    @params_for_create ||= {
+    static_params_for_create.deep_clone.tap do |params|
+      params[:fields].detect { |f| f[:id] == 'provider_id' }[:options] = provider_id_options
+    end
+  end
+
+  private_class_method def self.provider_id_options
+    (t = ManageIQ::Providers::Openstack::Provider)
+      .order(t.arel_table[:name].lower)
```
Fryguy, line 66: @agrare As discussed, here is the extra fix to order by name.  I think there are also needs to be an RBAC tenancy check here.  Thoughts on how best to do this?  Is it just `RBAC.filtered(...)`?

agrare: Yeah `Rbac.filtered` should do it

Fryguy:
@agrare Updated with the RBAC changes.

spec/models/manageiq/providers/openstack/cloud_manager_spec.rb
```diff
@@ -18,6 +18,54 @@ def lastlog
...
+    it "filters out providers based on RBAC" do
+      Tenant.seed
+      EvmSpecHelper.local_miq_server
+
+      tenant1 = FactoryBot.create(:tenant)
+      group1  = FactoryBot.create(:miq_group, :tenant => tenant1)
+      user1   = FactoryBot.create(:user, :miq_groups => [group1])
+      tenant2 = FactoryBot.create(:tenant)
+      group2  = FactoryBot.create(:miq_group, :tenant => tenant2)
+      user2   = FactoryBot.create(:user, :miq_groups => [group2])
```
Fryguy, line 50: @kbrock Is there a nicer way to do all this?  Feels like a common pattern, but `FactoryBot.create(:user, :tenant => tenant1)` didn't work like I expected.

agrare (approved):
Tried it out locally, LGTM :+1:

agrare:
@Fryguy you want to wait to see if there is a cleaner way to setup the spec or you good to merge as is?

Fryguy:
Nah we can merge as is.  If there's a cleaner way we can do a follow-up.

spec/models/manageiq/providers/openstack/cloud_manager_spec.rb
kbrock, line 50:
```suggestion
user1 = FactoryBot.create(:user_with_group, :tenant => FactoryBot.create(:tenant))
user2 = FactoryBot.create(:user_with_group, :tenant => FactoryBot.create(:tenant))
```

spec/models/manageiq/providers/openstack/cloud_manager_spec.rb
```diff
@@ -18,6 +18,54 @@ def lastlog
...
+
+      provider1 = FactoryBot.create(:provider_openstack, :tenant => tenant1)
+      provider2 = FactoryBot.create(:provider_openstack, :tenant => tenant2)
+
+      User.current_user = user1
+      options = DDF.find_field(described_class.params_for_create, "provider_id")[:options]
+      expect(options).to eq [
+        {:label => provider1.name, :value => provider1.id.to_s}
+      ]
```
kbrock, line 59: you may also want to try `with_user`, though that won't buy you much other than a cleaner declaration.
``` suggestion
User.with_user(user1 ) do
        options = DDF.find_field(described_class.params_for_create, "provider_id")[:options]
        expect(options).to eq [
          {:label => provider1.name, :value => provider1.id.to_s}
        ]
end
```

kbrock (commented):
this looks fine.

Just commenting because you asked...

If you don't care about the tenant in particular, then just a simple `:user_with_group` is your friend.

agrare:
We do not use DDF forms for provider creation on jansa so this isn't necessary

PR #636, opened 2020-09-17
Plugin Generator Updates

Fryguy:
Plugin Generator Updates

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -3,20 +3,22 @@ if ENV['CI']
   SimpleCov.start
 end
 
+Dir[Rails.root.join("spec/shared/**/*.rb")].each { |f| require f }
+Dir[File.join(__dir__, "support/**/*.rb")].each { |f| require f }
+
+require "manageiq-providers-openstack"
+
 VCR.configure do |config|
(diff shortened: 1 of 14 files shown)


spec/spec_helper.rb
```diff
@@ -3,20 +3,22 @@
...
+Dir[File.join(__dir__, "support/**/*.rb")].each { |f| require f }
+
+require "manageiq-providers-openstack"
+
 VCR.configure do |config|
   config.ignore_hosts 'codeclimate.com' if ENV['CI']
   config.cassette_library_dir = File.join(ManageIQ::Providers::Openstack::Engine.root, 'spec/vcr_cassettes')
 end
 
-NotificationType.seed
```
Fryguy, line 11: So this is interesting.  For now, I added this as a before(:suite) to be more accurate and to not pollute things globally.  Additionally, I added an after(:suite) to tear it all down.

agrare: This looks like it should be in the `before {}` of specific tests, not the whole suite...should be pretty easy to see which specs need it just run without it and see which ones fail.

agrare: I'll push a follow-up PR

agrare (approved):

PR #642, opened 2020-10-03
Queue name for ems no longer an array

agrare:
https://github.com/ManageIQ/manageiq/pull/20345 dropped combined child
manager queue names


andyvesel (commented):
Apart from Travis, LGTM :)

agrare:
Thanks @andyvesel you're right I missed the infra_manager one!

PR #645, opened 2020-10-12
Move VcrRecorder closer to the rake task

Fryguy:
The VcrRecorder is specific to the rake task and should not be in the general lib directory.

@agrare Please review.


andyvesel (commented):
lgtm, with fixed Travis :)

Fryguy:
That's weird...git did a delete instead of a rename 😮 ...incoming fix.

Fryguy:
@andyvesel updated

agrare (approved):

PR #646, opened 2020-10-12
Check qpid_proton installation on local machines before running tests

### LINKED ISSUE #643: Skip qpid_proton specs if not installed
Running the specs locally I get: `An error occurred while loading ./spec/legacy/events/openstack_stf_event_monitor_spec.rb` because it requires qpid_proton.

Similar to https://github.com/ManageIQ/manageiq-providers-openstack/pull/602, the test suite should skip/filter these if qpid_proton is not installed.

Fryguy:
Fixes #643

On CI, enforces the tests must run; on local dev will skip the tests unless qpid_proton is installed.  Either way, one can choose to skip the tests using rspec tags (i.e. `SPEC_OPTS="--tag ~qpid_proton"`)

### DIFF
--- a/spec/legacy/events/openstack_stf_event_monitor_spec.rb
+++ b/spec/legacy/events/openstack_stf_event_monitor_spec.rb
@@ -1,9 +1,18 @@
 require 'manageiq/providers/openstack/legacy/events/openstack_stf_event_monitor'
-require 'qpid_proton'
 
-describe OpenstackStfEventMonitor do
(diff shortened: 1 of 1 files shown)


agrare (approved):
:+1: nice

djberg96:
Dang, wish I had seen this first, woulda suggested using a `config.filter_run_excluding` in the spec_helper.rb file instead.

Fryguy:
You can still make a PR to do that.

PR #647, opened 2020-10-13
Update qpid_proton spec handling using exclude filter

djberg96:
Followup to https://github.com/ManageIQ/manageiq-providers-openstack/pull/646, this uses an exclude filter in the `spec_helper.rb` to skip `qpid_proton` tags if the `qpid_proton` gem is not loaded.

This is a little cleaner, and will work for any other spec examples that might eventually get tagged.

If you're wondering why I have `require 'qpid_proton'` in a `before(:all)` block, it's because the filter works on examples. If you remove that then it will still try to load `qpid_proton` and again we're back to failure.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -8,6 +8,10 @@ Dir[File.join(__dir__, "support/**/*.rb")].each { |f| require f }
 
 require "manageiq-providers-openstack"
 
+RSpec.configure do |config|
+  config.filter_run_excluding(:qpid_proton) unless Gem.loaded_specs.has_key?(:qpid_proton)
+end
+
 VCR.configure do |config|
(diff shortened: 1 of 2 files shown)


spec/spec_helper.rb
```diff
@@ -8,6 +8,10 @@
 
 require "manageiq-providers-openstack"
 
+RSpec.configure do |config|
+  config.filter_run_excluding(:qpid_proton) unless Gem.loaded_specs.key?(:qpid_proton)
```
Fryguy, line 12: I think we need to at least keep the check for CI env car.  While locally I don't mind if the specs are skipped, we should never skip them on CI.

djberg96: @Fryguy ok, added.

agrare (approved):

PR #650, opened 2020-10-26
Fix StorageManager initialization

agrare:
Openstack was calling ensure_managers which builds the NetworkManager and calls ensure_managers_zone_and_provider_region which is overridden here to set the zone/region for all child managers but before the storage managers were being built.

Typically cloud_managers with child storage managers will override the `HasNetworkManagerMixin#ensure_managers` method to include initializing the child storage managers as well, see https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/cloud_manager.rb#L77-L90

https://github.com/ManageIQ/manageiq/pull/20701 appears to have kept this "working" but very non-obviously
Fixes API spec failures: https://github.com/ManageIQ/manageiq-api/pull/940

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager.rb
@@ -48,9 +48,7 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
   supports :swift_service
   supports :create_host_aggregate
 
-  before_create :ensure_managers,
-                :ensure_cinder_managers,
-                :ensure_swift_managers
+  before_create :ensure_managers
 
   before_update :ensure_managers_zone_and_provider_region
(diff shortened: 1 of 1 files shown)


andyvesel (approved):

agrare:
cc @kbrock @NickLaMuro

NickLaMuro:
@miq-bot cross-repo-tests manageiq-api

agrare:
:+1: thanks @NickLaMuro looks like cross-repo tests are green
@andyvesel @Fryguy @kbrock anyone want to merge?

agrare:
This is another case of inconsistencies causing bugs, https://github.com/ManageIQ/manageiq/pull/20701

agrare:
Thanks @andyvesel !

PR #654, opened 2020-10-27
[WIP] extend relationships to define the appropriate class names

kbrock:
This  depends upon:
- [x] https://github.com/ManageIQ/manageiq/pull/20739

putting in this PR to show where I would like to go and to run cross repo tests

use parent definition of ensure since they are now pretty generic.
They may be no longer necessary and a build_ may suffice.

Use parent definition of propagation so all attributes go across
Extend parent managers to group the functionality together

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager.rb
@@ -25,13 +25,18 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
   require_nested :Template
   require_nested :Vm
 
+  # may want to just us an array of child managers instead of this
+  # it looks cleaner having this relationship,
+  # but this will not be in sync with cinder/swift manager references
   has_many :storage_managers,
            :foreign_key => :parent_ems_id,
            :class_name  => "ManageIQ::Providers::StorageManager",
            :autosave    => true
   has_many :snapshots, :through => :vms_and_templates
+
   include ManageIQ::Providers::Openstack::CinderManagerMixin
-  include SwiftManagerMixin
+  include ManageIQ::Providers::Openstack::SwiftManagerMixin
+  include ManageIQ::Providers::Openstack::NetworkManagerMixin
   include ManageIQ::Providers::Openstack::ManagerMixin
   include ManageIQ::Providers::Openstack::IdentitySyncMixin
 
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/openstack/cloud_manager.rb
```diff
@@ -49,9 +54,12 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
   supports :swift_service
   supports :create_host_aggregate
 
+  # TODO: move to after_initialization
   before_create :ensure_managers
 
+  # TODO: move to before_validation
   before_update :ensure_managers_zone_and_provider_region
```
agrare, line 61: One thing to keep in mind, the tenant_id is set also in a `before_validation` callback so we need to make sure those play nicely together.  Don't want to be setting the child.tenant_id = tenant_id before tenant_id has been set

https://github.com/ManageIQ/manageiq/blob/master/app/models/mixins/tenancy_mixin.rb#L5

app/models/manageiq/providers/openstack/cloud_manager.rb
```diff
@@ -419,52 +427,19 @@ def ensure_managers
     ensure_network_manager
     ensure_cinder_manager
     ensure_swift_manager
+    # TODO: remove when this moves to before_initialization
```
agrare, line 430: Do you mean before_validation?

app/models/manageiq/providers/openstack/cloud_manager.rb
```diff
@@ -419,52 +427,19 @@ def ensure_managers
...
-
-  after_save :save_on_other_managers
-
-  def save_on_other_managers
-    storage_managers.update_all(:tenant_mapping_enabled => tenant_mapping_enabled)
-    if network_manager
-      network_manager.tenant_mapping_enabled = tenant_mapping_enabled
-      network_manager.save!
+  def ensure_managers_zone_and_provider_region
+    child_manager_references.each do |child_manager|
```
agrare, line 440: If we could work out the list of child managers generically we could do this in core right?

app/models/manageiq/providers/openstack/cloud_manager.rb
kbrock: I think we should create the objects at initialization time - empty vessels
then in before validation, make sure they have the correct values

I seem to remember that rails could auto build the `has_one`  for us but can't find the references anymore.

But maybe before validation is where we will create these objects too

app/models/manageiq/providers/openstack/cloud_manager.rb
kbrock: I had child_managers but that messed with an already defined child_managers

here, we want to access the actual child managers that we are using.
If we want to stick with the child_managers association, then we could define swift_manager to be a select into child_managers or something like that?

agrare:
Hey @kbrock now that https://github.com/ManageIQ/manageiq/pull/20739 is merged can we get this in?

Fryguy:
#673 was merged moving the SwiftManager from core to this plugin, which also made the class names more consistent.

PR #660, opened 2020-10-30
[WIP]Fix incorrect metrics name

andyvesel:
According to https://bugzilla.redhat.com/show_bug.cgi?id=1872312#c9 CloudForms attempts to collect disk IO metrics from Gnocchi but holds the wrong name for it. In Gnocchi it's `disk.device.read.bytes` and `disk.device.write.bytes`


aufi:
This would break metrics for non-Ceph OSP deployments, the original metric names should stay in place too.

There could be needed pair disk.device.* metrics with its Instance manually (since such metrics are reported per disk device, not per an Instance). It could look similar to network interfaces metrics https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/base_metrics_capture.rb#L108 (check git blame to find Richard's PR adding a similar logic for networks).

PR #663, opened 2020-11-04
Rails 6.0

NickLaMuro:
Links
-----

* Epic:  https://github.com/ManageIQ/manageiq/issues/19977
* Current Cross Repo: https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/179

### DIFF
--- a/manageiq-providers-openstack.gemspec
+++ b/manageiq-providers-openstack.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_dependency "activesupport",        "~> 5.2.4", ">= 5.2.4.3"
+  spec.add_dependency "activesupport",        "~> 6.0"
   spec.add_dependency "bunny",                "~> 2.1.0"
   spec.add_dependency "excon",                "~> 0.71"
   spec.add_dependency "fog-openstack",        ">= 0.3.10"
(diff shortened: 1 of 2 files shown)


manageiq-providers-openstack.gemspec
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_dependency "activesupport",        "~> 5.2.4", ">= 5.2.4.3"
```
bdunne, line 21: Can we apply https://github.com/ManageIQ/activerecord-id_regions/pull/18/files#r537735807 here as well?

NickLaMuro: I don't see why we would, as this isn't a gem that ever needs to be released, and is only used by `ManageIQ/manageiq`, which now requires Rails 6.  The only place where I think this does make sense is in gems that are released to rubygems.org or rubygems.manageiq.org which this is not.

(This is going to be a carbon copy answer that I will be pasting in all provider repos where this comment exists)

bdunne: I still look at the provider plugins like they are regular gems and question why a gem would have a tight dependency like this unless something in the code base requires that tight dependency.

NickLaMuro: If this was a more broken out plugin, where customers could add gems as they see fit, we would probably have a `manageiq-core` gem that plugins would instead be consuming, which would probably be tightly coupled to a version of Rails.

I see this dependency as a gemspec version of that for now, where in reality, it really is being done as part of `bin/update` and the dependency of `spec/manageiq` being required to run anything in this repo standalone at the moment.

PR #644, opened 2020-10-12
Declared DDF schema for cloud volume creation and editing

skateman:
This is an initial Cloud Volume form schema for the fields that are specific to the ibm-cloud-provided cloud volumes.

@miq-bot add_label enhancement
@miq-bot add_reviewer @agrare

Parent issue: https://github.com/ManageIQ/manageiq-ui-classic/issues/7334

PR #664, opened 2020-11-05
Move OpenStack CloudVolume* from CloudManager to CinderManager

agrare:
These models belong to the Openstack::StorageManager::CinderManager but their class name indicates that they belong to the Openstack::CloudManager which is quite confusing.

Schema migration: https://github.com/ManageIQ/manageiq-schema/pull/532
Depends on: https://github.com/ManageIQ/manageiq/pull/20782
Ref: https://github.com/ManageIQ/manageiq-providers-openstack/pull/644#issuecomment-722281752


skateman:
Required for https://github.com/ManageIQ/manageiq-ui-classic/issues/7334

PR #669, opened 2020-12-16
Use log_start_user_event_message in Scanning::Job

djberg96:
Originally this started as a linter warning, but it appears to be a bug. The original warning was that the `user_event` variable wasn't being used. However, it also revealed that `start_user_event_message` doesn't actually do anything, it's just a string. Instead, I'm pretty sure we want `log_start_user_event_message`.

See https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_scan.rb#L242-L256 for more info.

It's also up for debate as to whether or not we should also be calling a corresponding `log_end_user_event_message` which is only called in a different method.


agrare:
So this was split up from core in an attempt to get rid of the provider specific logic in the base VmScan model.  If I had to guess `user_event` was probably set in a number of different conditional paths and then in one place logged.  When we split it up it was still being set in the conditional branch that we extracted but we never copied the common log_user_event call.

agrare (approved):

PR #670, opened 2020-12-17
Fix linter warnings

djberg96:
This PR addresses the linter warnings that rubocop found. Specifically:

* Lint/RescueException is disabled, since changing those could lead to breakage.
* Unused variables are underscored.
* A couple singleton methods were moved out of a private block since it has no effect.
* The shadowed `Timeout::Error` exception blocks are moved up. See https://github.com/ManageIQ/manageiq-providers-vmware/pull/682 for similar fixes and more detail.
* One stylistic change regarding assignment that looks like a conditional.


djberg96:
Updated to just remove the `ems_event` unused argument completely.

app/models/manageiq/providers/openstack/storage_manager/cinder_manager/cloud_volume.rb
```diff
@@ -227,6 +227,13 @@ def with_provider_connection
     super(cinder_connection_options)
   end
 
+  def self.cinder_connection_options(cloud_tenant = nil)
+    connection_options = {:service => "Volume"}
+    connection_options[:tenant_name] = cloud_tenant.name if cloud_tenant
+    connection_options[:proxy] = openstack_proxy if openstack_proxy
+    connection_options
+  end
+
```
agrare, line 236: Why make this public rather than adding `private_class_method` ?  It looks like the intent was for this to be private given that it is references from a private instance method.

app/models/manageiq/providers/openstack/cloud_manager/scanning/job.rb
```diff
@@ -30,17 +30,16 @@ def call_snapshot_create
...
     rescue Timeout::Error
       msg = case options[:snapshot]
             when :smartProxy, :skipped then "Request to log snapshot user event with EMS timed out."
             else "Request to create snapshot timed out"
             end
       _log.error(msg)
       signal(:abort, msg, "error")
+    rescue => err
+      _log.log_backtrace(err)
+      signal(:abort, err.message, "error")
```
Fryguy, line 42: This is a nice find...this is actually a bug fix (albeit a minor one).  In old versions, Timeout::Error did not derive from StandardError, but now it does.

app/models/manageiq/providers/openstack/storage_manager/cinder_manager/cloud_volume.rb
Fryguy: Unfortunately if it's made a private_class_method, then line 253 won't work.  As far as I'm aware there's no way to make a `protected_class_method`, which is effectively what we want.  That being said, I don't like moving this up in the file as the solution, as it was clearer when in locality to the caller.

app/models/manageiq/providers/openstack/cloud_manager/scanning/job.rb
djberg96: I wondered about that. I'm not sure when that changed.

app/models/manageiq/providers/openstack/storage_manager/cinder_manager/cloud_volume.rb
Fryguy: I wonder if it's almost better to just make it a private instance method.

.rubocop_local.yml
```diff
@@ -1,3 +1,5 @@
 Style/GlobalVars:
   AllowedVariables:
     - $fog_log
+Lint/RescueException:
+  Enabled: false
```
agrare, line 5: We should probably leave this up to the core manageiq-style (discussion ongoing over in https://github.com/ManageIQ/manageiq-style/pull/15) rather than override it here

app/models/manageiq/providers/openstack/storage_manager/cinder_manager/cloud_volume.rb
djberg96: I've updated it to use `send` internally so that 253 continues to work. I tried to just get rid of the singleton method, but that caused spec failures, and I didn't want risk breakage elsewhere.

.rubocop_local.yml
djberg96: Ok, removed it.

agrare:
> Remove local cop, wait for vote. 

This commit message probably isn't helpful to us in a years time, maybe rebase the commit that introduced it?

djberg96:
@agrare ok, squashed/fixed commits.

djberg96:
@agrare, @Fryguy Look ok now?

app/models/manageiq/providers/openstack/storage_manager/cinder_manager/cloud_volume.rb
agrare: I don't think making something private then changing its caller to use `.send()` is right and it can't be just a private instance method because `.raw_create_volume` also needs it as a class method.

@Fryguy it seems like duplicating this method at the instance and class level (:-1:) or leaving it as a public class method seem like the only options

Fryguy: Or just remove it from this PR, focusing on syntactical changes, and doing the semantic changes in a separate PR.

kbrock:
kicking to get it recent. so we can merge

kbrock:
ugh. kicking no work so good. sorry.
Will put back in in another PR

PR #671, opened 2020-12-20
New class for Swift Manager is created

pyon-yx:
…t Store Containers page.

- New code for Openstack swift manager is added. CloudObjectStoreContainer class is defined, and raw_cloud_object_store_container_create function is called when customer tries to add new container for openstack swift manager.

Goes with https://github.com/ManageIQ/manageiq-ui-classic/pull/7544


pyon-yx:
My code has some issues.

Fryguy:
@pyon-yx #673 was merged moving the SwiftManager from core to this plugin.

agrare:
I'm going to close this since it has been implemented in https://github.com/ManageIQ/manageiq-providers-openstack/pull/673

PR #674, opened 2021-01-11
Move environment builder from core

Fryguy:
Removed from core in ManageIQ/manageiq#20943

(from commit ManageIQ/manageiq@a26a849bdb17cc712f3f814aa0fd4d9f46c609a7)

@agrare Please review.


Fryguy:
@agrare There is also a [private key file](https://github.com/ManageIQ/manageiq/blob/master/spec/tools/openstack_ems_refresh_spec_key_pair_private) that looks like it was used for specs, but I don't see it used anywhere, so I'm gonna delete it

agrare (approved):

PR #677, opened 2021-01-14
Don't subclass from core cinder/swift managers

agrare:
Merge together with: https://github.com/ManageIQ/manageiq/pull/20959

### DIFF
--- a/app/models/manageiq/providers/openstack/cinder_manager_mixin.rb
+++ b/app/models/manageiq/providers/openstack/cinder_manager_mixin.rb
@@ -1,13 +1,12 @@
 module ManageIQ::Providers::Openstack::CinderManagerMixin
   extend ActiveSupport::Concern
-  include ::CinderManagerMixin
 
   included do
     # TODO: how about many storage managers???
     # Should use has_many :storage_managers,
     has_one  :cinder_manager,
              :foreign_key => :parent_ems_id,
-             :class_name  => "ManageIQ::Providers::StorageManager::CinderManager",
+             :class_name  => "ManageIQ::Providers::Openstack::StorageManager::CinderManager",
              :autosave    => true
 
     delegate :cloud_volumes,
--- a/app/models/manageiq/providers/openstack/storage_manager/cinder_manager.rb
+++ b/app/models/manageiq/providers/openstack/storage_manager/cinder_manager.rb
@@ -1,4 +1,4 @@
-class ManageIQ::Providers::Openstack::StorageManager::CinderManager < ManageIQ::Providers::StorageManager::CinderManager
+class ManageIQ::Providers::Openstack::StorageManager::CinderManager < ManageIQ::Providers::StorageManager
   require_nested :CloudVolume
   require_nested :CloudVolumeBackup
   require_nested :CloudVolumeSnapshot
@@ -7,6 +7,7 @@ class ManageIQ::Providers::Openstack::StorageManager::CinderManager < ManageIQ::
   require_nested :EventCatcher
   require_nested :EventParser
 
+  include ManageIQ::Providers::StorageManager::BlockMixin
   include ManageIQ::Providers::Openstack::ManagerMixin
 
   supports :cinder_volume_types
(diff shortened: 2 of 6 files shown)


andyvesel (approved):

app/models/manageiq/providers/openstack/storage_manager/cinder_manager.rb
```diff
@@ -1,4 +1,4 @@
-class ManageIQ::Providers::Openstack::StorageManager::CinderManager < ManageIQ::Providers::StorageManager::CinderManager
+class ManageIQ::Providers::Openstack::StorageManager::CinderManager < ManageIQ::Providers::StorageManager
```
kbrock, line 1: darn - this class no longer matches the filename.
Is that an oversight, or do we need it to be this way for class loading?

Fryguy (approved):

app/models/manageiq/providers/openstack/storage_manager/swift_manager.rb
```diff
@@ -1,4 +1,4 @@
-class ManageIQ::Providers::Openstack::StorageManager::SwiftManager < ManageIQ::Providers::StorageManager::SwiftManager
+class ManageIQ::Providers::Openstack::StorageManager::SwiftManager < ManageIQ::Providers::StorageManager
```
kbrock, line 1: ditto

app/models/manageiq/providers/openstack/cinder_manager_mixin.rb
```diff
@@ -1,13 +1,12 @@
 module ManageIQ::Providers::Openstack::CinderManagerMixin
   extend ActiveSupport::Concern
-  include ::CinderManagerMixin
```
kbrock, line 3: I can vouch that most of this is not needed

do we need to define `ensure_cinder_manager`?
it is only called from one spot in `providers/openstack/cloud_manager.rb` -- so if you want to tweak this, it can be in this same PR

kbrock (commented):
just the 2 file name changes and that one method.

This is a great change and makes me so happy - I had a similar PR hiding somewhere locally, but I didn't go so far as to remove the hierarchy for these classes. Glad you discovered we can do this. It makes everything flow better and reduces cross repo dependencies.

app/models/manageiq/providers/openstack/storage_manager/cinder_manager.rb
kbrock: ignore me -- changed the parent class not the main class

agrare: Still looks like it matches the filename to me...am I missing something?

app/models/manageiq/providers/openstack/storage_manager/swift_manager.rb
kbrock: and ditto on the oops

app/models/manageiq/providers/openstack/cinder_manager_mixin.rb
agrare: Yeah so the `ManageIQ::Providers::Openstack::CloudManager` was overriding `ensure_cinder_manager` from the core `CinderManagerMixin` anyway, I didn't think it was necessary to pull it out of there and into this mixin when it was already with the rest of the `ensure_*_manager` methods.

kbrock: and the last one is good - carry on

kbrock (approved):
This is great - please merge

app/models/manageiq/providers/openstack/cinder_manager_mixin.rb
agrare: Good catch on the :type, updated https://github.com/ManageIQ/manageiq-providers-openstack/pull/677/files#diff-6e82d48dbcba5ccd5c9dee5a5f722c52485f2c738505fe1698d54d9786e69fa0L452

PR #684, opened 2021-01-25
Flatten worker settings

Fryguy:
Depends on ManageIQ/manageiq#20975

### DIFF
--- a/app/models/manageiq/providers/openstack/base_metrics_capture.rb
+++ b/app/models/manageiq/providers/openstack/base_metrics_capture.rb
@@ -420,7 +420,7 @@ module ManageIQ::Providers::Openstack::BaseMetricsCapture
   end
 
   def metric_service_from_settings
-    Settings[:workers][:worker_base][:queue_worker_base][:ems_metrics_collector_worker][:ems_metrics_openstack_default_service]
+    Settings.workers.ems_metrics_collector_worker.ems_metrics_openstack_default_service
   rescue StandardError => err
     $log.warn "#{_log.prefix} Settings key ems_metrics_openstack_default_service is missing, #{err}."
     nil
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -129,83 +129,89 @@
 :log:
   :level_fog: info
 :workers:
-  :worker_base:
-    :event_catcher:
-      :event_catcher_openstack:
-        :poll: 15.seconds
-        :topics:
-          :nova: notifications.*
-          :cinder: notifications.*
-          :glance: notifications.*
-          :heat: notifications.*
-        :duration: 10.seconds
-        :capacity: 50
-        :amqp_port: 5672
-        :amqp_heartbeat: 30
-        :amqp_recovery_attempts: 4
-        :amqp_vhost: '/'
-        :ceilometer:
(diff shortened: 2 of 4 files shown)


config/settings.yml
```diff
@@ -129,83 +129,89 @@
...
-        :ems_metrics_collector_worker_openstack_infra: {}
-        :ems_metrics_openstack_default_service: "auto"
-        :ems_metrics_gnocchi_granularity: 300
-      :ems_refresh_worker:
-        :ems_refresh_worker_openstack: {}
-        :ems_refresh_worker_openstack_infra: {}
-        :ems_refresh_worker_openstack_network: {}
+  :ems_metrics_collector_worker:
+    :ems_metrics_gnocchi_granularity: 300
+    :ems_metrics_openstack_default_service: "auto"
```
Fryguy, line 134: @agrare As discussed we should open a separate issue to not "infect" the base setting, but instead use the `ems.ems_openstack` section for these.

config/settings.yml
```diff
@@ -129,83 +129,89 @@
...
+  :ems_metrics_collector_worker_openstack_infra:
+    :inherits_from: ems_metrics_collector_worker
+  :ems_refresh_worker_openstack:
+    :inherits_from: ems_refresh_worker
+  :ems_refresh_worker_openstack_infra:
+    :inherits_from: ems_refresh_worker
+  :ems_refresh_worker_openstack_network:
+    :inherits_from: ems_refresh_worker
+  :event_catcher:
+    :event_catcher_openstack_service: "auto"
```
Fryguy, line 146: Ditto for these.

app/models/manageiq/providers/openstack/base_metrics_capture.rb
```diff
@@ -420,7 +420,7 @@ def parse_datetime(datetime)
   end
 
   def metric_service_from_settings
-    Settings[:workers][:worker_base][:queue_worker_base][:ems_metrics_collector_worker][:ems_metrics_openstack_default_service]
+    Settings.workers.ems_metrics_collector_worker.ems_metrics_openstack_default_service
```
Fryguy, line 423: @agrare I don't understand how this could have blown up previously.  Like why is this rescued with StandardError?

agrare: :shrug: looks like it was added here https://github.com/ManageIQ/manageiq/pull/13918

Looks like he was worried about the settings not being there?

config/settings.yml
chessbyte: wondering if these should become settings per provider.  in general, wondering whether global settings for provider workers makes sense.

app/models/manageiq/providers/openstack/base_metrics_capture.rb
Fryguy: Strangely, the code doesn't care if the key is there but the value is nil, even though the log message talks about a nil value 🤷‍♂️

I can tweak this in a follow up to make more sense.

config/settings.yml
Fryguy: Ideally, yes, these should be per-provider-instance settings, though unfortunately I don't think we don't have a mechanism for that at the moment (without making it part of the "create provider" form)

PR #689, opened 2021-02-12
Convert `supports_#{op}?` to `supports?(op)`

chessbyte:


spec/models/manageiq/providers/openstack/cloud_manager/vm_spec.rb
```diff
@@ -120,96 +120,96 @@
...
+  context "is supported?" do
     let(:ems) { FactoryBot.create(:ems_openstack) }
     let(:vm)  { FactoryBot.create(:vm_openstack, :ext_management_system => ems) }
     let(:power_state_on)        { "ACTIVE" }
     let(:power_state_suspended) { "SUSPENDED" }
 
     context("with :start") do
       let(:state) { :start }
-      include_examples "Vm operation is available when not powered on"
+      include_examples "Vm operation is supported when not powered on"
```
agrare, line 177: @chessbyte `"Vm operation is available when not powered on"` is the name of a shared example from https://github.com/ManageIQ/manageiq/blob/master/spec/support/examples_group/shared_examples_for_vm_or_template_is_available.rb#L1

chessbyte: Undid the shared example changes.

agrare (approved):

PR #701, opened 2021-04-05
Fix service detection if there is no admin tenant

agrare:
Detecting the default_tenant_name if that default isn't "admin" isn't being successfully returned as memoized successfully due the  the trailing `if`

Required for:
* https://github.com/iv1111/manageiq-providers-ibm_power_vc/pull/1

### DIFF
--- a/lib/manageiq/providers/openstack/legacy/openstack_handle/handle.rb
+++ b/lib/manageiq/providers/openstack/legacy/openstack_handle/handle.rb
@@ -80,7 +80,7 @@ module OpenstackHandle
       else
         Fog.const_get(service).new(opts)
       end
-    rescue Fog::OpenStack::Auth::Catalog::ServiceTypeError
+    rescue Fog::OpenStack::Auth::Catalog::ServiceTypeError, Fog::Service::NotFound
       $fog_log.warn("MIQ(#{self.class.name}##{__method__}) "\
                     "Service #{service} not available for openstack provider #{auth_url}")
       raise MiqException::ServiceNotAvailable
@@ -353,7 +353,7 @@ module OpenstackHandle
     end
 
     def default_tenant_name
-      return @default_tenant_name ||= "admin" if tenant_accessible?("admin")
+      return @default_tenant_name ||= ("admin" if tenant_accessible?("admin"))
       tenant_names.each do |name|
         next if name == "services"
         return @default_tenant_name ||= name if tenant_accessible?(name)
--- a/lib/manageiq/providers/openstack/legacy/openstack_handle/handle.rb
+++ b/lib/manageiq/providers/openstack/legacy/openstack_handle/handle.rb
@@ -353,11 +353,17 @@ module OpenstackHandle
     end
 
     def default_tenant_name
-      return @default_tenant_name ||= ("admin" if tenant_accessible?("admin"))
+      @default_tenant_name ||= detect_default_tenant_name
+    end
+
+    def detect_default_tenant_name
+      return "admin" if tenant_accessible?("admin")
+
       tenant_names.each do |name|
         next if name == "services"
-        return @default_tenant_name ||= name if tenant_accessible?(name)
+        return name if tenant_accessible?(name)
       end
+
       nil
     end
 


lib/manageiq/providers/openstack/legacy/openstack_handle/handle.rb
```diff
@@ -80,7 +80,7 @@ def self.raw_connect(username, password, auth_url, service = "Compute", extra_op
       else
         Fog.const_get(service).new(opts)
       end
-    rescue Fog::OpenStack::Auth::Catalog::ServiceTypeError
+    rescue Fog::OpenStack::Auth::Catalog::ServiceTypeError, Fog::Service::NotFound
```
agrare, line 83: I was also seeing a lot of "nfv service not available" exceptions even though we have a conditional on the service being non-nil.  This was because `detect_service` here was only catching `MiqException::ServiceNotAvailable` from down here https://github.com/ManageIQ/manageiq-providers-openstack/pull/701/files#diff-8b5a0f2181b4ddbbe590078b4a29f5bf5a4f763b151dd02f4951102497ee2af9R86 but the actual exception was `Fog::Service::NotFound`

lib/manageiq/providers/openstack/legacy/openstack_handle/handle.rb
```diff
@@ -352,7 +352,7 @@ def tenant_accessible?(name)
     end
 
     def default_tenant_name
-      return @default_tenant_name ||= "admin" if tenant_accessible?("admin")
+      return @default_tenant_name ||= ("admin" if tenant_accessible?("admin"))
```
agrare, line 355: If this `tenant_accessible?("admin")` returned false then the memoized `@default_tenant_name` was never being returned

lib/manageiq/providers/openstack/legacy/openstack_handle/handle.rb
agrare: Looks like I was fixing the same issue as @aufi was here https://github.com/ManageIQ/manageiq-providers-openstack/pull/702

lib/manageiq/providers/openstack/legacy/openstack_handle/handle.rb
chessbyte: is the code supposed to be doing the return only `if tenant_accessible?("admin")`

Suggest majorly simplifying as follows (or similar to make the logic clearer)
```ruby
if tenant_accessible?("admin")
  @default_tenant_name ||= "admin"
  return @default_tenant_name
end
```

chessbyte: looking at the whole method, maybe even the following would be better:

```ruby
def default_tenant_name
  @default_tenant_name ||= determine_tenant_name
end

def determine_tenant_name
  return "admin" if tenant_accessible?("admin")
  
  tenant_names.each do |name|
    next if name == "services"
    return name if tenant_accessible?(name)
  end

  nil
end
```

agrare: The code is supposed to be returning `@default_tenant_name` if it is set, otherwise detecting what the tenant_name is (with "admin" as the default).

:+1: splitting this logic up into two methods will be more clear

lib/manageiq/providers/openstack/legacy/openstack_handle/handle.rb
```diff
@@ -353,7 +353,7 @@ def tenant_accessible?(name)
...
```
kbrock, line 356: drive by. are you sure this is not:

```suggestion
      (return @default_tenant_name ||= "admin") if tenant_accessible?("admin")
```

agrare: Yes I am :)

Fryguy:
Backported to `lasker` in commit 7babdc56d36d6e61880ffe76a607998b54825d5a.

```text
commit 7babdc56d36d6e61880ffe76a607998b54825d5a
Date:   Fri Apr 9 12:25:23 2021 -0400

    Merge pull request #701 from agrare/fix_service_detection_non_admin_tenant
    
    Fix service detection if there is no admin tenant
    
    (cherry picked from commit 32ffa2a13c7194f2406aa25a7bdeb5234563b01a)
```

PR #704, opened 2021-04-15
Make it easier to subclass the Openstack refresh classes

agrare:
Replace hard-coded "openstack" vendor strings so that subclasses of the openstack refresh code can easily override them.

Required for:
* https://github.com/iv1111/manageiq-providers-ibm_power_vc/pull/2

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager.rb
@@ -493,6 +493,10 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
     @description ||= "OpenStack".freeze
   end
 
+  def self.vm_vendor
+    @vm_vendor ||= "openstack".freeze
+  end
+
   def self.default_blacklisted_event_names
     %w(
       identity.authenticate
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/openstack/cloud_manager.rb
```diff
@@ -493,6 +493,10 @@ def self.description
     @description ||= "OpenStack".freeze
   end
 
+  def self.vm_vendor
+    @vm_vendor ||= "openstack".freeze
```
chessbyte, line 497: out of curiosity, why the name `vm_vendor`?

agrare: Because this is for the `vms#vendor` column and doesn't always match the provider "vendor"

Fryguy:
Backported to `morphy` in commit 64a408611c6992cbd3103eecddece4288858918a.

```text
commit 64a408611c6992cbd3103eecddece4288858918a
Date:   Thu Apr 15 14:42:56 2021 -0400

    Merge pull request #704 from agrare/allow_for_subclass_openstack_refresh
    
    Make it easier to subclass the Openstack refresh classes
    
    (cherry picked from commit 670c19bca21c2646ac29a75eb6689f2324c0d38f)
```

PR #705, opened 2021-04-16
Don't hardcode STI class in refresh

agrare:
Rather than hard-code the various STI types we should be using the auto_model_class mechanism from the core persister

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/persister/definitions/cloud_collections.rb
+++ b/app/models/manageiq/providers/openstack/inventory/persister/definitions/cloud_collections.rb
@@ -50,14 +50,13 @@ module ManageIQ::Providers::Openstack::Inventory::Persister::Definitions::CloudC
   # model_class defined due to ovirt dependency
   def add_vms
     add_collection_with_ems_param(cloud, :vms) do |builder|
-      builder.add_properties(:model_class => ManageIQ::Providers::Openstack::CloudManager::Vm)
       builder.add_default_values(:vendor => manager.class.vm_vendor)
     end
   end
 
   def add_miq_templates
     add_collection(cloud, :miq_templates) do |builder|
-      builder.add_properties(:model_class => ManageIQ::Providers::Openstack::CloudManager::BaseTemplate)
+      builder.add_properties(:model_class => manager.class::BaseTemplate)
       builder.add_default_values(:ems_id => manager.id, :vendor => manager.class.vm_vendor)
 
       # Extra added to automatic attributes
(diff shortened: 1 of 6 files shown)


app/models/manageiq/providers/openstack/inventory/persister/definitions/cloud_collections.rb
```diff
@@ -99,15 +90,11 @@ def add_cloud_services
...
   def add_host_aggregates
-    add_collection_with_ems_param(cloud, :host_aggregates) do |builder|
-      builder.add_properties(:model_class => ManageIQ::Providers::Openstack::CloudManager::HostAggregate)
-    end
+    add_collection_with_ems_param(cloud, :host_aggregates)
   end
 
   def add_orchestration_stacks(extra_properties = {})
     add_collection(cloud, :orchestration_stacks, extra_properties) do |builder|
-      builder.add_properties(:model_class => ManageIQ::Providers::CloudManager::OrchestrationStack)
```
agrare, line 109: NOTE this is going to require a schema migration

spec/models/manageiq/providers/openstack/cloud_manager/refresh_spec_common.rb
```diff
@@ -357,6 +357,7 @@ def assert_flavors
         expect(flavor.enabled).to               eq true
         expect(flavor.cpu_cores).to             eq nil
         expect(flavor.description).to           eq descr
+        expect(flavor.type).to                  eq "ManageIQ::Providers::Openstack::CloudManager::Flavor"
```
agrare, line 360: These ensure that the auto_model_class is doing the right thing, if it fails it will be `::Flavor`.

app/models/manageiq/providers/openstack/inventory/persister/definitions/cloud_collections.rb
```diff
@@ -50,14 +50,13 @@ def initialize_cloud_inventory_collections
   # model_class defined due to ovirt dependency
   def add_vms
     add_collection_with_ems_param(cloud, :vms) do |builder|
-      builder.add_properties(:model_class => ManageIQ::Providers::Openstack::CloudManager::Vm)
```
chessbyte, line 53: why is there nothing to replace this deletion like `builder.add_properties(:model_class => manager.class::Vm)` ?

agrare: Because as long as the model_class matches the `manager_class::model_name` paradigm the model_class is automatically built.  You only have to do this if the model class doesn't match (e.g. the BaseTemplate)

Fryguy:
Backported to `morphy` in commit ba0771733b0a99b2db248c563e84d899777fa9ec.

```text
commit ba0771733b0a99b2db248c563e84d899777fa9ec
Date:   Tue Apr 20 17:00:54 2021 -0400

    Merge pull request #705 from agrare/refresh_subclass_sti_classes
    
    Don't hardcode STI class in refresh
    
    (cherry picked from commit a46e240944f7b34fd5146361438d0d9f3a38ea49)
```

PR #711, opened 2021-04-27
[WIP] Subclass orchestration stacks under the cloud manager

agrare:
These should be subclassed under the `Openstack::CloudManager`

TODO:
- [ ] Ensure nothing is expecting these to have the _exact_ class `ManageIQ::Providers::CloudManager::OrchestrationStack`

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/vnf.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/vnf.rb
@@ -1,4 +1,4 @@
-class ManageIQ::Providers::Openstack::CloudManager::Vnf < ManageIQ::Providers::CloudManager::OrchestrationStack
+class ManageIQ::Providers::Openstack::CloudManager::Vnf < ManageIQ::Providers::Openstack::CloudManager::OrchestrationStack
   include ManageIQ::Providers::Openstack::HelperMethods
   require_nested :Status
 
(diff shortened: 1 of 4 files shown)


agrare:
This might be an issue, not sure how this works with other types of orchestration stacks besides OpenStack: https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/cloud_manager/orchestration_template_runner.rb#L30-L32

kbrock:
ooh. had an old tab with this comment in there. think it is still relevant

```ruby
class ManageIQ::Providers::Openstack::CloudManager::OrchestrationStack < ManageIQ::Providers::CloudManager::OrchestrationStack
```

Since `OpenStack` extends from the regular `OrchestrationStack`, I think this will work fine.
I also feel if we have report results available, the non-`OpenStack` version of the models will work fine for our purposes.

- [ ] migrate the `type` column for `orchestration_stacks`. 

hmm. I was mid typing for another TODO but can't remember what it was.

kbrock:
I was thinking a few steps could get us there:

- look for `'class.name.*constantize'` and replace with `class_for_ems`.
  - for this pr, change [OrchestrationStack#create_stack](https://github.com/ManageIQ/manageiq/blob/58efcdc5e42db5676cfa43d2b23f8b8089fb5b55/app/models/manageiq/providers/cloud_manager/orchestration_stack.rb#L24) to use override `class_for_ems`.
- change `class_for_name` to return the values that are not standard (via above change)
- introduce PRs like this one that revert the `class_for_name` special case

Of course, just fixing this one issue and not refactoring across the board with `class_for_name` would probably be a more intelligent course of action.

PR #714, opened 2021-05-06
Don't hard-code network manager subclasses

agrare:
Support Ovirt and PowerVS NetworkManager subclasses

Required for:
* https://github.com/ManageIQ/manageiq-providers-ovirt/pull/559


Fryguy:
`morphy` version: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/569

PR #717, opened 2021-06-14
Fix the unauthorized login error message

agrare:
There are a number of issues that could cause login to return unauthorized including domain_id, username, password.

Just saying "username or password" is misleading as it could be something else wrong as well.

### DIFF
--- a/app/models/manageiq/providers/openstack/manager_mixin.rb
+++ b/app/models/manageiq/providers/openstack/manager_mixin.rb
@@ -133,7 +133,7 @@ module ManageIQ::Providers::Openstack::ManagerMixin
       when Excon::Errors::NotFound
         MiqException::MiqHostError.new("Endpoint not found.")
       when Excon::Errors::Unauthorized
-        MiqException::MiqInvalidCredentialsError.new("Login failed due to a bad username or password.")
+        MiqException::MiqInvalidCredentialsError.new("Login failed.")
       when Excon::Errors::Timeout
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/openstack/manager_mixin.rb
```diff
@@ -133,7 +133,7 @@ def translate_exception(err)
       when Excon::Errors::NotFound
         MiqException::MiqHostError.new("Endpoint not found.")
       when Excon::Errors::Unauthorized
-        MiqException::MiqInvalidCredentialsError.new("Login failed due to a bad username or password.")
```
agrare, line 136: I considered "due to a bad domain_id, username, or password" but felt like we might just be chasing an exhaustive list

chessbyte: can we tell if it is because of an invalid domain_id vs userid/password?

agrare: No all we get is a 401 unauthorized exception

Fryguy:
Backported to `lasker` in commit bf52478a3522d95be6849e7dd63ce17357c0e469.

```text
commit bf52478a3522d95be6849e7dd63ce17357c0e469
Date:   Thu Jun 17 13:42:27 2021 -0400

    Merge pull request #717 from agrare/fix_unauthorized_error_message
    
    Fix the unauthorized login error message
    
    (cherry picked from commit 1049685f6443c59e764a97c37b8aee3e928d0a54)
```

PR #728, opened 2021-08-23
Make sure admin_state_up isn't `nil`

agrare:
Fix for https://app.travis-ci.com/github/ManageIQ/manageiq-providers-openstack/jobs/533324862 after https://github.com/ManageIQ/manageiq-schema/pull/602

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/parser/network_manager.rb
+++ b/app/models/manageiq/providers/openstack/inventory/parser/network_manager.rb
@@ -79,7 +79,7 @@ class ManageIQ::Providers::Openstack::Inventory::Parser::NetworkManager < Manage
       network_port = persister.network_ports.find_or_build(np.id)
       network_port.name = np.name.blank? ? mac_address : np.name
       network_port.status = np.status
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/openstack/inventory/parser/network_manager.rb
```diff
@@ -79,7 +79,7 @@ def network_ports
       network_port = persister.network_ports.find_or_build(np.id)
       network_port.name = np.name.blank? ? mac_address : np.name
       network_port.status = np.status
-      network_port.admin_state_up = np.admin_state_up
+      network_port.admin_state_up = np.admin_state_up if np.admin_state_up.present?
```
agrare, line 82: So even though there is a default_value at the db level if we explicitly set `admin_state_up` to `nil` it still sets it to `nil`

agrare: Also, usually I'd `||` here but this is a boolean :smile:

PR #731, opened 2021-09-08
[WIP] Moved fields for network routers form code to provider repo

GilbertCherrie:
Moved fields from react code to provider repo for add/edit network routers form.

@miq-bot add-label enhancement


app/models/manageiq/providers/openstack/network_manager/network_router.rb
```diff
@@ -30,6 +30,280 @@ class ManageIQ::Providers::Openstack::NetworkManager::NetworkRouter < ::NetworkR
 
   supports :remove_interface
 
+  def self.params_for_create(ems, cloud_network_id)
+    if cloud_network_id.nil?
```
agrare, line 34: Interesting that the form is completely different depending on if a cloud_network_id is passed it...but then [[down here]](https://github.com/ManageIQ/manageiq-providers-openstack/pull/731/files#diff-493079fa00c2f1ecbd1a2b59f7c9fb859429e054695f9c06501577c3e537ed7eR129) we have to select a cloud network anyway

GilbertCherrie: Yeah we end up calling this twice. Once with no cloud_network_id for the first part of the form and the second time for the last field of the form which requires a cloud_network_id to get the options for the subnets list.

agrare: Could you make the last one a subform that depends on the selection of a cloud network?

GilbertCherrie: Yeah I can try

agrare: Okay wonder if we should hold off on this and tackle it as part of the "dialog workflow" API that we've discussed (cc @Fryguy @kavyanekkalapu)

kbrock: darn. most or our options only take an `ems` here and not an additional parameter.
In a perfect world, this would be consistent across all models (though we do not live in that world)

I do feel options in general should not take an additional parameter. So that looks a little fishy to me

app/models/manageiq/providers/openstack/network_manager/network_router.rb
```diff
@@ -30,6 +30,281 @@ class ManageIQ::Providers::Openstack::NetworkManager::NetworkRouter < ::NetworkR
...
+                :value => ct.id.to_s,
+              }
+            end,
+            :includeEmpty    => true,
+            :clearOnUnmount  => true,
+          },
+          {
+            :component => 'sub-form',
+            :title     => _('Router Information'),
+            :id        => 'routerInformation',
```
kbrock, line 62: I thought our variables were snake case not camel case?
guess it doesn't matter if this is not going over the wire.

kbrock (commented):
This looks like it works.

again, not a fan of having this called multiple times. Wonder if it is possible to provide 2 drop downs with all values and some kind of logic to state which ones match up

app/models/manageiq/providers/openstack/network_manager/network_router.rb
GilbertCherrie: @kbrock I can change it to camel case

kbrock (commented):
just the one tweak in 2 places.

looking good

app/models/manageiq/providers/openstack/network_manager/network_router.rb
```diff
@@ -30,6 +30,219 @@ class ManageIQ::Providers::Openstack::NetworkManager::NetworkRouter < ::NetworkR
...
+              :key            => "network-#{ems.id}",
+              :label          => _('Network'),
+              :placeholder    => "<#{_('Choose')}>",
+              :includeEmpty   => true,
+              :clearOnUnmount => true,
+              :condition      => {
+                :when => 'enable',
+                :is   => true,
+              },
+              :options        => ems.cloud_networks.select { |cn| cn.ems_id == ems.id && cn.external_facing == true }.map do |cn|
```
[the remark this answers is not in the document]
GilbertCherrie: I'll try these out and push my changes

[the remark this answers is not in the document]
GilbertCherrie: @kbrock `:public_networks` looks like it has the same result but is exactly equivalent to doing 
` :options        => ems.cloud_networks.where(:external_facing => true).map do |cn|` ?

app/models/manageiq/providers/openstack/network_manager/network_router.rb
```diff
@@ -30,6 +30,219 @@ class ManageIQ::Providers::Openstack::NetworkManager::NetworkRouter < ::NetworkR
...
+              :key            => "network-#{ext_management_system.id}",
+              :label          => _('Network'),
+              :placeholder    => "<#{_('Choose')}>",
+              :includeEmpty   => true,
+              :clearOnUnmount => true,
+              :condition      => {
+                :when => 'enable',
+                :is   => true,
+              },
+              :options        => ext_management_system.cloud_networks.select { |cn| cn.ems_id == ext_management_system.id && cn.external_facing == true }.map do |cn|
```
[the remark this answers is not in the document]
GilbertCherrie: @kbrock for this change I am getting an error ems not found. In the params_for_create ems is passed as a function parameter but the params_for_update does not have this parameter.

app/models/manageiq/providers/openstack/network_manager/network_router.rb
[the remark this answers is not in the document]
kbrock: It looks like what you want.

Since it goes through routers, I'm not sure that they are exactly the same. but if it gives the same results, then maybe using that makes sense

app/models/manageiq/providers/openstack/network_manager/network_router.rb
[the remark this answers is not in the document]
kbrock: sorry. you made the right change here: `ext_management_system.public_networks`

kbrock:
It looks like `NetworkRouter#public_network` is the way to go.

But can you verify that it is accurate? I think it shows both public and private networks.

It is defined as 

https://github.com/ManageIQ/manageiq/blob/recursive_model_validation/app/models/network_router.rb#L23

when I think it should be something closer to (I have not verified this code and the `public_networks` looks a little suspect to me as well.

```
  has_many :public_networks, -> { distinct }, :through => :cloud_subnets

```

GilbertCherrie:
> It looks like `NetworkRouter#public_networks` is the way to go.
> 
> But can you verify that it is accurate? I think it shows both public and private networks.
> 
> It is defined as
> 
> https://github.com/ManageIQ/manageiq/blob/master/app/models/network_router.rb#L23
> 
> when I think it should be something closer to (I have not verified this code and the `public_networks` looks a little suspect to me as well.)
> 
> ```diff
> -  alias public_network cloud_network
> +  has_many :public_networks, -> { distinct }, :through => :cloud_subnets
> ```

The current code seems to work but if you think it may be incorrect should we leave it as:
`:options        => ems.cloud_networks.where(:external_facing => true).map do |cn|` ?

agrare:
@kbrock I think the `public_network` is the "uplink" aka the network the network_router belongs_to where as the private_networks are the "downlinks" aka the networks that belong_to the network_router.  The cloud_subnets also appear to belong_to the network_router no the other way around

PR #734, opened 2021-09-28
Update fog-openstack to v1.0

agrare:
Co-depends: https://github.com/ManageIQ/manageiq-gems-pending/pull/558

### DIFF
--- a/spec/models/manageiq/providers/openstack/infra_manager/refresher_rhos_juno_spec.rb
+++ b/spec/models/manageiq/providers/openstack/infra_manager/refresher_rhos_juno_spec.rb
@@ -5,11 +5,12 @@ describe ManageIQ::Providers::Openstack::InfraManager::Refresher do
 
   before(:each) do
     zone = EvmSpecHelper.local_miq_server.zone
-    @ems = FactoryBot.create(:ems_openstack_infra, :zone => zone, :hostname => "192.168.24.1",
-                              :ipaddress => "192.168.24.1", :port => 5000, :api_version => 'v2',
+    credentials = Rails.application.secrets.openstack
+    @ems = FactoryBot.create(:ems_openstack_infra, :zone => zone, :hostname => credentials[:hostname],
+                              :ipaddress => credentials[:hostname], :port => credentials[:port].to_i, :api_version => 'v3',
                               :security_protocol => 'no-ssl')
     @ems.update_authentication(
-      :default => {:userid => "admin", :password => "1fb03e3ec17f5840a5448ef2115a7ec7d645c982"})
+      :default => {:userid => credentials[:userid], :password => credentials[:password]})
   end
 
   it "will perform a full refresh" do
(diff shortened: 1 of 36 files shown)


agrare:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq-gems-pending/pull/525

agrare:
Getting some VCR refresher errors that I'll need to work out the root cause of

kbrock:
kicking

agrare:
Only failing on the ironic/infra_manager specs because microstack doesn't have that service enabled

agrare:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq-gems-pending/pull/558

agrare:
@nasark has incredibly gotten upstream openstack installed with the Ironic service, rebased this on master and locally only failing specs are the infra_manager refresher_spec.

nasark:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq-gems-pending/pull/558

agrare:
Cross-repo tests are green

nasark:
Just wanted to add that fog v1.0 will allow us to support cinder V3 which is used in upstream Openstack Wallaby> releases

agrare:
Hey @kbrock can you take a look?

spec/models/manageiq/providers/openstack/infra_manager/refresher_rhos_juno_spec.rb
```diff
@@ -5,11 +5,12 @@
 
   before(:each) do
     zone = EvmSpecHelper.local_miq_server.zone
-    @ems = FactoryBot.create(:ems_openstack_infra, :zone => zone, :hostname => "192.168.24.1",
-                              :ipaddress => "192.168.24.1", :port => 5000, :api_version => 'v2',
-                              :security_protocol => 'no-ssl')
+    credentials = Rails.application.secrets.openstack
+    @ems = FactoryBot.create(:ems_openstack_infra, :zone => zone, :hostname => credentials[:hostname],
+                              :ipaddress => credentials[:hostname], :port => credentials[:port].to_i, :api_version => 'v3',
```
kbrock, line 10: do we want to default the port?

```suggestion
                              :ipaddress => credentials[:hostname], :port => credentials[:port]&.to_i || 5000, :api_version => 'v3',
```

agrare: I don't think so because `credentials` comes from `config/secrets.defaults.yml` and so will always be present.

kbrock:
Are we concerned with:
- https://github.com/kbrock/manageiq-providers-openstack/blob/master/lib/tasks_private/openstack.rake#L41

Not sure if `Fog::{Compute,Network,Identity}` still works the same or if we need to add `OpenStack` in there as well.

agrare:
> Not sure if Fog::{Compute,Network,Identity} still works the same or if we need to add OpenStack in there as well.

These do still work but can update for consistency

agrare:
> Also, could you update this comment with 2 references to the 2 classes?

Done

PR #742, opened 2021-10-07
Network router changes for API

kbrock:
add parameters for create and update to enable a data driven form for openstack  network routers

currently create and update are working


kbrock:
ok, @agrare this has stabilized

What do we need to do to move this forward

PR #750, opened 2021-10-20
Validate ceilometer endpoint

agrare:
We currently assume that if we can connect to OpenStack Compute then we will be able to connect to Ceilometer, however this isn't always the case

- [x] Update DDF form to pass event endpoint with default credentials and verify in one step
- [x] Update verify_credentials to verify all endpoints
- [x] Live test on an Openstack without ceilometer (aka microstack) raises service not found error
- [x] Live test on an Openstack with ceilometer/stf/amqp

### DIFF
--- a/app/models/manageiq/providers/openstack/manager_mixin.rb
+++ b/app/models/manageiq/providers/openstack/manager_mixin.rb
@@ -15,10 +15,10 @@ module ManageIQ::Providers::Openstack::ManagerMixin
     def amqp_available?(password, params)
       require 'manageiq/providers/openstack/legacy/events/openstack_rabbit_event_monitor'
       OpenstackRabbitEventMonitor.available?(
-        :hostname => params[:amqp_hostname],
-        :username => params[:amqp_userid],
+        :hostname => params[:hostname],
+        :username => params[:userid],
         :password => ManageIQ::Password.try_decrypt(password),
-        :port     => params[:amqp_api_port] || params[:amqp_port].to_s
+        :port     => params[:api_port] || params[:port].to_s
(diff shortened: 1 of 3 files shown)


agrare:
Okay I couldn't find a live openstack with stf/amqp to test with, but I confirmed that the proper attributes are still passed to the underlying `OpenstackEventMonitor.available?` code that actually does the check so dropping the `stf_` and `amqp_` prefixes hasn't broken anything.

app/models/manageiq/providers/openstack/manager_mixin.rb
```diff
@@ -93,25 +98,33 @@ def ems_connect?(password, params, service)
...
+        password = ManageIQ::Password.try_decrypt(password)
+        password ||= find(args["id"]).authentication_password(endpoint_name) if args["id"]
 
-      !!raw_connect(password, params.symbolize_keys)
+        params = {
+          "userid"   => userid,
+          "password" => password,
+        }.merge(endpoint.slice("hostname", "port", "security_protocol"))
+
+        params.merge!(args.slice('name', 'provider_region', 'api_version', 'uid_ems'))
```
Fryguy, line 123: I feel like this would read nicer as...

```suggestion
        endpoint_params = endpoint.slice("hostname", "port", "security_protocol")
        args_params     = args.slice('name', 'provider_region', 'api_version', 'uid_ems')

        params = {
            "userid"   => userid,
            "password" => password
        }.merge(endpoint_params).merge(args_params)
```

or something like that.

Fryguy (approved):
LGTM

PR #753, opened 2021-11-03
refactoring code so the vm_details can be called from powervc subclass instance

Kuldip-Nanda:
… instance

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
@@ -291,104 +291,112 @@ class ManageIQ::Providers::Openstack::Inventory::Parser::CloudManager < ManageIQ
     hosts = related_infra_ems.try(:hosts)
 
     collector.vms.each do |s|
-      if hosts && !s.os_ext_srv_attr_host.blank?
-        parent_host = hosts.find_by('lower(hypervisor_hostname) = ? OR lower(hypervisor_hostname) = ?', s.os_ext_srv_attr_host.split('.').first.downcase, s.os_ext_srv_attr_host.downcase)
-        parent_cluster = parent_host.try(:ems_cluster)
-      else
-        parent_host = nil
-        parent_cluster = nil
-      end
-
-      availability_zone = s.availability_zone.blank? ? "null_az" : s.availability_zone
-      miq_template_lazy = persister.miq_templates.lazy_find(s.image["id"])
-
-      server = persister.vms.find_or_build(s.id.to_s)
-      server.uid_ems = s.id
-      server.name = s.name
-      server.raw_power_state = s.state || "UNKNOWN"
-      server.connection_state = "connected"
-      server.location = "unknown"
-      server.host = parent_host
-      server.ems_cluster = parent_cluster
-      server.availability_zone = persister.availability_zones.lazy_find(availability_zone)
-      server.key_pairs = [persister.auth_key_pairs.lazy_find(s.key_name)].compact
-      server.cloud_tenant = persister.cloud_tenants.lazy_find(s.tenant_id.to_s)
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -299,96 +299,104 @@ def vms
         parent_cluster = nil
       end
 
-      availability_zone = s.availability_zone.blank? ? "null_az" : s.availability_zone
```
agrare, line 302: Hm cutting this method here instead of above to work around the shared `hosts` looks awkward.

Prefer:
```ruby
collector.vms.each do |server|
  parse_server(server)
end
```

And deal with the hosts another way like move it to the collector or add a parser instance variable

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -299,96 +299,104 @@ def vms
...
-            # reuse the device names from above in the event that this is an
-            # instance that was booted from a volume
-            :device_name => attachment_names.fetch(dev, dev)
-          ).assign_attributes(
-            :location        => dev,
-            :device_type     => "disk",
-            :controller_type => "openstack"
-          )
-        end
+  def get_flavor(vm)
```
agrare, line 306: Where is this method used?

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -299,96 +299,104 @@ def vms
...
+    collector.find_flavor(vm.flavor["id"].to_s)
+  end
+
+  def vm_details(vm, parent_host, parent_cluster)
```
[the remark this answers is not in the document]
Kuldip-Nanda: sure I can change the name from vm_details to parse

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
jaywcarman: Looks like only in `ManageIQ::Providers::IbmPowerVc::Inventory::Parser::CloudManager` so we should move it to that class:
https://github.com/ManageIQ/manageiq-providers-ibm_power_vc/pull/28/files#diff-2dbea85c89843b8aa05cba1d5a6caa08e2b2389d21580c2176026180d415727eR7

agrare: Actually looks like it could be used down [[here]](https://github.com/ManageIQ/manageiq-providers-openstack/pull/753/files#diff-46c42db89009bb02993acce873dd594e15baf153ebda72079095b9946a480802R331)

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -291,104 +291,112 @@ def vms
...
+    server.availability_zone = persister.availability_zones.lazy_find(availability_zone)
+    server.key_pairs = [persister.auth_key_pairs.lazy_find(vm.key_name)].compact
+    server.cloud_tenant = persister.cloud_tenants.lazy_find(vm.tenant_id.to_s)
+    server.genealogy_parent = miq_template_lazy unless vm.image["id"].nil?
+
+    # to populate the hardware, we need some fields from the flavor object
+    # that we don't already have from the flavor field on the server details
+    # returned from the openstack api. It's possible that no such flavor was found
+    # due to some intermittent network issue or etc, so we use try to not break.
+    flavor = collector.find_flavor(vm.flavor["id"].to_s)
```
agrare, line 331: @Kuldip-Nanda you can use your new `get_flavor` method here

agrare (approved):

PR #757, opened 2021-12-10
converting supported_for_create to supports :create

kbrock:
part of ManageIQ/manageiq#21573


app/models/manageiq/providers/openstack/provider.rb
```diff
@@ -16,6 +16,8 @@ class ManageIQ::Providers::Openstack::Provider < ::Provider
 
   validates :name, :presence => true, :uniqueness => true
 
+  supports :create
```
agrare, line 19: Do we check the Provider class for `create` ?

kbrock: don't think so

agrare: We can clean this up later

PR #758, opened 2021-12-13
Don't hard code cloud_tenant STI type

agrare:
This causes any subclasses of openstack (e.g. powervc) to have the wrong type on their cloud_tenant records

We're already checking in specs that the cloud_tenant records have the appropriate class name for this repo


Fryguy:
Do we need a data migration for this?

agrare:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq-providers-ibm_power_vc/pull/34

PR #760, opened 2021-12-22
Use the snapshots collection from core

agrare:
Now that the core persister has a snapshots inventory collection we can use that rather than redefine it here

Follow-up to https://github.com/ManageIQ/manageiq/pull/21629


agrare:
@Fryguy green

PR #762, opened 2022-01-04
Cleanup the TargetCollection persister

agrare:
The strategy is handled automatically for targeted persisters and with the add_network_collection changes we can drop the parent ivar weirdness around adding the network manager collections.

### DIFF
--- a/app/models/manageiq/providers/openstack/inventory/persister/target_collection.rb
+++ b/app/models/manageiq/providers/openstack/inventory/persister/target_collection.rb
@@ -7,26 +7,10 @@ class ManageIQ::Providers::Openstack::Inventory::Persister::TargetCollection < M
     true
   end
 
-  def strategy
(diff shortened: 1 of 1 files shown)


agrare:
@miq-bot cross-repo-tests manageiq-providers-ibm_cic, manageiq-providers-ibm_power_vc

app/models/manageiq/providers/openstack/inventory/persister/target_collection.rb
```diff
@@ -7,26 +7,10 @@ def targeted?
...
-    end
-  end
-
   def initialize_inventory_collections
     initialize_tag_mapper
     initialize_cloud_inventory_collections
-
-    @init_network_collections = true
     initialize_network_inventory_collections
-    @init_network_collections = false
```
agrare, line 28: This always bugged me, nice to be able to ditch it

PR #764, opened 2022-01-05
Standardize storage manager collections

agrare:


app/models/manageiq/providers/openstack/inventory/persister/definitions/storage_collections.rb
```diff
@@ -1,39 +1,34 @@
...
 
   def add_cinder_collection(collection_name, extra_properties = {}, settings = {}, &block)
     settings[:parent] ||= cinder_manager
     add_collection(storage, collection_name, extra_properties, settings, &block)
   end
+
+  def add_swift_collection(collection_name, extra_properties = {}, settings = {}, &block)
+    settings[:parent] ||= swift_manager
+    add_collection(storage, collection_name, extra_properties, settings, &block)
+  end
```
agrare, line 33: These are intended to be similar to the `add_storage_collection` from core, but since openstack has two storage managers we have to be more specific.

PR #765, opened 2022-01-06
[WIP] Combine child managers in one refresh

agrare:


kbrock:
Can you comment in the second commit if a file is simply renaming/moving?
git is not really telling the whole story and there are so many lines that I can't tell what is moved and what is actually changed.

But all in all, this seems like a simple PR and I think we should just merge.

kbrock:
@agrare Surprised we didn't merge this.
Do you remember if there was actually something causing us to stall here?

agrare:
@kbrock I think the issue was not having access to the VCR environments back at RH but now that we use microstack it should be simple once I get some time

PR #768, opened 2022-01-27
Symbolize key for cloud_volume CRUD options

agrare:
The options passed in to cloud_volume create/update have string keys when invoked from the API.
This was causing issues where the :display_name / :cloud_tenant weren't being picked up on because we were looking for symbol keys.

### DIFF
--- a/app/models/manageiq/providers/openstack/storage_manager/cinder_manager/cloud_volume.rb
+++ b/app/models/manageiq/providers/openstack/storage_manager/cinder_manager/cloud_volume.rb
@@ -85,15 +85,14 @@ class ManageIQ::Providers::Openstack::StorageManager::CinderManager::CloudVolume
   end
 
   def self.raw_create_volume(ext_management_system, options)
+    options = options.symbolize_keys
+
     cloud_tenant = options.delete(:cloud_tenant)
     volume = nil
 
     # provide display_name for Cinder V1
     options[:display_name] |= options[:name]
-    with_notification(:cloud_volume_create,
-                      :options => {
-                        :volume_name => options[:name],
-                      }) do
+    with_notification(:cloud_volume_create, :options => {:volume_name => options[:name]}) do
       ext_management_system.with_provider_connection(cinder_connection_options(cloud_tenant)) do |service|
         volume = service.volumes.new(options)
         volume.save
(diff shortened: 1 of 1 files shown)


agrare:
I considered adding `.symbolize_keys` to the API that calls this, but figured this way would work with both symbols and strings

app/models/manageiq/providers/openstack/storage_manager/cinder_manager/cloud_volume.rb
```diff
@@ -85,15 +85,14 @@ def self.params_for_create(ems)
   end
 
   def self.raw_create_volume(ext_management_system, options)
+    options.symbolize_keys!
```
Fryguy, line 88: My only concern here is that we are modifying the caller's options...not sure we should do that or not. Similarly on raw_update_volume

chessbyte:
```suggestion
    options = options.dup.symbolize_keys!
```

agrare: > My only concern here is that we are modifying the caller's options

Yeah that's fair

>     options = options.dup.symbolize_keys!

well don't need the dup and the bang version of symbolize_keys!, can just `options = options.symbolize_keys`

Fryguy:
Backported to `morphy` in commit ef56c99ff0611f8a9159bf733af226282ea478f1.

```text
commit ef56c99ff0611f8a9159bf733af226282ea478f1
Date:   Mon Jan 31 09:10:15 2022 -0500

    Merge pull request #768 from agrare/cloud_volume_symbolize_keys
    
    Symbolize key for cloud_volume CRUD options
    
    (cherry picked from commit 4ebbbc050586da6a7a7a556f023bb793c358d3a9)
```

Fryguy:
Backported to `lasker` in commit f42f4cbef675baa6bdd3d05041ab9663475f7e06.

```text
commit f42f4cbef675baa6bdd3d05041ab9663475f7e06
Date:   Mon Jan 31 09:10:15 2022 -0500

    Merge pull request #768 from agrare/cloud_volume_symbolize_keys
    
    Symbolize key for cloud_volume CRUD options
    
    (cherry picked from commit 4ebbbc050586da6a7a7a556f023bb793c358d3a9)
```

PR #769, opened 2022-01-31
Fixed cloud volumes add page missing fields

GilbertCherrie:
Fixes: https://github.com/ManageIQ/manageiq-ui-classic/issues/8077

Fixes the missing field issue for the add cloud volume form.

Before:
Screen Shot 2022-01-31 at 2 25 24 PM

After:
Screen Shot 2022-01-31 at 2 16 13 PM

### DIFF
--- a/app/models/manageiq/providers/openstack/storage_manager/cinder_manager/cloud_volume.rb
+++ b/app/models/manageiq/providers/openstack/storage_manager/cinder_manager/cloud_volume.rb
@@ -35,10 +35,6 @@ class ManageIQ::Providers::Openstack::StorageManager::CinderManager::CloudVolume
           :isRequired   => true,
           :includeEmpty => true,
           :validate     => [{:type => 'required'}],
-          :condition    => {
-            :when => 'edit',
-            :is   => false,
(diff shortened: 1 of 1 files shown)


GilbertCherrie:
The edit cloud volume page is also broken but that is a separate issue I can address in a follow up PR.

GilbertCherrie:
@miq-bot add-label miq-bot add-label

GilbertCherrie:
@miq-bot add-label bug

Fryguy:
I thought these things were not editable...like you can't change the size of a cloud volume once it's been created in many cases.

agrare:
@Fryguy the "new" implementation of how the API gets DDF params from OPTIONS uses a params_for_update method not a single method for create/update.  Edit is currently completely busted and this params_for_create method is not being called by edit so it doesn't make sense to have edit conditionals.

agrare (approved):
Tested against a microstack with cinder, able to create cloud volumes in a specific project and az

chessbyte:
@agrare Should this be backported to Morphy?

PR #771, opened 2022-02-01
Fix CloudVolume Create - DDF sends cloud_tenant_id not cloud_tenant record

agrare:
Follow-up to https://github.com/ManageIQ/manageiq-providers-openstack/pull/768 now that I am able to test with the proper UI form after https://github.com/ManageIQ/manageiq-providers-openstack/pull/769

### DIFF
--- a/app/models/manageiq/providers/openstack/storage_manager/cinder_manager/cloud_volume.rb
+++ b/app/models/manageiq/providers/openstack/storage_manager/cinder_manager/cloud_volume.rb
@@ -87,7 +87,8 @@ class ManageIQ::Providers::Openstack::StorageManager::CinderManager::CloudVolume
   def self.raw_create_volume(ext_management_system, options)
     options = options.symbolize_keys
 
-    cloud_tenant = options.delete(:cloud_tenant)
+    cloud_tenant_id = options.delete(:cloud_tenant_id)
+    cloud_tenant    = CloudTenant.find_by(:id => cloud_tenant_id) if cloud_tenant_id
     volume = nil
 
     # provide display_name for Cinder V1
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/openstack/storage_manager/cinder_manager/cloud_volume.rb
```diff
@@ -87,7 +87,8 @@ def self.params_for_create(ems)
   def self.raw_create_volume(ext_management_system, options)
     options = options.symbolize_keys
 
-    cloud_tenant = options.delete(:cloud_tenant)
+    cloud_tenant_id = options.delete(:cloud_tenant_id)
+    cloud_tenant    = CloudTenant.find_by(:id => cloud_tenant_id) if cloud_tenant_id
```
Fryguy, line 91: Does this need RBAC? (i.e. should this be `resource_search` instead?)

agrare: We should definitely be doing that on the API side before this is queued (would need to check to see if it is), but this is on the generic worker side of the queue we don't have access to resource_search

Fryguy: Ah ok...this is fine then

Fryguy:
Backported to `morphy` in commit c79c78b67be4527d523bfc9bc13bf0011a2fe08a.

```text
commit c79c78b67be4527d523bfc9bc13bf0011a2fe08a
Date:   Tue Feb 8 12:01:51 2022 -0500

    Merge pull request #771 from agrare/fix_cloud_volume_create_cloud_tenant_id
    
    Fix CloudVolume Create - DDF sends cloud_tenant_id not cloud_tenant record
    
    (cherry picked from commit e1b99d6856ef094023d26ffbd8b8a3126b8813cb)
```

PR #775, opened 2022-02-05
Restore deleted before_install

Fryguy:


bin/before_install
```diff
@@ -4,6 +4,39 @@ if [ -n "$CI" ]; then
...
+    cmake .. -DCMAKE_INSTALL_PREFIX=/usr -DBUILD_BINDINGS=
+
+    # Compile system libraries.
+    make all
+
+    # Install system libraries
+    sudo make install
+
+    # Enable the qpid_proton bundler group
+    [ -z "$BUNDLE_WITH" ] && bundle config with qpid_proton
```
Fryguy, line 34: @agrare Technically this failed with 

```
bin/before_install: line 34: bundle: command not found
```

because we don't have Ruby yet, and I'm not sure why it doesn't have BUNDLE_WITH in the env like in https://github.com/ManageIQ/manageiq-providers-nuage/pull/267

agrare: Does it work with BUNDLE_WITH like nuage has?

Fryguy: It seems to work regardless (both repos passed), but maybe the tests themselves are gated somehow that I'm not seeing?

agrare: We do `config.filter_run_excluding(:qpid_proton) unless ENV['CI'] || Gem.loaded_specs.key?(:qpid_proton)` so it looks like this would exclude those tests if qpid_proton weren't installed.

Seems like we'd want to never skip on `ENV['CI']` regardless of if qpid_proton were available, since that would indicate something was wrong with the env setup.

Okay I'll convert this to use the env var like nuage and make sure the qpid_proton event_catcher specs are running.

PR #776, opened 2022-02-06
Add bundle with GitHub workflows

agrare:


agrare:
@Fryguy confirmed with a `puts` that the qpid specs are running

PR #780, opened 2022-02-17
The Openstack NetworkManager cannot be added standalone

agrare:
This is a child manager of a CloudManager and is not supported for create by itself.

This had `supports :create` incorrectly for a long time, but after https://github.com/ManageIQ/manageiq/pull/21573 started checking for it now it shows up as supported for creation.


chessbyte (approved):

agrare:
This was originally added in https://github.com/ManageIQ/manageiq/pull/12071 I think as just "the network manager supports creating things"

PR #783, opened 2022-03-03
Minor Cleanup child manager init/ensure_

kbrock:
Similar in vein to #677
This was part of a very old more drastic PR that I wrote.

I noticed since cinder and swift already have this refactor here, implementing it for network manager makes sense

- Don't subclass from core NetworkManager. (mimics work already done for swift/cinder)
- Remove duplicate `ensure_swift_manager` (all of the ensures seem to be defined in the parent class right now
except for network manager

Keeping duplicate `ensure_network_manager`. It feels like it is best to define where it is actually used

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager.rb
@@ -21,6 +21,11 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
   require_nested :Template
   require_nested :Vm
 
+  has_one :network_manager,
+          :foreign_key => :parent_ems_id,
+          :class_name  => "ManageIQ::Providers::Openstack::NetworkManager",
+          :autosave    => true,
+          :dependent   => :destroy
   has_many :storage_managers,
            :foreign_key => :parent_ems_id,
            :class_name  => "ManageIQ::Providers::StorageManager",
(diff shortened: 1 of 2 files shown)


kbrock:
A little confused about all these VCR errors.
Running locally works :(

agrare:
I actually get the same vcr errors locally

app/models/manageiq/providers/openstack/cloud_manager.rb
```diff
@@ -21,6 +21,12 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
   require_nested :Template
   require_nested :Vm
 
+  has_one :network_manager,
+          :foreign_key => :parent_ems_id,
+          :class_name  => "ManageIQ::Providers::Openstack::NetworkManager",
+          :autosave    => true,
+          :dependent   => :destroy,
+          :inverse_of  => :parent_manager
```
agrare, line 29: Okay some weird rails stuff going on, this is causing the failure somehow:
```
           :class_name  => "ManageIQ::Providers::Openstack::NetworkManager",
           :autosave    => true,
-          :dependent   => :destroy,
-          :inverse_of  => :parent_manager
+          :dependent   => :destroy
   has_many :storage_managers,
```

All green with inverse_of removed

agrare (approved):

PR #784, opened 2022-03-04
Refactor AuthKeyPair.raw_create_key_pair

agrare:
Merge together with: https://github.com/ManageIQ/manageiq/pull/21759

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/auth_key_pair.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/auth_key_pair.rb
@@ -7,7 +7,9 @@ class ManageIQ::Providers::Openstack::CloudManager::AuthKeyPair < ManageIQ::Prov
   def self.raw_create_key_pair(ext_management_system, create_options)
     connection_options = {:service => 'Compute'}
     ext_management_system.with_provider_connection(connection_options) do |service|
-      service.key_pairs.create(create_options)
+      kp = service.key_pairs.create(create_options)
+
+      {:name => kp.name, :fingerprint => kp.fingerprint, :auth_key => kp.private_key}
     end
   rescue => err
     _log.error "keypair=[#{name}], error: #{err}"


app/models/manageiq/providers/openstack/cloud_manager/auth_key_pair.rb
```diff
@@ -7,7 +7,9 @@ class ManageIQ::Providers::Openstack::CloudManager::AuthKeyPair < ManageIQ::Prov
   def self.raw_create_key_pair(ext_management_system, create_options)
     connection_options = {:service => 'Compute'}
     ext_management_system.with_provider_connection(connection_options) do |service|
-      service.key_pairs.create(create_options)
+      kp = service.key_pairs.create(create_options)
+
+      {:name => kp.name, :fingerprint => kp.fingerprint, :auth_key => kp.private_key}
```
jaywcarman, line 12: I think this is only used to store public keys right?
```suggestion
      {:name => kp.name, :fingerprint => kp.fingerprint, :auth_key => kp.public_key}
```

agrare: `:auth_key` is used to store the private key, see https://github.com/ManageIQ/manageiq/pull/21759/files#diff-b80f64b28488206bf8d41b9b53fb77a5490267a1a91b5d1e4f0897af88951554L46

jaywcarman: Thanks for the explanation! You're right - I was looking at the `:public_key` column. Makes sense now, though :)

agrare:
Test failures are due to codependency on the core PR, cross-repo tests are green

Fryguy (approved):

Fryguy:
Backported to `najdorf` in commit f66db09632fbc7df77a096f7862d05e6cc541387.

```text
commit f66db09632fbc7df77a096f7862d05e6cc541387
Date:   Tue Mar 8 14:45:55 2022 -0500

    Merge pull request #784 from agrare/refactor_raw_create_key_pair
    
    Refactor AuthKeyPair.raw_create_key_pair
    
    (cherry picked from commit b72699ade55e08d4a2e5ecd87f621dbe0a8be1fc)
```

PR #785, opened 2022-03-28
Added fields for cloud object store container form

GilbertCherrie:
part of: https://github.com/ManageIQ/manageiq-api/pull/1142

Moved fields to provider repo for the create cloud object store container form as well as add support for creating cloud object store container form.

Fields:
Screen Shot 2022-03-28 at 8 25 51 AM

### DIFF
--- a/app/models/manageiq/providers/openstack/storage_manager/swift_manager/cloud_object_store_container.rb
+++ b/app/models/manageiq/providers/openstack/storage_manager/swift_manager/cloud_object_store_container.rb
@@ -7,9 +7,16 @@ class ManageIQ::Providers::Openstack::StorageManager::SwiftManager::CloudObjectS
 
   def self.params_for_create(ems)
     cloud_tenants = ems.parent_manager.cloud_tenants
-
     {
       :fields => [
+        {
+          :component  => 'text-field',
+          :name       => 'name',
+          :id         => 'name',
+          :label      => _('Container Name'),
+          :isRequired => true,
+          :validate   => [{:type => 'required'}],
+        },
         {
           :component    => 'select',
           :name         => 'cloud_tenant_id',
@@ -18,10 +25,6 @@ class ManageIQ::Providers::Openstack::StorageManager::SwiftManager::CloudObjectS
           :isRequired   => true,
           :includeEmpty => true,
           :validate     => [{:type => 'required'}],
-          :condition    => {
-            :when => 'edit',
-            :is   => false,
-          },
           :options      => cloud_tenants.map do |ct|
             {:label => ct.name, :value => ct.id}
           end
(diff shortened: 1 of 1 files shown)


GilbertCherrie:
@miq-bot add-label enhancement

GilbertCherrie:
@miq-bot assign @agrare

app/models/manageiq/providers/openstack/storage_manager/swift_manager/cloud_object_store_container.rb
```diff
@@ -31,6 +34,8 @@ def self.params_for_create(ems)
   end
 
   def self.raw_cloud_object_store_container_create(ext_management_system, options)
+    cloud_tenant_id = options.delete(:cloud_tenant_id)
+    cloud_tenant = CloudTenant.find_by(:id => cloud_tenant_id) if cloud_tenant_id
     cloud_tenant = options.delete(:cloud_tenant)
```
agrare, line 39: @GilbertCherrie should this line be deleted since you're setting `cloud_tenant` one line above?

app/models/manageiq/providers/openstack/storage_manager/swift_manager/cloud_object_store_container.rb
```diff
@@ -31,6 +34,8 @@ def self.params_for_create(ems)
   end
 
   def self.raw_cloud_object_store_container_create(ext_management_system, options)
+    cloud_tenant_id = options.delete(:cloud_tenant_id)
```
agrare, line 37: We should make sure these options are symbols since typically these will be strings when they are passed down through the API

app/models/manageiq/providers/openstack/storage_manager/swift_manager/cloud_object_store_container.rb
kbrock: I think I remember leaving a line like this in there for supporting both the old ui-classic controller as well as the api.

If anything, please move this line above the one using `cloud_tenant_id` for sure.
Not sure if we need to support both of them during the transition.

Also, could you find a way to get more of the logic for extracting the `cloud_tenant_id` from the original data and move it into here?

agrare: Yeah if you actually want to have both we need to only set this if it isn't already so `cloud_tenant ||= ...` but I'd prefer only having one way of doing this if at all possible

app/models/manageiq/providers/openstack/storage_manager/swift_manager/cloud_object_store_container.rb
kbrock: translation: we want to use that `symbolize_keys` when in the api when we call into here

GilbertCherrie: @agrare @kbrock Just updated the api pr to ensure they are being passed in as symbols.

agrare (approved):

app/models/manageiq/providers/openstack/storage_manager/swift_manager/cloud_object_store_container.rb
```diff
@@ -31,20 +34,21 @@ def self.params_for_create(ems)
   end
 
   def self.raw_cloud_object_store_container_create(ext_management_system, options)
-    cloud_tenant = options.delete(:cloud_tenant)
+    cloud_tenant_id = options.delete(:cloud_tenant_id)
+    cloud_tenant = CloudTenant.find_by(:id => cloud_tenant_id) if cloud_tenant_id
```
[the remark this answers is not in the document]
GilbertCherrie: @kbrock So should I add this to the api?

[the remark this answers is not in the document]
agrare: https://github.com/ManageIQ/manageiq-api/pull/1142#discussion_r837802274 does mitigate this to a degree, though still allows a user to see the full list of cloud-tenants in the dropdown.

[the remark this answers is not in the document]
kbrock: Have we figured out a solution for this or are we letting it slide?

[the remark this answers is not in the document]
agrare: I don't think this is something that we are going to solve here, going to need a larger discussion.  This is the same pattern that we have for the rest of these ddf forms

PR #787, opened 2022-03-31
Cloud volume snapshot

kbrock:
part of https://github.com/ManageIQ/manageiq-api/pull/1134

move crud for cloud volume snapshots to core and use `raw_` naming convention


app/models/manageiq/providers/openstack/storage_manager/cinder_manager/cloud_volume_snapshot.rb
```diff
@@ -98,26 +59,7 @@ def update_snapshot(options = {})
     raise MiqException::MiqVolumeSnapshotUpdateError, parse_error_message_from_fog_response(e), e.backtrace
   end
 
-  def delete_snapshot_queue(userid = "system", _options = {})
```
agrare, line 101: Ha I was looking for this to be added in core but see it is already there

agrare (approved):

agrare:
@kbrock looks like we just need to update the specs which pass `{}` in to `#delete_snapshot()`

kbrock:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq/pull/21801 https://github.com/ManageIQ/manageiq-api/pull/1134 https://github.com/ManageIQ/manageiq-providers-amazon/pull/755

PR #788, opened 2022-04-06
Improve error messages in verify_credentials

agrare:
Improve the error message returned in a number of cases when verifying credentials

Non-ssl to an SSL endpoint
Before:
Screenshot from 2022-04-06 12-23-32

After:
Screenshot from 2022-04-06 12-23-55

SSL Verify with self-signed cert:
Before:
Screenshot from 2022-04-06 12-23-06

After:
Screenshot from 2022-04-06 12-24-04

### DIFF
--- a/app/models/manageiq/providers/openstack/manager_mixin.rb
+++ b/app/models/manageiq/providers/openstack/manager_mixin.rb
@@ -154,7 +154,9 @@ module ManageIQ::Providers::Openstack::ManagerMixin
       when Excon::Errors::Timeout
         MiqException::MiqUnreachableError.new("Login attempt timed out")
       when Excon::Errors::SocketError
-        MiqException::MiqHostError.new("Socket error: #{err.message}")
+        MiqException::MiqHostError.new("Socket error: #{err.socket_error}")
+      when Excon::Error::BadRequest
+        MiqException::MiqHostError.new("Bad request: #{err.response.body}")
       when MiqException::MiqInvalidCredentialsError, MiqException::MiqHostError, MiqException::ServiceNotAvailable
         err
       else


app/models/manageiq/providers/openstack/manager_mixin.rb
```diff
@@ -154,7 +154,9 @@ def translate_exception(err)
       when Excon::Errors::Timeout
         MiqException::MiqUnreachableError.new("Login attempt timed out")
       when Excon::Errors::SocketError
-        MiqException::MiqHostError.new("Socket error: #{err.message}")
+        MiqException::MiqHostError.new("Socket error: #{err.socket_error}")
+      when Excon::Error::BadRequest
+        MiqException::MiqHostError.new("Bad request: #{err.response.body}")
```
agrare, line 159: NOTE: we could make the BadRequest error more readable if we parsed the XML but unsure if the response is _always_ XML or I need to handle the fallback case and just print the response body as-is, will probably go this route

Fryguy: I think this is a good start, and we can get fancy later

Fryguy (approved):

Fryguy:
Backported to `najdorf` in commit a28e5c10b364e42c320d467af32c3869948146c4.

```text
commit a28e5c10b364e42c320d467af32c3869948146c4
Date:   Wed Apr 6 13:03:38 2022 -0400

    Merge pull request #788 from agrare/improve_openstack_error_messages
    
    Improve error messages in verify_credentials
    
    (cherry picked from commit 45d4e94f1b386f48eb4f952989f6875903666ed0)
```

PR #789, opened 2022-04-06
Improve OpenStack operations error parsing

agrare:
Cloud Volume Backup Create example:
Before:
Screenshot from 2022-04-06 14-52-12

After:
Screenshot from 2022-04-06 14-53-26

I need to go through the rest of the operations and make sure we are printing the "parsed" exception to the logs otherwise we end up with:
```
[----] E, [2022-04-06T14:52:40.085584 #38769:b4280] ERROR -- evm: MIQ(cloud_volume_controller-wait_for_task): Unable to create backup for Cloud Volume "test": Expected([200, 202]) <=> Actual(400 Bad Request)
excon.error.response
  :body          => "{\"badRequest\": {\"code\": 400, \"message\": \"Invalid volume: Volume to be backed up must be available or in-use, but the current status is \\\"backing-up\\\".\"}}"
  :cookies       => [
  ]
  :headers       => {
    "Connection"             => "keep-alive"
    "Content-Length"         => "153"
    "Content-Type"           => "application/json"
    "Date"                   => "Wed, 06 Apr 2022 18:52:37 GMT"
    "Server"                 => "nginx/1.19.0"
    "x-compute-request-id"   => "req-404f445b-28f0-42bd-9c79-9263fa8f5737"
    "x-openstack-request-id" => "req-404f445b-28f0-42bd-9c79-9263fa8f5737"
  }
  :host          => "10.2.2.20"
  :local_address => "10.2.4.95"
  :local_port    => 46770
  :path          => "/v2/37f2c80029ea4056a090957ad79cddaf/backups"
  :port          => 8776
  :reason_phrase => "Bad Request"
  :remote_ip     => "10.2.2.20"
  :status        => 400
  :status_line   => "HTTP/1.1 400 Bad Request\r\n"
```

### DIFF
--- a/app/models/manageiq/providers/openstack/helper_methods.rb
+++ b/app/models/manageiq/providers/openstack/helper_methods.rb
@@ -24,8 +24,8 @@ module ManageIQ::Providers::Openstack::HelperMethods
     end
 
     def parse_error_message_from_fog_response(exception)
-      exception_string = exception.to_s
-      matched_message = exception_string.match(/message\\\": \\\"(.*)\\\", /)
+      exception_string = exception.respond_to?(:response) ? exception.response&.body : exception.to_s
+      matched_message = exception_string.match(/"message": "(.*)"/)
       matched_message ? matched_message[1] : exception_string
     end
 
@@ -68,7 +68,7 @@ module ManageIQ::Providers::Openstack::HelperMethods
         yield
       rescue => ex
         # Fog specific
-        error_message = parse_error_message_from_fog_response(ex.to_s)
+        error_message = parse_error_message_from_fog_response(ex)
         Notification.create(:type => "#{type}_error".to_sym, :options => error_options.merge(:error_message => error_message), **named_options)
         raise
       else
--- a/app/models/manageiq/providers/openstack/helper_methods.rb
+++ b/app/models/manageiq/providers/openstack/helper_methods.rb
@@ -25,8 +25,23 @@ module ManageIQ::Providers::Openstack::HelperMethods
 
     def parse_error_message_from_fog_response(exception)
       exception_string = exception.respond_to?(:response) ? exception.response&.body : exception.to_s
-      matched_message = exception_string.match(/"message": "(.*)"/)
-      matched_message ? matched_message[1] : exception_string
+
+      error_message = nil
+
+      begin
+        parsed_exception = JSON.parse(exception_string)
+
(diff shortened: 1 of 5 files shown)


app/models/manageiq/providers/openstack/helper_methods.rb
```diff
@@ -24,8 +24,8 @@ def openstack_proxy
     end
 
     def parse_error_message_from_fog_response(exception)
-      exception_string = exception.to_s
-      matched_message = exception_string.match(/message\\\": \\\"(.*)\\\", /)
+      exception_string = exception.respond_to?(:response) ? exception.response&.body : exception.to_s
+      matched_message = exception_string.match(/"message": "(.*)"/)
```
Fryguy, line 28: This match might be too greedy _or_ possibly not greed enough depending on the content.  1st question - can the message have embedded `"` characters?

Fryguy:
Noticed the body is a json response.  Wonder if it's better to try to parse it, and on failure, fallback to the regex?

agrare:
> Noticed the body is a json response. Wonder if it's better to try to parse it, and on failure, fallback to the regex?

Yeah this was actually what I wanted to try first, but depending on the error the message is nested under a different key let me find an example.  Could probably do something like `parsed_exception.dig(parsed_exception.keys.first, "message")`

app/models/manageiq/providers/openstack/helper_methods.rb
agrare: Yeah so this wasn't matching any of the exceptions that I was getting back:
```
exception.to_s.match(/message\\\": \\\"(.*)\\\", /) # The original
nil
exception_string.match(/message\\\": \\\"(.*)\\\", /) # Pulling the body out of the exception
nil
```

Here is an example of `exception.response.body`: `"{\"badRequest\": {\"code\": 400, \"message\": \"Invalid volume: Volume to be backed up must be available or in-use, but the current status is \\\"backing-up\\\".\"}}"`

Fryguy:
Maybe something like this?

```ruby
# See if the message is at the root
message = parsed_exception["message"]
# Otherwise, see if the message is in a nested hash
message ||= parsed_exception.values.detect { |v| v.include?("message") }.try(:[], "message")
# Otherwise, just use the raw exception
message ||= exception_string.match(/"message": "(.*)"/)[1]
```

PR #790, opened 2022-04-11
Fixed cloud tenant id for object store containers

GilbertCherrie:
Fixed an issue in this form where the cloud tenant selection gets reset when the user changes the name value.

Screen Shot 2022-04-11 at 3 31 17 PM

@miq-bot add_reviewer @agrare 
@miq-bot assign @agrare 
@miq-bot add-label bug


app/models/manageiq/providers/openstack/storage_manager/swift_manager/cloud_object_store_container.rb
```diff
@@ -26,7 +26,7 @@ def self.params_for_create(ems)
           :includeEmpty => true,
           :validate     => [{:type => 'required'}],
           :options      => cloud_tenants.map do |ct|
-            {:label => ct.name, :value => ct.id}
+            {:label => ct.name, :value => ct.id.to_s}
```
agrare, line 29: Interesting, not clear to me how this fixes the issue other than "DDF Magic"

agrare (approved):

agrare:
Not sure how this fixes it but can't argue with results

PR #805, opened 2022-06-21
Improve event monitor available check

agrare:
Currently sync_event_monitor_available runs in the `EventCatcher#all_valid_ems_in_zone` method which is run A. by MiqServer not in the worker zone and B. every sync_workers loop multiple times per minute.

It is more advantageous to run this in the verify_credentials method which is run on the ems_operations role in the right zone and also runs far less frequently.

We can still run `event_monitor_available` before starting the event_catcher in case something went wrong since the last verify_credentials run.

### DIFF
--- a/app/models/manageiq/providers/openstack/event_catcher_mixin.rb
+++ b/app/models/manageiq/providers/openstack/event_catcher_mixin.rb
@@ -5,7 +5,7 @@ module ManageIQ::Providers::Openstack::EventCatcherMixin
     def all_valid_ems_in_zone
       require 'manageiq/providers/openstack/legacy/openstack_event_monitor'
       super.select do |ems|
-        ems.sync_event_monitor_available?.tap do |available|
+        ems.event_monitor_available?.tap do |available|
           _log.info("Event Monitor unavailable for #{ems.name}.  Check log history for more details.") unless available
         end
       end
--- a/app/models/manageiq/providers/openstack/event_catcher_mixin.rb
+++ b/app/models/manageiq/providers/openstack/event_catcher_mixin.rb
@@ -1,7 +1,15 @@
 module ManageIQ::Providers::Openstack::EventCatcherMixin
   extend ActiveSupport::Concern
 
-  def self.all_valid_ems_in_zone
-    super.select { |ems| ems.supports?(:events) }
+  def after_initialize
+    super
+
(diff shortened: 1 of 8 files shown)


app/models/manageiq/providers/openstack/event_catcher_mixin.rb
```diff
@@ -1,14 +1,7 @@
...
-    def all_valid_ems_in_zone
-      require 'manageiq/providers/openstack/legacy/openstack_event_monitor'
-      super.select do |ems|
-        ems.sync_event_monitor_available?.tap do |available|
-          _log.info("Event Monitor unavailable for #{ems.name}.  Check log history for more details.") unless available
-        end
-      end
-    end
+  def self.all_valid_ems_in_zone
+    super.select { |ems| ems.supports?(:events) }
```
agrare, line 5: Tempted to move this check to the core EventCatcher class and add `supports :events` to all providers which have it.  Then only providers which need to run a check (e.g. openstack/azure) need an instance block like https://github.com/ManageIQ/manageiq-providers-openstack/pull/805/files#diff-6e82d48dbcba5ccd5c9dee5a5f722c52485f2c738505fe1698d54d9786e69fa0R55

app/models/manageiq/providers/openstack/manager_mixin.rb
```diff
@@ -314,16 +309,27 @@ def verify_amqp_credentials(_options = {})
...
+      verify_api_credentials
+    when :amqp
+      verify_amqp_credentials
+    else
+      verify_api_credentials
+
+      capabilities["events"] = !!event_monitor_available?
+      save! if changed?
+
+      true
```
agrare, line 332: I don't love how this looks, but the general idea is if you request a specific auth type we'll do only that, but if auth_type is nil (which is the case when doing the scheduled or from the UI authentication check) we'll check API and event_monitor.

app/models/manageiq/providers/openstack/network_manager.rb
```diff
@@ -36,6 +36,7 @@ class ManageIQ::Providers::Openstack::NetworkManager < ManageIQ::Providers::Netw
            :hostname,
            :default_endpoint,
            :endpoints,
+           :supports_events?,
```
agrare, line 39: Since event support is tied to the endpoint, any manager which delegates `verify_credentials` and `connect` should also delegate if events are supported.

Fryguy (approved):

PR #806, opened 2022-06-23
placement_group in openstack

Kuldip-Nanda:
This shows the  placement group  of a VM. It is known as server groups  or collocation rules in OpenStack terminology.
Please see associated changes in the the manageiq core for this.

1. Manageiq schema changes.
    https://github.com/ManageIQ/manageiq-schema/pull/644
2. ManageIQ Core changes
     https://github.com/ManageIQ/manageiq/pull/21807


app/models/manageiq/providers/openstack/inventory/collector/cloud_manager.rb
```diff
@@ -55,6 +55,13 @@ def host_aggregates
     @host_aggregates = safe_list { compute_service.aggregates.all }
   end
 
+  def placement_groups
+    return @placement_groups if @placement_group and @placement_groups.any?
+    @placement_groups = compute_service.handled_list(:server_groups, {}, openstack_admin?)
+    return @placement_groups
```
agrare, line 61: No need to explicitly return this, `val = ` also returns `val`

app/models/manageiq/providers/openstack/inventory/collector/cloud_manager.rb
```diff
@@ -55,6 +55,22 @@ def host_aggregates
...
+    @placement_group_by_vm_id ||= placement_groups.each_with_object({}) { |sg, result| sg.members.each { |vm_id| result[vm_id] = sg } }
+  end
+
+  def placement_groups
+    return @placement_groups if @placement_group && @placement_groups.any?
+
+    @placement_groups = compute_service.handled_list(:server_groups, {}, openstack_admin?)
+  end
+
+  def server_group_by_vm_id
```
jaywcarman, line 68: Is there a difference between placement group and server group?

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -146,6 +147,43 @@ def host_aggregates
...
+  def placement_group_by_vm_id
+    @placement_group_by_vm_id ||= collector.placement_groups.each_with_object({}) { |sg, result| sg.members.each { |vm_id| result[vm_id] = sg } }
+  end
+
+  def placement_groups
+    collector.placement_groups.each do |spgrp|
+      pgrp         = persister.placement_groups.find_or_build(spgrp.name)
+      pgrp.name    = spgrp.name
+      pgrp.ems_ref = spgrp.id
+      pgrp.policy  = spgrp.policies[0]
```
agrare, line 159: Should placement_group policies be its own table if there can be multiple?

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -146,6 +147,43 @@ def host_aggregates
...
+      #    id="1734c483-803b-4dcf-94e3-de058a6ddb87",
+      #    name="jay-collection-rule",
+      #    policies=["anti-affinity"],
+      #    members=[]
+      #  >]
+      #
+      # however like
+      # https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb#L183
+      # we are getting persister.cloud_manager.uid_ems as "default", which is not correctr.
+      # pgrp.availability_zone = persister.availability_zones.lazy_find(persister.cloud_manager.uid_ems),
```
agrare, line 182: If server groups cannot be linked to an availability zone this all should be deleted.

Is there another way to get this info?  Can you list server_group for a single availability zone for example?

app/models/manageiq/providers/openstack/inventory/collector/cloud_manager.rb
Kuldip-Nanda: no difference.

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
Kuldip-Nanda: right now it is just one either affinity or anti-affinity.

agrare: Okay can you comment here that there can only be one otherwise arbitrarily taking the first looks suspicious

app/models/manageiq/providers/openstack/inventory/collector/cloud_manager.rb
agrare: It is common practice to have the nomenclature in the Collector match the native provider, since the Parser is responsible for the translation between provider-native and MIQ-generic

app/models/manageiq/providers/openstack/inventory/parser/cloud_manager.rb
```diff
@@ -146,6 +147,43 @@ def host_aggregates
     end
   end
 
+  def placement_group_by_vm_id
+    @placement_group_by_vm_id ||= collector.placement_groups.each_with_object({}) { |sg, result| sg.members.each { |vm_id| result[vm_id] = sg } }
+  end
```
agrare, line 152: This looks like a duplicate of https://github.com/ManageIQ/manageiq-providers-openstack/pull/806/files#diff-88d95edff4569231d0581d79243806e6ffe4badd621dc136e41e18e93ce615c9R68-R72

Can you just call `collector.server_group_by_vm_id` in the parser?

```suggestion
```

agrare (approved):
We can fix the remaining comments in a follow-up

PR #807, opened 2022-06-28
Fix ceilometer endpoint creation

agrare:
The ceilometer service doesn't require a separate endpoint for connection information but we have to have some way of identifying that the event monitor type selected is ceilometer.  This has historically been done by creating a ceilometer endpoint but using the default authentication.

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager.rb
@@ -228,7 +228,6 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
                           :component    => 'protocol-selector',
                           :id           => 'event_stream_selection',
                           :name         => 'event_stream_selection',
-                          :skipSubmit   => true,
                           :initialValue => 'none',
                           :label        => _('Type'),
                           :options      => [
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/openstack/cloud_manager.rb
```diff
@@ -403,6 +403,24 @@ def self.params_for_create
...
+    end
+
+    super
+  end
+
+  def edit_with_params(params, endpoints, authentications)
+    if params.delete("event_stream_selection") == "ceilometer"
+      default_endpoint = endpoints.detect { |ep| ep["role"] == "default" }
+      endpoints << {"role" => "ceilometer", "hostname" => default_endpoint["hostname"]}
+    end
```
Fryguy, line 419: Are both of these methods the same code (aside from the super)?  If so, can we dry it up into a private method?

agrare: :+1: yeah can definitely do that

agrare: Done

Fryguy (approved):

agrare:
Hey @Fryguy this is green

PR #810, opened 2022-07-12
Convert validates to supports

kbrock:
part of https://github.com/ManageIQ/manageiq/issues/21179

converting AvaliableMixin to SupportsFeatureMixin

### DIFF
--- a/app/models/manageiq/providers/openstack/infra_manager.rb
+++ b/app/models/manageiq/providers/openstack/infra_manager.rb
@@ -27,6 +27,7 @@ class ManageIQ::Providers::Openstack::InfraManager < ManageIQ::Providers::InfraM
   supports :events do
     unsupported_reason_add(:events, _("Events are not supported")) unless capabilities["events"]
   end
+  supports_not :shutdown
 
   def self.params_for_create
     @params_for_create ||= {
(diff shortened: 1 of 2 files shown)


kbrock:
WIP: using newer supports feature for string return.
need to update if that is not going in soon.

Also double check the localization

kbrock:
un-WIP: just going with current valid syntax. can change later

app/models/manageiq/providers/openstack/infra_manager.rb
```diff
@@ -27,6 +27,7 @@ class ManageIQ::Providers::Openstack::InfraManager < ManageIQ::Providers::InfraM
   supports :events do
     unsupported_reason_add(:events, _("Events are not supported")) unless capabilities["events"]
   end
+  not_supported :shutdown
```
agrare, line 30: @kbrock does this depend on changes in core?  Or was this supposed to be `supports_not :shutdown` ?

kbrock: it was a flub. thanks

kbrock:
update:
- fixed `supports_not`
- fixed cops

agrare (approved):

PR #811, opened 2022-07-15
fix issue finding parallel

kbrock:
The legacy code requires 'parallel'
But that is not always called in the tests. Or possibly production
This requires it to make sure it is present

```
     NameError:
       uninitialized constant Parallel

rspec ./spec/legacy/openstack_handle/handle_spec.rb:27 # OpenstackHandle::Handle errors from services ignores 404 errors from services
rspec ./spec/legacy/openstack_handle/handle_spec.rb:34 # OpenstackHandle::Handle errors from services ignores 404 errors from services returning arrays
```

---

Does seem like this should not be a problem as it is required by the parent file, but alas, I'm getting these errors locally.

I'm also getting a table not found error, so maybe my database is just not working correctly.

### DIFF
--- a/lib/manageiq/providers/openstack/legacy/openstack_handle/handle.rb
+++ b/lib/manageiq/providers/openstack/legacy/openstack_handle/handle.rb
@@ -1,5 +1,6 @@
 require 'active_support/inflector'
 require 'util/miq-exception'
+require 'parallel'
 
 module OpenstackHandle
   class Handle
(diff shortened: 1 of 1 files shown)


Fryguy:
Did this gem go away when we dropped the parallel tests gem?

kbrock:
We reference the gem from the [gemspec]. So it looks like it is still around.

[gemspec]: https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/manageiq-providers-openstack.gemspec#L26

agrare (approved):

PR #812, opened 2022-07-15
Fixed cloud provider form clearable select field

GilbertCherrie:
Fixed cloud provider form `isclearable` select field bug where the entire object is returned when an option is selected instead of just the selected value.

Before:
Before

After:
After

@miq-bot add_reviewer @agrare 
@miq-bot add_reviewer @Fryguy
@miq-bot assign @agrare 
@miq-bot add-label bug

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager.rb
@@ -98,7 +98,8 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
           :name        => "provider_id",
           :label       => _("Openstack Infra Provider"),
           :isClearable => true,
-          :options     => provider_id_options
+          :simpleValue => true,
+          :options     => provider_id_options,
         },
         {
           :component    => "select",


agrare (approved):

agrare:
@GilbertCherrie was this recently introduced?  What versions are impacted by this?

agrare:
@GilbertCherrie I meant was the bug recently introduced?  I'm pretty sure selecting a provider for openstack worked with previous versions including the DDF forms

GilbertCherrie:
@agrare The DDF team told me that returning the object is actually an intended feature, not a bug, and the simpleValue prop seems to have been added almost 2 years ago in this commit: https://github.com/data-driven-forms/react-forms/commit/dea7daf3ae92190bbb70dabe87dcdc8f61ae5443.

agrare:
@GilbertCherrie okay so what was broken?  Would the provider_id not get saved properly?  It sounds like this should have been broken for years but I don't believe that is the case

GilbertCherrie:
@agrare I don't know if anything was actually broken on the backend because I have seen some APIs on other forms have been able to handle this object being passed in correctly. The issue was that the object was passed in when we only need to pass in the id value so I think making this change is just safer for the values the API expects.

PR #814, opened 2022-07-28
Record refresh VCRs rake task

nasark:
- rake task to generate refresh VCRs

@miq-bot add_labels developer, test
@miq-bot assign @agrare


lib/tasks_private/openstack.rake
```diff
@@ -34,4 +34,65 @@ namespace :vcr do
     Rake::Task['vcr:spec:run'].invoke
     Rake::Task['vcr:credentials:obfuscate'].invoke
   end
+
+  task :rerecord_v2 => [] do
```
nasark, line 38: Looking for suggestions to have a better name for the rake task

lib/tasks_private/openstack.rake
```diff
@@ -34,4 +34,65 @@ namespace :vcr do
...
+      sleep(10)
+    end
+
+    network.delete_subnet(cs.id)
+
+    10.times do
+      sleep(10)
+    end
+
+    network.delete_network(cn_id)
```
nasark, line 94: Added delays to allow the completion of deleting resources that are dependent on each other (vms, subnets, and cloud networks)

nasark:
@agrare Ready for review

lib/tasks_private/openstack.rake
```diff
@@ -34,4 +34,65 @@ namespace :vcr do
...
     Rake::Task['vcr:credentials:obfuscate'].invoke
   end
+
+  task :rerecord_v2 => [] do
+    require 'fog/openstack'
+
+    host = YAML.load_file("config/secrets.yml").dig("test", "openstack", "hostname")
+    port = YAML.load_file("config/secrets.yml").dig("test", "openstack", "port")
+    username = YAML.load_file("config/secrets.yml").dig("test", "openstack", "userid")
+    password = YAML.load_file("config/secrets.yml").dig("test", "openstack", "password")
```
agrare, line 44: This should already be handled in `spec/spec_helper.rb` here https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/spec/spec_helper.rb#L15-L22
You can access it like `Rails.application.secrets.openstack.hostname`

agrare: Just noticed this is in a rake task not a spec test, ignore ^ for now it'll be useful later

agrare: More efficient to read and parse the file once then access the hash
```suggestion
    host, port, username, password = YAML.load_file("config/secrets.yml").dig("test", "openstack").values_at("hostname", "port", "username", "password")
```

lib/tasks_private/openstack.rake
agrare: We can just replace the old `vcr:rerecord` task competely.  Using Rails secrets and VCR `config.define_cassette_placeholder` we don't need the credentials:load and credentials:obfuscate at all

lib/tasks_private/openstack.rake
```diff
@@ -34,4 +34,65 @@ namespace :vcr do
...
+
+    `rm spec/vcr_cassettes/manageiq/providers/openstack/cloud_manager/refresher.yml`
+    `rm spec/vcr_cassettes/manageiq/providers/openstack/network_manager/refresher.yml`
```
agrare, line 47: Can you reuse the https://github.com/ManageIQ/manageiq-providers-openstack/pull/814/files#diff-bafe67d15a26691f184fadd47493d35e585f1f5dc3f283c7372b091b7c6ef5feR32 rake task?

spec/models/manageiq/providers/openstack/cloud_manager/refresher_spec.rb
```diff
@@ -1,5 +1,64 @@
 describe ManageIQ::Providers::Openstack::CloudManager::Refresher do
+  include Spec::Support::EmsRefreshHelper
+
+  before(:each) do
+    host = YAML.load_file("config/secrets.yml").dig("test", "openstack", "hostname")
+    port = YAML.load_file("config/secrets.yml").dig("test", "openstack", "port")
+    username = YAML.load_file("config/secrets.yml").dig("test", "openstack", "userid")
+    password = YAML.load_file("config/secrets.yml").dig("test", "openstack", "password")
```
agrare, line 8: So _here_ is where you can use `Rails.application.secrets.openstack` and that will handle loading the defaults if secrets.yml isn't present

lib/tasks_private/openstack.rake
nasark: @agrare Does that mean we want to remove the old dependent tasks as well? Such as `lib/tasks_private/spec.rake` and `lib/tasks_private/vcr_recorder.rb`

lib/tasks_private/openstack.rake
nasark: If you're referring to https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/lib/tasks_private/openstack.rake#L32 in the link above, that would also delete all the other refresh VCRs i.e. `refresher_rhos*.yml`. The new rake task doesn't regenerate those type of VCRs and I'm not sure if there is a way to invoke the delete VCR rake task in a way that would only delete `refresher.yml` files

agrare: Yeah eventually (sooner rather than later maybe) we'll need to drop those since the first time we make a change to the openstack refresh to collect something new they will fail.

lib/tasks_private/openstack.rake
```diff
@@ -1,37 +1,88 @@
...
+    flavor = compute.flavors.get(1)
+    vm = compute.servers.create(:name => 'test-server', :flavor_ref => flavor.id, :image_ref => image.id, :nics => [{:net_id => cn_id}])
+    compute.create_key_pair("test-key")
+    vol = compute.create_volume("test-vol", "test description", 1)
+    vol_id = vol.data.dig(:body, "volume", "id")
+    agg = compute.create_aggregate("test-aggregate")
+    agg_id = agg.data.dig(:body, "aggregate", "id")
+
+    puts "Generating VCRs..."
+    `bundle exec rspec ./spec/models/manageiq/providers/openstack/cloud_manager/refresher_spec.rb`
```
agrare, line 66: I think we should put this in a `begin/rescue` block so that if we hit an exception we can still clean up created resources

spec/models/manageiq/providers/openstack/cloud_manager/refresher_spec.rb
```diff
@@ -1,5 +1,96 @@
...
+  end
+
   it ".ems_type" do
     expect(described_class.ems_type).to eq(:openstack)
   end
+
+  it "will perform a full refresh" do
+    inventory = nil
+    2.times do # Run twice to verify that a second run with existing data does not change anything
+      with_vcr { @ems.refresh }
```
agrare, line 26: We should refresh the network_manager and two storage managers here also:
```suggestion
with_vcr do
  @ems.refresh
  @ems.network_manager.refresh
  @ems.cinder_manager.refresh
  @ems.swift_manager.refresh
end
```

lib/tasks_private/openstack.rake
```diff
@@ -1,37 +1,92 @@
...
+
+    network = Fog::Network.new(
+      :provider               => "OpenStack",
+      :openstack_auth_url     => "https://#{host}:#{port}/v3/",
+      :openstack_username     => username,
+      :openstack_api_key      => password,
+      :openstack_project_name => "admin",
+      :openstack_domain_id    => "default",
+      :connection_options     => {:ssl_verify_peer => false}
+    )
```
agrare, line 44: Super minor but we could clean this up by setting something like a `connect_opts`
```suggestion
    fog_connect_opts = {
      :provider               => "OpenStack",
      :openstack_auth_url     => "https://#{host}:#{port}/v3/",
      :openstack_username     => username,
      :openstack_api_key      => password,
      :openstack_project_name => "admin",
      :openstack_domain_id    => "default",
      :connection_options     => {:ssl_verify_peer => false}
    }

    compute = Fog::Compute.new(fog_connect_opts)
    network = Fog::Network.new(fog_connect_opts)
```

lib/tasks_private/openstack.rake
```diff
@@ -1,37 +1,85 @@
...
+      flavor = compute.flavors.get(1)
+      vm = compute.servers.create(:name => 'test-server', :flavor_ref => flavor.id, :image_ref => image.id, :nics => [{:net_id => cn_id}])
+      compute.create_key_pair("test-key")
+      vol = compute.create_volume("test-vol", "test description", 1)
+      vol_id = vol.data.dig(:body, "volume", "id")
+      agg = compute.create_aggregate("test-aggregate")
+      agg_id = agg.data.dig(:body, "aggregate", "id")
+    rescue => err
+      puts err
+    end
```
agrare, line 60: It would be nice if we extracted this to a separate task like `vcr:environment:create` and then use the existing `    Rake::Task['vcr:spec:run'].invoke`, cleanup can be in a `vcr:environment:destroy`

lib/tasks_private/openstack.rake
```diff
@@ -1,37 +1,85 @@
...
+      :openstack_domain_id    => "default",
+      :connection_options     => {:ssl_verify_peer => false}
+    }
+
+    compute = Fog::Compute.new(fog_connect_opts)
+    network = Fog::Network.new(fog_connect_opts)
+
+    begin
+      puts "Creating resources..."
+      cn = network.create_network(:name => "test-network")
```
agrare, line 41: If you do
```suggestion
      cn = network.create_network(:name => "test-network").body["network"]
```
then `cn` will be the actual cloud network body not the entire response object

lib/tasks_private/openstack.rake
```diff
@@ -1,37 +1,85 @@
...
+      cn_id = cn.data.dig(:body, "network", "id")
+      tenant_id = cn.data.dig(:body, "network", "tenant_id")
```
agrare, line 43: Then these can be cleaned up a bit to be more readable
```suggestion
      cn_id = cn["id"]
      tenant_id = cn["tenant_id"]
```

lib/tasks_private/openstack.rake
```diff
@@ -1,37 +1,85 @@
...
+      cs = network.subnets.create(:network_id => cn_id, :cidr => "10.0.0.0/21", :ip_version => 4)
+      router = network.create_router("test-router")
```
agrare, line 45: Similar here
```suggestion
      router = network.create_router("test-router").body["router"]
```

lib/tasks_private/openstack.rake
```diff
@@ -1,37 +1,85 @@
...
+
+    puts "Cleaning up resources..."
+    compute.delete_server(vm.id)
+    compute.delete_key_pair("test-key")
+    compute.delete_volume(vol_id)
+    compute.delete_aggregate(agg_id)
+
+    10.times do
+      sleep(10)
+    end
```
agrare, line 73: Wonder if we should loop until `compute.get_server_details(vm.id)` etc... returns a 404 instead of sleeping for 100sec

agrare (approved):
This is a great start, all comments can be addressed in follow-up PRs

agrare:
Also before we start removing the other `refresher_rhos*` cassettes and specs (which we'll have to do before we re-record to pick up new features) we will need to re-implement those tests like targeted refresh and is_admin

PR #818, opened 2022-08-18
Add spec tests to microstack refresher spec

agrare:
Move the keystone v3 and unsupported version specs to the microstack refresher_spec.rb

This is part of the effort to move all spec tests to the microstack based VCRs rake tests so that they can be reliably re-recorded by anyone started in https://github.com/ManageIQ/manageiq-providers-openstack/pull/814


spec/models/manageiq/providers/openstack/cloud_manager/refresher_spec.rb
```diff
@@ -6,24 +6,82 @@
   end
 
   context "#refresh" do
-    before(:each) do
```
kbrock, line 9: yea, think you wanted to keep this line

agrare:
> yea, think you wanted to keep this line

No I thought it was better to use `let()` instead of a before and using 
ivars so that we could control the parameters easier

agrare:
@kbrock can you take a look?

agrare:
@kbrock :bow: thanks!

PR #819, opened 2022-08-18
Refresher spec and rerecord task enhancements

nasark:
- fleshed out refresher spec to include more scenarios
- improve rake task by modularizing code and error handling

TODO:
- [x] fix vcr generation with new scenarios
- [x] might be better to create resources in a new project
- [x] integrate `vcr:spec:run`

@miq-bot assign @agrare 
@miq-bot add_labels enhancement, test

### DIFF
--- a/lib/tasks_private/openstack.rake
+++ b/lib/tasks_private/openstack.rake
@@ -4,82 +4,113 @@ namespace :vcr do
     task :delete => :environment do
       vcr_dir = ManageIQ::Providers::Openstack::Engine.root.join("spec/vcr_cassettes/manageiq/providers/openstack/cloud_manager")
       Dir.glob(vcr_dir.join("refresher*.yml")).each { |f| File.delete(f) }
+      Dir.glob(vcr_dir.join("refresher/*.yml")).each { |f| File.delete(f) }
     end
   end
 
   namespace :spec do
     desc 'Run specs needed for rerecording of VCRs'
     task :run => :environment do
+      puts "Generating VCRs..."
       ENV['SPEC'] = Dir.glob(ManageIQ::Providers::Openstack::Engine.root.join("spec/models/manageiq/providers/openstack/cloud_manager/refresher_spec.rb")).first
       Rake::Task['spec'].invoke
     end
   end
 
-  desc 'Rerecord all of VCR cassettes'
-  task :rerecord => :environment do
-    require 'fog/openstack'
+  namespace :environment do
+    def secrets_yml
+      @secrets_yml ||= YAML.load_file("config/secrets.yml").dig("test", "openstack")
+    end
 
-    host, port, username, password = YAML.load_file("config/secrets.yml").dig("test", "openstack").values_at("hostname", "port", "userid", "password")
+    def fog_connect_opts
+      host, port, username, password = secrets_yml.values_at("hostname", "port", "userid", "password")
 
-    Rake::Task['vcr:cassettes:delete'].invoke
+      {
+        :provider               => "OpenStack",
+        :openstack_auth_url     => "https://#{host}:#{port}/v3/",
+        :openstack_username     => username,
+        :openstack_api_key      => password,
+        :openstack_project_name => "admin",
+        :openstack_domain_id    => "default",
+        :connection_options     => {:ssl_verify_peer => false}
+      }
+    end
 
-    fog_connect_opts = {
-      :provider               => "OpenStack",
-      :openstack_auth_url     => "https://#{host}:#{port}/v3/",
-      :openstack_username     => username,
-      :openstack_api_key      => password,
-      :openstack_project_name => "admin",
-      :openstack_domain_id    => "default",
-      :connection_options     => {:ssl_verify_peer => false}
-    }
+    def compute_client
+      @compute_client ||= begin
+        require 'fog/openstack'
+        Fog::Compute.new(fog_connect_opts)
+      end
+    end
 
-    compute = Fog::Compute.new(fog_connect_opts)
-    network = Fog::Network.new(fog_connect_opts)
+    def network_client
+      @network_client ||= begin
+        require 'fog/openstack'
+        Fog::Network.new(fog_connect_opts)
+      end
+    end
 
-    begin
+    def with_retry(retry_count: 10, retry_sleep: 10)
+      retry_count.times do
+        yield
+        sleep(retry_sleep)
+      rescue Fog::Compute::OpenStack::NotFound
+        break
+      end
+    end
+
+    desc 'Initialize environment with resources'
+    task :create => :environment do
       puts "Creating resources..."
-      cn = network.create_network(:name => "test-network")
-      cn_id = cn.data.dig(:body, "network", "id")
-      tenant_id = cn.data.dig(:body, "network", "tenant_id")
-      cs = network.subnets.create(:network_id => cn_id, :cidr => "10.0.0.0/21", :ip_version => 4)
-      router = network.create_router("test-router")
-      router_id = router.data.dig(:body, "router", "id")
-      sg = network.create_security_group(:name => "test-sg-group", :description => "test description", :tenant_id => tenant_id)
-      sg_id = sg.data.dig(:body, "security_group", "id")
+      cn = network_client.create_network(:name => "manageiq-spec-network").body["network"]
+      cn_id = cn["id"]
+      tenant_id = cn["tenant_id"]
+      network_client.subnets.create(:network_id => cn_id, :cidr => "10.0.0.0/21", :ip_version => 4)
+      network_client.create_router("manageiq-spec-router")
+      network_client.create_security_group(:name => "manageiq-spec-sg", :description => "test description", :tenant_id => tenant_id)
 
-      image = compute.images.first
-      flavor = compute.flavors.get(1)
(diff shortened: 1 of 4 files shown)


lib/tasks_private/openstack.rake
```diff
@@ -15,14 +15,10 @@ namespace :vcr do
     end
   end
 
-  desc 'Rerecord all of VCR cassettes'
-  task :rerecord => :environment do
+  namespace :environment do
     require 'fog/openstack'
 
     host, port, username, password = YAML.load_file("config/secrets.yml").dig("test", "openstack").values_at("hostname", "port", "userid", "password")
```
agrare, line 21: The reason the rake tasks are failing is because this code is right in the namespace so it is being evaluated when the rake file is loaded rather than just when it is being called.

You can create methods to accomplish the same thing like:
```ruby
namespace :environment do
    def secrets_yml
      @secrets_yml ||= YAML.load_file("config/secrets.yml").dig("test", "openstack")
    end

    def fog_connect_opts
      host, port, username, password = secrets_yml.values_at("hostname", "port", "userid", "password")

      {
        :provider               => "OpenStack",
        :openstack_auth_url     => "https://#{host}:#{port}/v3/",
        etc...
      }
    end

    def compute_client
      require 'fog/openstack'
      Fog::Compute.new(fog_connect_opts)
    end

    def network_client
      require 'fog/openstack'
      Fog::Network.new(fog_connect_opts)
    end
```
```

File: lib/tasks_private/openstack.rake [hunk 2]
```diff
@@ -32,54 +28,77 @@ namespace :vcr do
...
-    end
+      10.times do
+        break if vm.nil?
 
-    network.delete_subnet(cs.id)
+        compute.get_server_details(vm["id"])
+        sleep(10)
+      rescue Fog::Compute::OpenStack::NotFound
+        break
+      end
```
agrare, line 77: I think we could use a method to share this logic, similar to the `with_retry` we have in `VPC`

lib/tasks_private/openstack.rake
```diff
@@ -32,54 +28,77 @@ namespace :vcr do
...
 
-    network.delete_network(cn_id)
-    network.delete_security_group(sg_id)
-    network.delete_router(router_id)
+  desc 'Rerecord all of VCR cassettes'
+  task :rerecord => :environment do
+    Rake::Task['vcr:cassettes:delete'].invoke
+    Rake::Task['vcr:environment:create'].invoke
+    puts "Generating VCRs..."
+    `bundle exec rspec ./spec/models/manageiq/providers/openstack/cloud_manager/refresher_spec.rb`
```
agrare, line 101: Can this be `Rake::Task['vcr:spec:run'].invoke` ?

agrare: Oh I just saw you have this in the TODO list :+1:

spec/models/manageiq/providers/openstack/cloud_manager/refresher_spec.rb
```diff
@@ -14,92 +14,129 @@
...
     @ems.authentications << FactoryBot.create(:authentication, credentials)
+    @inventory = nil
+  end
+
+  def reset_cache
+    @ems.reset_openstack_handle
+
+    require "fog/openstack"
+    Fog::OpenStack.instance_variable_set(:@version, nil)
+  end
```
agrare, line 25: :+1: nice looks much nicer grouped in a method

spec/models/manageiq/providers/openstack/cloud_manager/refresher_spec.rb
```diff
@@ -14,92 +14,129 @@
...
 
-        @ems.refresh
-        @ems.network_manager.refresh
-        @ems.cinder_manager.refresh
-        @ems.swift_manager.refresh
+  context "when using an admin account for fast refresh" do
+    it "will perform a fast full refresh" do
+      ::Settings.ems_refresh.openstack.is_admin = true
+      ::Settings.ems_refresh.openstack_network.is_admin = true
```
agrare, line 63: I know this is how the current test is written but it is much better to use stub_settings_merge:
```ruby
```suggestion
  context "when using an admin account for fast refresh" do
    before do
      stub_settings_merge(
        :ems_refresh => {
          :openstack => {:is_admin => true},
          :openstack_network => {:is_admin => true}
        }
      )
    end
    
    it "will perform a fast full refresh" do
```

And then you don't have to set it back down [[here]](https://github.com/ManageIQ/manageiq-providers-openstack/pull/819/files#diff-47be47aae1eb015036827f5434eed1ba3b3981cc68ad9e3967cf01311cfbfd37R67-R68)

agrare: The reason being messing with the global state is a recipe for sporadic test failures

spec/models/manageiq/providers/openstack/cloud_manager/refresher_spec.rb
```diff
@@ -14,92 +14,129 @@
...
+    assert_specific_cloud_subnet
+    assert_specific_security_group
+    assert_specific_network_router
   end
 
   it ".ems_type" do
     expect(described_class.ems_type).to eq(:openstack)
   end
 
   it "will perform a full refresh" do
```
agrare, line 54: So I think it'd be good to group all of these in a `context "#refresh"` since the `.ems_type` is testing a completely different method.  We can move the code that sets up the ems with the secrets and the other helper methods into this context since they aren't needed by other tests.

spec/models/manageiq/providers/openstack/cloud_manager/refresher_spec.rb
```diff
@@ -14,92 +14,129 @@
...
+
+  def reset_cache
+    @ems.reset_openstack_handle
+
+    require "fog/openstack"
+    Fog::OpenStack.instance_variable_set(:@version, nil)
+  end
+
+  def refresh_inventory
+    with_vcr do
```
agrare, line 28: We're going to need to take a parameter to pass in as the suffix since the different tests will need different VCRs (e.g. the admin account "fast" refresh will make different API calls and will need its own VCR cassette.

```suggestion
  def refresh_inventory(vcr_suffix = nil)
    with_vcr(vcr_suffix) do
```

spec/models/manageiq/providers/openstack/cloud_manager/refresher_spec.rb
```diff
@@ -14,92 +14,129 @@
...
+    @ems.reload
+
+    assert_specific_vm
+    assert_specific_auth_key_pair
+    assert_specific_cloud_volume
+    assert_specific_host_aggregates
+    assert_specific_cloud_network
+    assert_specific_cloud_subnet
+    assert_specific_security_group
+    assert_specific_network_router
```
agrare, line 47: I wonder if we should split up the "refresh_inventory" from the checks since different tests might want to check different things.  We can still move all of these into an `assert_common` to reduce the duplication like you've done here

nasark: > we should split up the "refresh_inventory" from the checks **since different tests might want to check different things**

Wouldn't moving all the checks into an `assert_common` prevent this? Since all the checks are in one place

nasark: Also not sure how we would go about asserting certain checks while leaving out others depending on the test

spec/models/manageiq/providers/openstack/cloud_manager/refresher_spec.rb
nasark: Would all the refresh tests go in `context "#refresh"` as well?

agrare: Yes

spec/models/manageiq/providers/openstack/cloud_manager/refresher_spec.rb
```diff
@@ -47,61 +53,102 @@
...
-    expect(cs.network_protocol).to eq("ipv4")
-    expect(cs.type).to eq("ManageIQ::Providers::Openstack::NetworkManager::CloudSubnet")
-  end
+    def assert_specific_cloud_subnet
+      cn = @ems.cloud_networks.find_by(:name => "manageiq-spec-network")
+      cs = @ems.cloud_subnets.find_by(:cloud_network_id => cn.id)
+      expect(cs.cidr).to eq("10.0.0.0/21")
+      expect(cs.network_protocol).to eq("ipv4")
+      expect(cs.type).to eq("ManageIQ::Providers::Openstack::NetworkManager::CloudSubnet")
+    end
```
nasark, line 96: We were actually asserting the wrong subnet here before (subnet created in rake task doesn't have a name)

nasark:
@agrare Ready for another look

lib/tasks_private/openstack.rake
```diff
@@ -4,82 +4,113 @@ namespace :vcr do
     task :delete => :environment do
       vcr_dir = ManageIQ::Providers::Openstack::Engine.root.join("spec/vcr_cassettes/manageiq/providers/openstack/cloud_manager")
       Dir.glob(vcr_dir.join("refresher*.yml")).each { |f| File.delete(f) }
+      Dir.glob(vcr_dir.join("refresher/*.yml")).each { |f| File.delete(f) }
     end
```
agrare, line 8: Think we can do this in one line with `**/refresher*.yml` but this is good for now

agrare (approved):

PR #822, opened 2022-08-26
Add microstack target refresher specs

nasark:
- for microstack setups, based on https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/spec/models/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty_spec.rb#L56

@miq-bot assign @agrare 
@miq-bot add_labels enhancement, test


lib/tasks_private/openstack.rake
```diff
@@ -75,6 +82,11 @@ namespace :vcr do
       compute_client.create_key_pair("manageiq-spec-key")
       compute_client.create_volume("manageiq-spec-vol", "test description", 1)
       compute_client.create_aggregate("manageiq-spec-aggregate")
+
+      project = identity_client.create_project(:name => 'manageiq-spec-project', :description => "test description").body["project"]
+      role = identity_client.list_roles.body["roles"].find { |roles| roles["name"] == "admin" }
+      user = identity_client.list_users.body["users"].find { |users| users["name"] == "admin" }
+      identity_client.grant_project_user_role(project["id"], user["id"], role["id"])
```
nasark, line 89: This is necessary in order for the tenant to be picked up during a refresh

agrare:
This looks great, I think we can delete https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/spec/models/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty_spec.rb now

PR #833, opened 2022-12-07
Install openssl package to fix qpid-proton build

agrare:


bin/before_install
```diff
@@ -10,29 +10,13 @@ if [ -n "$CI" ]; then
   set -v
   pushd $PWD
 
-    # Install the dev dependencies for building Qpid proton system library.
-    sudo apt-get install -y gcc cmake cmake-curses-gui uuid-dev
-    sudo apt-get install -y libssl-dev
-    sudo apt-get install -y libsasl2-2 libsasl2-dev
+  cd /tmp
+  wget http://mirrors.kernel.org/ubuntu/pool/universe/q/qpid-proton/libqpid-proton11-dev_0.37.0-1_amd64.deb
+  wget http://mirrors.kernel.org/ubuntu/pool/universe/q/qpid-proton/libqpid-proton11_0.37.0-1_amd64.deb
```
agrare, line 15: NOTE this is _not_ the right way to do this.  GHA runs on ubuntu 22.04 jammy which has `libqpid-proton11 0.22.0-5.1ubuntu2` which is too old for our purpose.  The newer ubuntu 22.10 kinetic has `libqpid-proton11 0.37.0-1` which does work for us.

I wasn't able to successfully build from source qpid-proton older than 0.38, and 0.38 is missing the ruby gem.

I'm going to try:
1. Installing the 0.38.0 ruby gem that gets built from source when we build qpid-proton
2. Using AptPinning and adding the kinetic repo to install the newer libqpid-proton11 package properly

PR #835, opened 2023-01-11
Fix port for class-level verify_credentials

agrare:
The class-level verify_credentials method takes DDF parameters with the port being an integer.  The ems_connect? method was assuming the port was a string so it was doing `params[:port].try(:strip)` which was returning `nil` for an integer.  This then used the default port `5000` since no port appeared to be passed.

The result is the class-level verify succeeds, then the instance-level fails.

### DIFF
--- a/app/models/manageiq/providers/openstack/manager_mixin.rb
+++ b/app/models/manageiq/providers/openstack/manager_mixin.rb
@@ -46,7 +46,7 @@ module ManageIQ::Providers::Openstack::ManagerMixin
       ems.security_protocol      = params[:security_protocol].strip
       ems.keystone_v3_domain_id  = params[:uid_ems]
 
-      user, hostname, port = params[:userid], params[:hostname].strip, params[:port].try(:strip)
+      user, hostname, port = params[:userid], params[:hostname].strip, params[:port]
 
       endpoint = {:role => :default, :hostname => hostname, :port => port, :security_protocol => ems.security_protocol}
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/openstack/manager_mixin.rb
```diff
@@ -46,7 +46,7 @@ def ems_connect?(password, params, service)
       ems.security_protocol      = params[:security_protocol].strip
       ems.keystone_v3_domain_id  = params[:uid_ems]
 
-      user, hostname, port = params[:userid], params[:hostname].strip, params[:port].try(:strip)
+      user, hostname, port = params[:userid], params[:hostname].strip, params[:port]
```
agrare, line 49: This is only called by the `verify_credentials?` method which takes DDF params, if a user does pass in a string port for some reason this still works

Fryguy: Out of curiosity, if the port is a string and has, say, leading spaces, will this still work?

agrare: No, but you also can't provider a port with a leading space in the ddf form

Fryguy:
Backported to `oparin` in commit 671510ee5294986ad3f08beb5e2705c3ad852a31.

```text
commit 671510ee5294986ad3f08beb5e2705c3ad852a31
Date:   Wed Jan 11 17:20:25 2023 -0500

    Merge pull request #835 from agrare/fix_verify_credentials
    
    Fix port for class-level verify_credentials
    
    (cherry picked from commit cd4cc159417db4d48568fe68de86dfe319a9f76a)
```

PR #842, opened 2023-03-16
Provider validate credentials

kbrock:
We hashed around some ideas for provider's validate_credentials

- Wanted to get Compute and Baremetal into there
- Wanted the logic for each of those to be specific
- I didn't touch the auth_type.nil? case - we can discuss.
- I didn't handle cinder or swift managers. Anything special there?
- I did not see an issue with `ActiveSupport::Concerns`, but it is wrong and was bugging me

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager.rb
@@ -503,6 +503,17 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
     vs&.name == :swift ? vs : nil
   end
 
+  def verify_credentials(auth_type = nil, options = {})
+    options[:service] ||= "Compute"
+    ret = super
+    return ret unless auth_type.nil?
+
+    capabilities["events"] = !!event_monitor_available?
+    save! if changed?
+
+    true
+  end
+
   def self.ems_type
     @ems_type ||= "openstack".freeze
   end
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/openstack/cloud_manager.rb
```diff
@@ -503,6 +503,17 @@ def swift_service
...
 
+  def verify_credentials(auth_type = nil, options = {})
+    options[:service] ||= "Compute"
+    ret = super
+    return ret unless auth_type.nil?
+
+    capabilities["events"] = !!event_monitor_available?
+    save! if changed?
+
+    true
```
kbrock, line 514: this is where the manager_mixin `else` went

app/models/manageiq/providers/openstack/infra_manager/host/operations.rb
```diff
@@ -1,5 +1,5 @@
 module ManageIQ::Providers::Openstack::InfraManager::Host::Operations
-  include ActiveSupport::Concern
+  extend ActiveSupport::Concern
```
kbrock, line 2: this is a tangent.
I think the existing one is a bug but can pull into another PR if you want

app/models/manageiq/providers/openstack/manager_mixin.rb
```diff
@@ -318,18 +317,16 @@ def verify_event_monitor(_options = {})
   end
 
   def verify_credentials(auth_type = nil, options = {})
-    case auth_type
-    when :default
-      verify_api_credentials
-    when :amqp
-      verify_amqp_credentials
-    else
```
kbrock, line 326: I'm pretty sure the `else` is for a `nil` and not for a different `auth_type`.

agrare (approved):

PR #846, opened 2023-04-11
Drop validate_credentials_task

kbrock:
The UI is calling `verify_credentials_task` and not `validate_credentials_task`.

If we want this functionality, we need to move to `verify_credentials`

I have not tested this at all.
More of a conversation starter.

### DIFF
--- a/app/models/manageiq/providers/openstack/infra_manager.rb
+++ b/app/models/manageiq/providers/openstack/infra_manager.rb
@@ -414,44 +414,8 @@ class ManageIQ::Providers::Openstack::InfraManager < ManageIQ::Providers::InfraM
     n_('Infrastructure Provider (OpenStack)', 'Infrastructure Providers (OpenStack)', number)
   end
 
-  # For infra, validate primary endpoint *and* verify presence of ironic
-  def self.validate_credentials_task(args, user_id, zone)
-    task_opts = {
-      :action => "Validate EMS Provider Credentials",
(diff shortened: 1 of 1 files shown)


agrare:
:+1: I don't see this being called
```
$ grep -r validate_credentials_task
manageiq-providers-openstack/app/models/manageiq/providers/openstack/infra_manager.rb:  def self.validate_credentials_task(args, user_id, zone)
manageiq/app/models/mixins/verify_credentials_mixin.rb:    def validate_credentials_task(args, user_id, zone)
manageiq/spec/models/mixins/authentication_mixin_spec.rb:    context ".validate_credentials_task" do
manageiq/spec/models/mixins/authentication_mixin_spec.rb:        expect(ExtManagementSystem.validate_credentials_task(args, 'userid', 'zone')).to eq([true, nil])
manageiq/spec/models/mixins/authentication_mixin_spec.rb:        expect(ExtManagementSystem.validate_credentials_task(args, 'userid', 'zone')).to eq([false, message])
```

https://github.com/search?q=org%3AManageIQ%20validate_credentials_task&type=code

agrare:
The last caller appears to have been removed by https://github.com/ManageIQ/manageiq-ui-classic/pull/7251/files#diff-33d80cb50caa7953c9054b101eab437d867226aa8c7b4eb139eee4a974029a3fL87

agrare (approved):

PR #863, opened 2023-10-30
Fix Gem.loaded_specs looking for wrong key name

agrare:
`Gem.loaded_specs` has string keys which was causing this check to
always return false

```
(byebug) Gem.loaded_specs.key?(:qpid_proton)
false
(byebug) Gem.loaded_specs.key?("qpid_proton")
true
```


agrare:
Found while testing https://github.com/ManageIQ/manageiq-providers-nuage/pull/299

jrafanie:
Wow, nice find @agrare.  I wonder when this changed.

jrafanie (approved):
LGTM

PR #865, opened 2023-11-15
Psych 4 support

jrafanie:
Part of https://github.com/ManageIQ/manageiq/issues/22696

### DIFF
--- a/app/models/manageiq/providers/openstack/helper_methods.rb
+++ b/app/models/manageiq/providers/openstack/helper_methods.rb
@@ -64,6 +64,11 @@ module ManageIQ::Providers::Openstack::HelperMethods
     end
 
     def with_notification(type, options: {})
+      # We're explicitly serializing objects in the options subject key so we should permit yaml loading those classes
+      if options[:subject]
+        ActiveRecord::Base.yaml_column_permitted_classes = ActiveRecord::Base.yaml_column_permitted_classes | [options[:subject].class]
+      end
+
       # extract success and error options from options
       # :success and :error keys respectively
       # with all other keys common for both cases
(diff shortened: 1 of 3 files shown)


spec/models/manageiq/providers/openstack/cloud_manager/provision/volume_attachment_spec.rb
```diff
@@ -5,6 +5,8 @@
     @flavor = FactoryBot.create(:flavor_openstack)
     @volume = FactoryBot.create(:cloud_volume_openstack)
 
+    # We're storing objects in the instance_type, so we must permit loading this class
+    ActiveRecord::Base.yaml_column_permitted_classes = ActiveRecord::Base.yaml_column_permitted_classes | [@flavor.class]
```
jrafanie, line 9: I'm not a big fan of this but if we use YAML for serialized columns and want to use safe_load, we need to explicitly allow classes to be YAML.loaded at the point we deserialize them... or we have to populate this with all possible models.

app/models/manageiq/providers/openstack/helper_methods.rb
```diff
@@ -64,6 +64,11 @@ def parse_error_message_excon_http_status(exception)
     end
 
     def with_notification(type, options: {})
+      # We're explicitly serializing objects in the options subject key so we should permit yaml loading those classes
```
agrare, line 67: Wow really?  We should definitely not be doing this, I'll look into this after

spec/models/manageiq/providers/openstack/cloud_manager/provision/volume_attachment_spec.rb
agrare: Agreed AR instances in MiqRequestTask options is really terrible, but it is everywhere

agrare (approved):
I want to fix that Notification subject object issue but will do that as a follow-up

app/models/manageiq/providers/openstack/helper_methods.rb
agrare: https://github.com/ManageIQ/manageiq-providers-openstack/pull/866

PR #866, opened 2023-11-17
Don't store AR objects in Notification options

agrare:
The `:subject`, `:initiator`, and `:cause` attributes are all `belongs_to` associations on the `Notification` model and should not be stored in the `:options` yaml column.

This will need a data migration to clean up existing records

Ref: https://github.com/ManageIQ/manageiq-providers-openstack/pull/865#discussion_r1397169505

### DIFF
--- a/app/models/manageiq/providers/openstack/helper_methods.rb
+++ b/app/models/manageiq/providers/openstack/helper_methods.rb
@@ -64,11 +64,6 @@ module ManageIQ::Providers::Openstack::HelperMethods
     end
 
     def with_notification(type, options: {})
-      # We're explicitly serializing objects in the options subject key so we should permit yaml loading those classes
-      if options[:subject]
-        ActiveRecord::Base.yaml_column_permitted_classes = ActiveRecord::Base.yaml_column_permitted_classes | [options[:subject].class]
-      end
-
       # extract success and error options from options
       # :success and :error keys respectively
       # with all other keys common for both cases
@@ -77,12 +72,8 @@ module ManageIQ::Providers::Openstack::HelperMethods
       success_options.merge!(options)
       error_options.merge!(options)
 
-      # copy subject, initiator and cause from options
-      named_options_keys = [:subject, :initiator, :cause]
-      named_options = {}
-      named_options_keys.map do |key|
-        named_options[key] = options.fetch(key, nil)
-      end
+      # extract subject, initiator and cause from options
+      named_options = options.extract!(:subject, :initiator, :cause)
 
       begin
         yield


app/models/manageiq/providers/openstack/helper_methods.rb
```diff
@@ -77,11 +72,11 @@ def with_notification(type, options: {})
       success_options.merge!(options)
       error_options.merge!(options)
 
-      # copy subject, initiator and cause from options
+      # extract subject, initiator and cause from options
       named_options_keys = [:subject, :initiator, :cause]
       named_options = {}
       named_options_keys.map do |key|
-        named_options[key] = options.fetch(key, nil)
+        named_options[key] = options.delete(key)
```
jrafanie, line 79: Wow, we weren't even using it? Crazy.  Nice find.

agrare: Yeah in fact `options[:subject]` is supposed to be a backup copy of the `subject.name` in the event that the actual subject is deleted.

I'm going to have to check but it looks like this might overwrite the `options[:subject]` object in a before_save hook (though we'd still have an issue with the initiator and cause)

jrafanie: Sounds good. I vaguely recall the `subject.name` backup name thing and part of the problem was there were different clients creating these notifications so you'd have to remove passing the subject for each or at least try to normalize them to not use it and only provide a possible identifier for the subject of the notification

agrare: The subject backup is just taking it from the `subject_type/subject_id` polymorphic assoc so that should work for any/all callers

Fryguy: I think this entire thing simplifies to

```diff
-       # copy subject, initiator and cause from options
-       named_options_keys = [:subject, :initiator, :cause]
-       named_options = {}
-       named_options_keys.map do |key|
-         named_options[key] = options.fetch(key, nil)
+       named_options = options.extract!(:subject, :initiator, :cause)
```

agrare: Okay I checked all of the callers and none of them pass in initiator or cause, only subject and that is changed by the pre-save hook to be a simple name.  I don't think we need a schema migration since there are no notifications which would have serialized AR objects in `options` at least for openstack.

jrafanie (approved):
LGTM.  I'm glad to see we don't even need the options to have the subject (or initiator or cause) and we can just drop it entirely.  Less is more!

jrafanie:
@agrare do you have any concerns with merging this?  I think you said :shipit: but wasn't sure.

Fryguy (approved):

agrare:
No concerns, go for it

PR #877, opened 2024-03-12
Only check for duplicate endpoints for CloudManagers

agrare:
Fix hostname_uniqueness_valid? failing if an `Openstack::InfraManager` has the same connection details as an `Openstack::CloudManager`

image

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager.rb
@@ -518,14 +518,15 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
     return unless hostname_required?
     return unless hostname.present? # Presence is checked elsewhere
 
-    existing_providers = Endpoint.where(:hostname => hostname.downcase)
-                                 .where.not(:resource_id => id).includes(:resource)
-                                 .select do |endpoint|
-                                   unless endpoint.resource.nil?
-                                     endpoint.resource.uid_ems == keystone_v3_domain_id &&
-                                       endpoint.resource.provider_region == provider_region
-                                   end
-                                 end
+    existing_providers =
+      self.class
+          .joins(:endpoints)
+          .where.not(:id => id)
+          .where(
+            :endpoints       => {:hostname => hostname.downcase},
+            :uid_ems         => keystone_v3_domain_id,
+            :provider_region => provider_region
+          )
 
     errors.add(:hostname, "has already been taken") if existing_providers.any?
   end
--- a/spec/models/manageiq/providers/openstack/cloud_manager_spec.rb
+++ b/spec/models/manageiq/providers/openstack/cloud_manager_spec.rb
@@ -225,65 +225,68 @@ describe ManageIQ::Providers::Openstack::CloudManager do
   end
 
   context "validation" do
-    before :each do
-      @ems = FactoryBot.create(:ems_openstack_with_authentication)
-      require 'manageiq/providers/openstack/legacy/openstack_event_monitor'
-    end
+    let!(:ems) { FactoryBot.create(:ems_openstack_with_authentication) }
 
     it "verifies AMQP credentials" do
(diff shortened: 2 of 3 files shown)


spec/models/manageiq/providers/openstack/cloud_manager_spec.rb
```diff
@@ -225,65 +225,77 @@ def lastlog
   end
 
   context "validation" do
-    before :each do
-      @ems = FactoryBot.create(:ems_openstack_with_authentication)
-      require 'manageiq/providers/openstack/legacy/openstack_event_monitor'
-    end
+    let!(:ems) { FactoryBot.create(:ems_openstack_with_authentication) }
```
agrare, line 228: Did some refactoring here to make it cleaner to override the existing ems, also the event_monitor_available? tests were mixed in to the hostname_uniqueness_valid? tests in the same context.

Refactoring is in a separate commit.

spec/models/manageiq/providers/openstack/cloud_manager_spec.rb
```diff
@@ -225,65 +225,77 @@ def lastlog
...
+      end
+
+      context "with an InfraManager" do
+        let!(:ems) { FactoryBot.create(:ems_openstack_infra) }
+
+        it "passes uniqueness check with with same hostname domain and region" do
+          new_ems = FactoryBot.build(:ems_openstack_with_authentication, :hostname => ems.hostname, :uid_ems => ems.uid_ems, :provider_region => ems.provider_region)
+          expect(new_ems.valid?).to be_truthy
+        end
+      end
```
agrare, line 298: This is the added test that fails without the hostname_uniqueness_valid? change
```
Failures:

  1) ManageIQ::Providers::Openstack::CloudManager validation #hostname_uniqueness_valid? with an InfraManager passes uniqueness check with with same hostname domain and region
     Failure/Error: expect(new_ems.valid?).to be_truthy
     
       expected: truthy value
            got: false
     # ./spec/models/manageiq/providers/openstack/cloud_manager_spec.rb:296:in `block (5 levels) in <main>'
     # /home/grare/adam/.gem/ruby/3.1.0/gems/webmock-3.23.0/lib/webmock/rspec.rb:39:in `block (2 levels) in <main>'

Finished in 3.94 seconds (files took 4.54 seconds to load)
29 examples, 1 failure

Failed examples:

rspec ./spec/models/manageiq/providers/openstack/cloud_manager_spec.rb:294 # ManageIQ::Providers::Openstack::CloudManager validation #hostname_uniqueness_valid? with an InfraManager passes uniqueness check with with same hostname domain and region
```

app/models/manageiq/providers/openstack/cloud_manager.rb
```diff
@@ -518,14 +518,15 @@ def hostname_uniqueness_valid?
     return unless hostname_required?
     return unless hostname.present? # Presence is checked elsewhere
 
-    existing_providers = Endpoint.where(:hostname => hostname.downcase)
-                                 .where.not(:resource_id => id).includes(:resource)
-                                 .select do |endpoint|
-                                   unless endpoint.resource.nil?
-                                     endpoint.resource.uid_ems == keystone_v3_domain_id &&
-                                       endpoint.resource.provider_region == provider_region
```
agrare, line 526: Since all the rest of these checks are off of endpoint.resource I thought a `self.class.join(:endpoints)` would be cleaner

Fryguy:
Backported to `radjabov` in commit 888a39f9def61ae3585602fe628a815b3ed4211a.

```text
commit 888a39f9def61ae3585602fe628a815b3ed4211a
Author: Keenan Brock <keenan@thebrocks.net>
Date:   Thu Mar 21 12:07:06 2024 -0400

    Merge pull request #877 from agrare/dont_include_infra_manager_hostname_duplicate_check
    
    Only check for duplicate endpoints for CloudManagers
    
    (cherry picked from commit 776ba68dded6736dc73a25f97e01ccf741dd6ceb)
```

PR #880, opened 2024-04-04
fix spec for supports_feature change

kbrock:
part of:
- https://github.com/ManageIQ/manageiq/pull/22898

Since we will be no longer storing the `unsupported_reason` in a temporary variable,
calling `supports?()` and `unsupported_reason()` each check the `supports :feature` code.

So in this test, it calls it 2 times.

I removed the stub to allow the ems to be called multiple times without freaking a stub out
Alternative would have been to add to the ems stub: `at_least(1).times`

### DIFF
--- a/spec/models/manageiq/providers/openstack/storage_manager/cinder_manager/cloud_volume_spec.rb
+++ b/spec/models/manageiq/providers/openstack/storage_manager/cinder_manager/cloud_volume_spec.rb
@@ -20,13 +20,11 @@ describe ManageIQ::Providers::Openstack::StorageManager::CinderManager::CloudVol
 
   let(:raw_volumes) do
     double.tap do |volumes|
-      handle = double
-      allow(handle).to receive(:volumes).and_return(volumes)
       allow(ExtManagementSystem).to receive(:find).with(ems.id).and_return(ems)
       allow(ExtManagementSystem).to receive(:find).with(ems.parent_manager.id).and_return(ems.parent_manager)
       allow(ems.parent_manager).to receive(:connect)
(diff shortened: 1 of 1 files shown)


spec/models/manageiq/providers/openstack/storage_manager/cinder_manager/cloud_volume_spec.rb
```diff
@@ -86,7 +86,7 @@
       end
 
       it "validates the volume update operation when ems is missing" do
-        expect(cloud_volume).to receive(:ext_management_system).and_return(nil)
+        expect(cloud_volume).to receive(:ext_management_system).at_least(1).times.and_return(nil)
```
agrare, line 89: Think it would be better to add a context and set up the cloud_volume to not have an ems_id?  Stubbing association calls seems more fragile
```
context "with a cloud_volume without an EMS" do
  let(:cloud_volume) { FactoryBot.create(:cloud_volume_openstack) }

  it "" 
    etc...
end
```

Fryguy: This association stubbing was there before this PR, though, right?

agrare: Yes

kbrock: I'll fix. thanks

agrare (approved):

PR #885, opened 2024-06-28
Provide backward compatible way to globally set permitted classes

jrafanie:
This supports rails 7 and 6.1.  Similar to the change in: https://github.com/ManageIQ/manageiq/pull/22887


jrafanie:
@Fryguy this is ready to go.  I'll come back and clean these up in the follow task: https://github.com/ManageIQ/manageiq/issues/22052.  When we drop rails 6.1, it should be easy to mass remove them.

Fryguy (approved):

Issue #892, opened 2024-10-09
Issue in provisioning instance with zero root disk flavor in OpenStack


Dhamo1107:
### **OpenStack Behavior:**
In OpenStack, while launching an instance with a zero root disk flavor, the following options are available under the "Source" tab:

- Select Boot Source:

    - Options: Image, Instance Snapshot, Volume, Volume Snapshot

- Create New Volume: Yes/No

    - If "Yes" is selected, additional fields appear:

        - Volume Size (GB): (Field to input volume size in GB)

        - Delete Volume on Instance Delete: Yes/No

For example, if "Create New Volume" is selected as "Yes" and a volume size of 1GB is specified, this volume is treated as the root disk.
![image](https://github.com/user-attachments/assets/7866fca2-4bde-47fa-a47c-fb43813b25b7)
![image](https://github.com/user-attachments/assets/59c53274-32c8-4ac5-b6a6-548933a21c36)

### **Issue in ManageIQ:**
In ManageIQ, there seems to be no option to specify a root disk size while provisioning an instance with a zero root disk flavor. If we try to add a volume using the "Add Volume" option in the instance provisioning page, the volume is attached as an additional volume instead of being treated as the root disk.
![image](https://github.com/user-attachments/assets/41098d09-a4f5-4d47-b760-255a4ab6f72e)
![image](https://github.com/user-attachments/assets/098458fb-0c2e-44ae-a833-069173db333e)

**Request:**
Could you guide us on how to add root disk space while provisioning an instance with a zero root disk flavor type in ManageIQ? If this feature does not exist, would it be possible to implement a way to handle root disk size in this scenario?

**Steps to Reproduce:**
Provision an instance with a zero root disk flavor in OpenStack via ManageIQ.
Attempt to add a root disk volume during provisioning.

**Expected Behavior:**
The ability to specify and add root disk size during instance provisioning with zero root disk flavor in ManageIQ, similar to the OpenStack behavior.

**Actual Behavior:**
The "Add Volume" option adds an additional volume but does not handle root disk size.

**ManageIQ version: Petrosian**

Dhamo1107:
Hi @agrare,

I would like to know if there are any updates on this issue. Please let me know if further information is needed from my side.

Thank you!

Dhamo1107:
Hi @agrare,

By using OpenStack's "Create New Volume" option, the volume is added under `/vda`.

[...]

PR #894, opened 2024-10-19
Add the volume to the root disk (vda) if the selected flavor has a ro…

Dhamo1107:
This Pull Request addresses the handling of root disk provisioning when using flavors with a root disk size of 0. Specifically, if a flavor has a root disk size of 0, the volume provided during provisioning will be designated as the root disk (vda). Conversely, if the flavor has a non-zero root disk size (e.g., 100GB), the specified volume will be treated as an additional volume (vdb). 

  
   Before the change:
   - Volume specified during provisioning would not be correctly set as the root disk when using a 0 root disk size flavor (10GB volume specified on the add volume page is assigned under vdb).
      image

   After the change:
   - Volume specified during provisioning is correctly assigned as the root disk (vda) if the flavor's root disk size is 0 (25GB volume specified on the add volume page is assigned under vda).
image

   - Additional volumes specified will automatically come under vdb when the flavor has a root disk size greater than 0 (With a 100GB root disk size flavor, the root disk is assigned under vda, while the 50GB volume specified on the add volume page is assigned under vdb).
image

   Fixes [#892]

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/provision/volume_attachment.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/provision/volume_attachment.rb
@@ -1,10 +1,20 @@
 module ManageIQ::Providers::Openstack::CloudManager::Provision::VolumeAttachment
   def create_requested_volumes(requested_volumes)
-    volumes_attrs_list = [default_volume_attributes]
+    volumes_attrs_list = instance_type.root_disk_size > 0 ? [default_volume_attributes] : []
 
     connection_options = {:service => "volume", :tenant_name => cloud_tenant.try(:name)}
     source.ext_management_system.with_provider_connection(connection_options) do |service|
-      requested_volumes.each do |volume_attrs|
+      requested_volumes.each_with_index do |volume_attrs, index|
+        if instance_type.root_disk_size == 0 # Handle root disk logic only when root_disk_size is 0
+          if index == 0  # The first volume should be the root disk, typically assigned to vda
+            volume_attrs[:imageRef] = source.ems_ref # Set the image reference for booting
+            volume_attrs[:bootable] = true
+            volume_attrs[:boot_index] = 0
+          else # Subsequent volumes will be assigned to vdb, vdc, etc.
+            volume_attrs[:boot_index] = -1
+          end
+        end
+
         new_volume_id = service.volumes.create(volume_attrs).id
         new_volume_attrs = volume_attrs.clone
         new_volume_attrs[:uuid]             = new_volume_id
--- a/app/models/manageiq/providers/openstack/cloud_manager/provision/volume_attachment.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/provision/volume_attachment.rb
@@ -10,8 +10,6 @@ module ManageIQ::Providers::Openstack::CloudManager::Provision::VolumeAttachment
             volume_attrs[:imageRef] = source.ems_ref # Set the image reference for booting
             volume_attrs[:bootable] = true
             volume_attrs[:boot_index] = 0
-          else # Subsequent volumes will be assigned to vdb, vdc, etc.
-            volume_attrs[:boot_index] = -1
           end
         end
 
--- a/spec/models/manageiq/providers/openstack/cloud_manager/provision/volume_attachment_spec.rb
+++ b/spec/models/manageiq/providers/openstack/cloud_manager/provision/volume_attachment_spec.rb
@@ -29,6 +29,30 @@ describe ManageIQ::Providers::Openstack::CloudManager::Provision::VolumeAttachme
 
       expect(task.create_requested_volumes(task.options[:volumes])).to eq([default_volume, requested_volume])
     end
+
+    context "with a flavor that has no root disk" do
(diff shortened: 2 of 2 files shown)


agrare:
Specs look like they fail to start due to https://github.com/ManageIQ/manageiq-schema/pull/757 this should be resolved shortly

agrare:
Since I have significant amount of code in here now I'm going to assign to @Fryguy 

@Fryguy I refactored the openstack volume_attachment specs to make it easier to write the following tests in a separate commit (https://github.com/ManageIQ/manageiq-providers-openstack/pull/894/commits/f52ab04199bbb361051fec2e9c3104de8f7f6f75)
Then this commit adds tests (https://github.com/ManageIQ/manageiq-providers-openstack/pull/894/commits/835a879482366a721d9a0429d1ffd47613c05a9a)

app/models/manageiq/providers/openstack/cloud_manager/provision/volume_attachment.rb
```diff
@@ -1,10 +1,20 @@
...
   def create_requested_volumes(requested_volumes)
-    volumes_attrs_list = [default_volume_attributes]
+    volumes_attrs_list = instance_type.root_disk_size > 0 ? [default_volume_attributes] : []
 
     connection_options = {:service => "volume", :tenant_name => cloud_tenant.try(:name)}
     source.ext_management_system.with_provider_connection(connection_options) do |service|
-      requested_volumes.each do |volume_attrs|
+      requested_volumes.each_with_index do |volume_attrs, index|
+        if instance_type.root_disk_size == 0 # Handle root disk logic only when root_disk_size is 0
+          if index == 0  # The first volume should be the root disk, typically assigned to vda
```
agrare, line 9: This might be better done up above like
```
requested_volumes.first&.merge!(:boot_index => 0, :bootable => true, :imageRef => source.ems_ref) if instance_type.root_disk_size.zero?
```

spec/models/manageiq/providers/openstack/cloud_manager/provision/volume_attachment_spec.rb
```diff
@@ -1,82 +1,102 @@
...
+    ActiveRecord.yaml_column_permitted_classes = YamlPermittedClasses.app_yaml_permitted_classes | [flavor.class]
   end
 
   context "#configure_volumes" do
-    it "create volumes" do
-      service = double
-      allow(service).to receive_message_chain('volumes.create').and_return @volume
-      allow(@task.source.ext_management_system).to receive(:with_provider_connection)\
+    before do
+      allow(service).to receive_message_chain('volumes.create').and_return(volume)
```
Fryguy, line 18: I see this was here already, but I generally prefer

```suggestion
      allow(service).to receive_message_chain(:volumes, :create).and_return(volume)
```

Fryguy:
Overall LGTM- have one question about -1 vs null

spec/models/manageiq/providers/openstack/cloud_manager/provision/volume_attachment_spec.rb
agrare: Yeah it was here previously, I do have a "cleanup specs" commit in here already though so I'd rather change all of them to be consistent

app/models/manageiq/providers/openstack/cloud_manager/provision/volume_attachment.rb
agrare: > I don't really like the -1, but if that we there previously, then it's fine. I would think null would be a better choice.

I don't think that is a choice since these volume_attrs are for fog-openstack volume create.  From https://docs.openstack.org/nova/rocky/user/launch-instance-from-volume.html I don't see any mention of -1 so I'm going to drop this part so we can try to get this in.

agrare:
@Fryguy I forgot this was still open, if I drop the -1 are you good with this?

agrare:
Thanks @Dhamo1107 for starting this, if you get a chance can you confirm that my changes still work for you?

Dhamo1107:
Sure @agrare I'll confirm it by tomorrow

Dhamo1107:
Hi @agrare, I’ve tested the changes, and they’re working correctly on my end!

agrare:
Awesome great news thanks @Dhamo1107 !

agrare:
For example I created a Service Dialog with the following dialog fields:
Screenshot From 2025-10-23 13-04-56
Screenshot From 2025-10-23 13-05-00
Screenshot From 2025-10-23 13-05-05
Screenshot From 2025-10-23 13-05-08

And when I order the service, I can get this in `create_requested_volumes`:
```
[1, 10] in /home/grare/adam/src/manageiq/manageiq-providers-openstack/app/models/manageiq/providers/openstack/cloud_manager/provision/volume_attachment.rb
    1: module ManageIQ::Providers::Openstack::CloudManager::Provision::VolumeAttachment
    2:   def create_requested_volumes(requested_volumes)
    3:     byebug
=>  4:     volumes_attrs_list = instance_type.root_disk_size > 0 ? [default_volume_attributes] : []
    5: 
    6:     connection_options = {:service => "volume", :tenant_name => cloud_tenant.try(:name)}
(byebug) get_option(:root_volume_size)
1
(byebug) get_option(:root_volume_name)
"example_root_volume_name"
```

PR #897, opened 2024-12-10
Use ubuntu-24.04 to fix libqpid-proton11

agrare:
The lunar dist is no longer available, and installing from the noble dist instead (24.04) causes a conflict due to the version of libssl.

We can move the entire runner to ubuntu-24.04 to resolve the issue, this will be the default for ubuntu-latest "soon" :tm: https://github.com/actions/runner-images/issues/10636

> Rollout will begin on December 5th and will complete on January 17th, 2025.

Once that is complete we can revert back to using ubuntu-latest here.

Fixes https://github.com/ManageIQ/manageiq-providers-openstack/actions/runs/12258034148/job/34196883353
```
Ign:8 http://us.archive.ubuntu.com/ubuntu lunar InRelease
Err:9 http://us.archive.ubuntu.com/ubuntu lunar Release
  404  Not Found [IP: 91.189.91.81 80]
Reading package lists...
E: The repository 'http://us.archive.ubuntu.com/ubuntu lunar Release' does not have a Release file.
Repository: 'deb http://us.archive.ubuntu.com/ubuntu lunar universe'

 Reading package lists...
E: The value 'lunar' is invalid for APT::Default-Release as such a release is not available in the sources
```

PR #898, opened 2024-12-11
Add Openstack prefix to Cinder and Swift manager descriptions

nasark:
@miq-bot assign @agrare 
@miq-bot add_reviewer @agrare

### DIFF
--- a/app/models/manageiq/providers/openstack/storage_manager/cinder_manager.rb
+++ b/app/models/manageiq/providers/openstack/storage_manager/cinder_manager.rb
@@ -58,11 +58,11 @@ class ManageIQ::Providers::Openstack::StorageManager::CinderManager < ManageIQ::
   end
 
   def self.description
-    @description ||= "Cinder ".freeze
+    @description ||= "OpenStack Cinder".freeze
   end
 
   def description
-    @description ||= "Cinder ".freeze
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/openstack/storage_manager/cinder_manager.rb
```diff
@@ -58,11 +58,11 @@ def self.ems_type
   end
 
   def self.description
-    @description ||= "Cinder ".freeze
+    @description ||= "Openstack Cinder ".freeze
```
agrare, line 61: The CloudManager and NetworkManager both use `OpenStack` can you match that for all of these?

Fryguy:
Also, since we're making this change in a base provider, please run cross-repo tests with the descendant providers

agrare (approved):

nasark:
manageiq-providers-ibm_cic and manageiq-providers-ibm_power_vc tests passed in the cross-repo. The openstack run failed with a qpid_proton failure and is not related
  ```
  An error occurred while installing qpid_proton (0.37.0), and Bundler cannot
  continue.
  
  In Gemfile:
    qpid_proton
  ```

agrare:
Oh we need a fix like https://github.com/ManageIQ/manageiq-providers-openstack/pull/897 in those also

PR #911, opened 2025-06-18
Use virtual_has_many instead of virtual_delegate for relations

kbrock:
Part of:

- https://github.com/ManageIQ/manageiq/pull/23486


kbrock:
An ActiveSupport `delegate` works with relations or attributes. So saying that `virtual_delegate` only works with attributes is confusing. It is even more specific since it is only valid for a sql relation and a sql attribute.

We have resolved the relation side by supporting `virtual_has_many :through`.
I have a PR to put together that can migrate us from `virtual_delegate` to `virtual_attribute :through`.

We may drop `virtual_delegate` in the future to resolve the confusion.

PR #915, opened 2025-07-01
Switch to use VcrSecrets helper from core

Fryguy:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/23497

### DIFF
--- a/spec/config/secrets.defaults.yml
+++ b/spec/config/secrets.defaults.yml
@@ -1,0 +1,6 @@
+---
+openstack:
+  hostname: 11.22.33.44
+  port: '5000'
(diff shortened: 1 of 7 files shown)


spec/config/secrets.defaults.yml
```diff
@@ -0,0 +1,6 @@
+---
+openstack:
+  hostname: 11.22.33.44
+  port: '5000'
```
agrare, line 4: NOTE to self I'm surprised port is explicitly a string here seems like it'd be easier to let it be an int, I'll follow-up

agrare (approved):

PR #921, opened 2025-09-29
[WIP] Fix NetworkManager EventTargetParser

agrare:
Some event payload types don't have a `resource_id` in the payload, but they do have e.g. `{"network" => {"id" => ...}}`

Ref: https://github.com/orgs/ManageIQ/discussions/23589


niccoloalfredo:
Hi @agrare!

Yes, absolutely. I agree that a more complete solution is needed for the Network Manager's target parsing.

I have opened a new Pull Request, **#930**, which proposes a comprehensive fix and enhancement for the `network_manager/event_target_parser.rb`.

The PR includes logic to correctly handle various resource targets (like Ports, Security Groups, etc.) from modern OpenStack event payloads, addressing the need for better targeting examples.

Please feel free to review the changes directly in **PR #930**.

PR #928, opened 2025-10-30
Fix: Enhance CloudManager EventTargetParser for Identity project events

Niccolo-Alfredo:
## Description

This PR enhances the `ManageIQ::Providers::Openstack::CloudManager::EventTargetParser` to correctly extract the refresh target (Tenant ID) from **Identity (Keystone) project events** (`identity.project.*`).

In modern OpenStack deployments, the Tenant ID for events like `identity.project.created/deleted/updated` is no longer reliably found in the top-level generic keys. This prevents the targeted refresh of the **Cloud Tenant** inventory when project lifecycle events occur.

The fix ensures we prioritize extraction from the specific Keystone fields (`resource_info` or `target/id`) for Identity events, falling back to the generic fields for all other event types (Compute, Image, etc.).

---

### Location of Change
`app/models/manageiq/providers/openstack/cloud_manager/event_target_parser.rb` (in `collect_identity_tenant_references!`)

### Proposed Code Change

The `collect_identity_tenant_references!` method is updated to check for `identity.project.*` events first:

```ruby
  def collect_identity_tenant_references!(target_collection)
    if ems_event.event_type.start_with?("identity.project.")
      tenant_id = event_payload['resource_info'] || event_payload.fetch_path('target', 'id')
    else
      tenant_id = event_payload['tenant_id'] ||
                  event_payload['project_id'] ||
                  event_payload.fetch_path('initiator', 'project_id')
    end
    add_target(target_collection, :cloud_tenants, tenant_id) if tenant_id
  end
```

### Motivation: Payload Comparison

The generic extraction logic works for Nova (Compute) events because the `tenant_id` is present at the top level of the payload. However, Keystone events do not expose this ID directly, requiring specialized lookup:

| Event Type | Field Location for Target ID | Original Parser Status |
| :--- | :--- | :--- |
| **Compute (Nova)** | `payload['tenant_id']` | **WORKS** (Target ID is easily accessible) |
| **Identity (Keystone)** | `payload['resource_info']` or `payload['target']['id']` | **FAILS** (ID is deeply nested or renamed) |

-----

## Scope Note

This change specifically targets the `CloudManager`'s dependency on Identity events (Keystone) and Compute events (Nova). We have not investigated the payload structures of other services like Glance (Image) or Heat (Orchestration), as the primary tenant refresh failure was observed with Keystone events.

-----

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager/event_target_parser.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager/event_target_parser.rb
@@ -34,6 +34,8 @@ class ManageIQ::Providers::Openstack::CloudManager::EventTargetParser
       collect_host_aggregate_references!(target_collection)
     elsif ems_event.event_type.start_with?("keypair")
       collect_key_pair_references!(target_collection)
+    elsif ems_event.event_type.start_with?("identity.project.")
+      collect_identity_project_references!(target_collection)
     end
 
     target_collection.targets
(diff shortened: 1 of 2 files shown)


Niccolo-Alfredo:
Thank you so much! I appreciate you taking the time to review the changes and agree with the approach of splitting the tenant parser logic. It was necessary to ensure we capture the project target while preserving the "initiated by" information from the older payload structures.

I'm currently working on several other enhancements and fixes for OpenStack event processing, including a PR for the **network_manager/event_target_parser.rb**.

I look forward to your feedback on the next submissions!

PR #933, opened 2026-01-05
Update the `bunny` `AMQ` SDK

agrare:
The version of `bunny` that we were using was from 2015


manageiq-providers-openstack.gemspec
```diff
@@ -19,7 +19,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency "activesupport",        ">=6.0"
-  spec.add_dependency "bunny",                "~> 2.1.0"
+  spec.add_dependency "bunny",                "~> 2.24.0"
```
Fryguy: Does it make sense to just do 
```suggestion
  spec.add_dependency "bunny",                "~> 2.24"
```

PR #941, opened 2026-04-14
Fix OpenStack events capability not being set

### LINKED ISSUE #925: AMQP credential verification does not set 'events' capability
## Summary

The `verify_credentials` method in `app/models/manageiq/providers/openstack/cloud_manager.rb` contains an early exit (`return ret unless auth_type.nil?`) that prevents the automatic setting of the **`capabilities["events"]`** flag when credentials are verified with an explicit `auth_type` (e.g., `:amqp` or `:default`).

This means the capability is only updated during the rare calls where **`auth_type` is `nil`**, while the explicit verification calls (which confirm AMQP connectivity) exit early, leaving the capability flag disabled and preventing the event catchers from starting automatically.

## Environment

  * **ManageIQ Version**: Radjabov-1 (and likely earlier/later versions using the same logic)
  * **OpenStack Version**: Any version with AMQP

-----

### Location

`app/models/manageiq/providers/openstack/cloud_manager.rb` $\rightarrow$ `verify_credentials`

### Current Code (BROKEN)

```ruby
def verify_credentials(auth_type = nil, options = {})
  options[:service] ||= "Compute"
  ret = super
  return ret unless auth_type.nil?  # ← EARLY EXIT HERE for auth_type != nil

  capabilities["events"] = !!event_monitor_available?  # ← This line is often skipped!
  save! if changed?
  true
end
```

### Log Evidence and Flow Analysis

The following log snippet clearly shows two failed attempts to update the capability where `auth_type` is *not* `nil`, followed by the original code's logic being bypassed.

| Timestamp | `auth_type` | Log Message | Analysis |
| :--- | :--- | :--- | :--- |
| `10:56:28` | `:default` | `✗ auth_type is NOT nil (default) - ORIGINAL CODE WOULD EXIT HERE!` | **Failure 1:** Called with `:default` (during initial save/validation). Original code exits before checking/setting capabilities. |
| `10:56:29` | `:amqp` | `✗ auth_type is NOT nil (amqp) - ORIGINAL CODE WOULD EXIT HERE!` | **Failure 2:** Called explicitly for AMQP endpoint verification. Original code exits before checking/setting capabilities. |

**Crux of the Issue:** The `event_monitor_available?` check, which determines if the capability should be set to `true`, is skipped whenever `auth_type` is `:amqp` or `:default`. These calls are the critical moments where connectivity is confirmed and the capability *should* be set.

### Impact

  * Event catchers (cloud\_manager, storage, network) **do not start** automatically.
  * Requires a manual Rails console workaround to enable events:
    ```ruby
    ems.capabilities['events'] = true
    ems.save!
    ```

### Proposed Fix

The capability update logic must be moved to execute after `super` and only when checking for the default connection or the AMQP connection.

```ruby
def verify_credentials(auth_type = nil, options = {})
  options[:service] ||= "Compute"
  ret = super
  
  # Logic to update the event capability is moved here to ensure it runs
  # after super() succeeds, covering both auth_type=nil and auth_type="amqp".
  if auth_type.nil? || auth_type.to_s == "amqp"
    capabilities["events"] = !!event_monitor_available?
        
    save! if changed?
  end
  
  return ret unless auth_type.nil?
  
  true
end
```

Wrapping the capability update in `if auth_type.nil? || auth_type.to_s == "amqp"` ensures that the capability is checked and set only when relevant. By positioning this block *before* the early exit check, we guarantee that the event capability is updated immediately upon successful AMQP verification.

-----

## ❓ Alternative Considerations / Questions

We propose the above change to automate the capability setting. However, we want to confirm the original design intent:

  * **Was the explicit early exit for non-`nil` `auth_types` intentional, requiring the `capabilities["events"]` to be set manually or via a different code path for security or architectural reasons?**

We are ready to submit a Pull Request (PR) with the proposed fix but are open to discussion if the original implementation was deliberate.

-----

## Labels

`bug`, `critical`, `openstack`, `events`, `amqp`

-----

agrare:
The OpenStack `verify_credentials` method has logic to check if the events monitor is available only if the requested `auth_type` is `nil`.

This was originally to prevent checking the events monitor when specific auth_type was requested to be checked.

In practice however there is always an auth_type passed in when called through `authentication_check_types` because:
```
types = authentications_to_validate if respond_to?(:authentications_to_validate)
```

This does have the downside of checking the event monitor on both default and (amqp|stf).

Fixes https://github.com/ManageIQ/manageiq-providers-openstack/issues/925

### DIFF
--- a/app/models/manageiq/providers/openstack/cloud_manager.rb
+++ b/app/models/manageiq/providers/openstack/cloud_manager.rb
@@ -479,11 +479,19 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
   end
 
   def verify_credentials(auth_type = nil, options = {})
+    auth_type         ||= "default"
     options[:service] ||= "Compute"
-    ret = super
-    return ret unless auth_type.nil?
 
-    capabilities["events"] = !!event_monitor_available?
+    return unless super
+
+    # Only check if the event monitor is available if we have an events endpoint
+    # and the auth_type that we're currently checking is being used for events.
+    if events_endpoint.present?
+      capabilities["events"] = !!event_monitor_available? if auth_type_for_events == auth_type.to_s
+    else
+      capabilities["events"] = false
+    end
+
     save! if changed?
 
     true
@@ -548,6 +556,22 @@ class ManageIQ::Providers::Openstack::CloudManager < ManageIQ::Providers::CloudM
 
   def authentications_to_validate
     authentication_for_providers.collect(&:authentication_type) - [:ssh_keypair]
+  end
+
+  EVENTS_ENDPOINT_ROLES = %w[amqp ceilometer stf].freeze
+
+  def events_endpoint
+    endpoints.find_by(:role => EVENTS_ENDPOINT_ROLES)
+  end
+
+  def auth_type_for_events
+    role = events_endpoint&.role
+    case role
+    when "amqp"
+      role
+    when "ceilometer", "stf"
+      "default"
+    end
   end
 
   def volume_availability_zones
--- a/spec/models/manageiq/providers/openstack/cloud_manager_spec.rb
+++ b/spec/models/manageiq/providers/openstack/cloud_manager_spec.rb
@@ -225,16 +225,196 @@ describe ManageIQ::Providers::Openstack::CloudManager do
   end
 
   context "validation" do
-    let!(:ems) { FactoryBot.create(:ems_openstack_with_authentication) }
+    let(:zone) { EvmSpecHelper.local_miq_server.zone }
+    let!(:ems) { FactoryBot.create(:ems_openstack_with_authentication, :zone => zone) }
 
-    it "verifies AMQP credentials" do
-      EvmSpecHelper.stub_amqp_support
+    describe "#verify_credentials" do
+      context "default" do
+        context "with valid credentials" do
+          before { allow(ems).to receive(:verify_api_credentials).and_return(true) }
 
-      creds = {}
-      creds[:amqp] = {:userid => "amqp_user", :password => "amqp_password"}
-      ems.endpoints << Endpoint.create(:role => 'amqp', :hostname => 'amqp_hostname', :port => '5672')
-      ems.update_authentication(creds, :save => false)
-      expect(ems.verify_credentials(:amqp)).to be_truthy
+          it "verifies default credentials" do
+            expect(ems.verify_credentials).to be_truthy
+          end
+
+          it "sets events capability to false if no events endpoint present" do
+            ems.verify_credentials
+            expect(ems.reload.capabilities).to include("events" => false)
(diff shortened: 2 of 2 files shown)


agrare:
NOTE these specs are a little gross, going to try to clean them up a bit.

app/models/manageiq/providers/openstack/cloud_manager.rb
```diff
@@ -480,10 +480,10 @@ def swift_service
 
   def verify_credentials(auth_type = nil, options = {})
     options[:service] ||= "Compute"
-    ret = super
-    return ret unless auth_type.nil?
 
-    capabilities["events"] = !!event_monitor_available?
+    super
+
+    capabilities["events"] = events_endpoint.present? && !!event_monitor_available?
```
agrare: TODO ideally `event_monitor_available?` would only be checked for a single `auth_type`.

This is complicated by the fact that `amqp` and `stf` have their own auth_type, but `ceilometer` doesn't.
So the logic should be something like:
`if we have an stf or amqp endpoint and that is the auth_type given, or if we have ceilometer and the auth_type is nil or default`

app/models/manageiq/providers/openstack/cloud_manager.rb
```diff
@@ -550,6 +550,16 @@ def authentications_to_validate
     authentication_for_providers.collect(&:authentication_type) - [:ssh_keypair]
   end
 
+  def events_endpoint_roles
+    %w[amqp ceilometer stf].freeze
```
Fryguy: Why a method as opposed to a constant? will it be overridden by subclasses?

app/models/manageiq/providers/openstack/cloud_manager.rb
Fryguy: Is that a TODO for this PR or for a future PR? If future, did you want to add a TODO comment in the source?

agrare: I want to fix this in this PR

app/models/manageiq/providers/openstack/cloud_manager.rb
agrare: Originally I had more logic in this method, it could be a constant now

app/models/manageiq/providers/openstack/cloud_manager.rb
agrare: Okay now we are only checking `event_monitor_avaiable?` if the auth_type requested is used for events

agrare:
> NOTE these specs are a little gross, going to try to clean them up a bit.

I refactored the test contexts to cover more cases and read a little cleaner.  I separated the contexts for setting up the endpoints, if the auth is valid or not, and if the event monitor is available or not.

Fryguy (approved):

Fryguy:
Backported to tal via merge of master into tal

PR #946, opened 2026-06-30
[WIP] Drop key from schema to resolve JSX key prop spread warning

asirvadAbrahamVarghese:
Part of https://github.com/ManageIQ/manageiq-ui-classic/pull/10147, note that https://github.com/ManageIQ/manageiq-ui-classic/pull/10147 should be merged first before this

PR to drop key from schema to resolve JSX key prop spread warning:
image


agrare:
Marking `[WIP]` since the dependent PR is WIP

asirvadAbrahamVarghese:
Turns out dynamic keys are required to force the component to re-render, so closing this in favor of https://github.com/data-driven-forms/react-forms/pull/1620, which prevents the key from being spread. That fix is pulled in by https://github.com/ManageIQ/manageiq-ui-classic/pull/10154

